PR #10, opened 2019-03-29
test(cactus-web): configure jest & add button tests

Dhalton:
Configures jest for `cactus-web` and adds tests for the button component


modules/cactus-web/src/Button/Button.test.tsx
```diff
@@ -1,19 +1,85 @@
...
+  test('should default to standard variant', () => {
+    const button = render(
+      <ThemeProvider theme={cactusTheme}>
+        <Button>
+          Click me!
+        </Button>
+      </ThemeProvider>
+    )
+
+    expect(button).toMatchSnapshot()
```
JamesNimlos: We'll need to snapshot just the rendered component instead of the whole wrapper
```suggestion
    expect(button.asFragment()).toMatchSnapshot()
```

Dhalton: Ah, right! I'll update the tests

PR #16, opened 2019-04-02
feat(cactus-web): Add text buttons

Dhalton:
CACTUS-20

### DIFF
--- a/modules/cactus-theme/index.ts
+++ b/modules/cactus-theme/index.ts
@@ -31,7 +31,7 @@ export interface GeneratorOptions {
   primaryHue: number
 }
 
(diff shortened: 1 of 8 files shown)


modules/cactus-theme/index.ts
```diff
@@ -31,7 +31,7 @@ export interface GeneratorOptions {
   primaryHue: number
 }
 
-const repayOptions: GeneratorOptions = { primaryHue: 210 }
+const repayOptions: GeneratorOptions = { primaryHue: 200 }
```
JamesNimlos: oops, we'll need to re-publish with this update

PR #19, opened 2019-04-04
Separate i18n functionality to separate library

JamesNimlos:


modules/cactus-fwk/tests/featureFlags.test.js
```diff
@@ -0,0 +1,5 @@
+describe('feature flags', () => {
+  test('TODO', () => {
+    expect(true).toBe(true)
```
Dhalton: true === true ✅

PR #21, opened 2019-04-08
feat(cactus-web): Text icon buttons

Dhalton:
I branched off of the `icon-buttons` branch, so that should be reviewed and merged first. (#20)

I noticed that I didn't add a strikethrough on disabled buttons, so there's a commit in here to do that as well.

Text+Icon buttons didn't require a new component; you can just put an icon inside `TextButton` along with some text to achieve the effect, so I just added onto the `TextButton` stories/tests

### DIFF
--- a/modules/cactus-web/package.json
+++ b/modules/cactus-web/package.json
@@ -8,7 +8,6 @@
   "scripts": {
     "new": "node scripts/make-component.js",
     "test": "jest",
-    "debug": "node --inspect-brk ./node_modules/.bin/jest --runInBand --testNamePattern=\"IconButton.test.tsx\"",
     "tdd": "jest --watch",
     "test:ci": "yarn test:types && yarn test",
     "test:types": "tsc -p ./tsconfig.json --noEmit",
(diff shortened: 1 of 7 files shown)


modules/cactus-web/package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "new": "node scripts/make-component.js",
     "test": "jest",
+    "debug": "node --inspect-brk ./node_modules/.bin/jest --runInBand --testNamePattern=\"IconButton.test.tsx\"",
```
JamesNimlos: probably can remove this

Dhalton: Oh I thought we might want to keep it for debugging in the future. I'll get rid of it!

JamesNimlos:
It looks like there are some duplicated commits in this PR that came from the original icon-buttons, probably from a force push. If you 
```
git reset HEAD~5
git stash
git pull --ff-only origin master
git stash pop
```
you can then recommit just the differences.

Dhalton:
> It looks like there are some duplicated commits in this PR that came from the original icon-buttons, probably from a force push. If you
> 
> ```
> git reset HEAD~5
> git stash
> git pull --ff-only origin master
> git stash pop
> ```
> you can then recommit just the differences.

Gotcha, I was wondering why those commits were still there even after you merged. I'll do that right now. Thanks!

Dhalton:
@JamesNimlos I updated this PR with your suggestions. Should be good to go now!

PR #29, opened 2019-04-17
feat(cactus-web): Add checkbox component

Dhalton:
At long last! Checkboxes... ✅

### DIFF
--- a/modules/cactus-web/src/CheckBox/CheckBox.tsx
+++ b/modules/cactus-web/src/CheckBox/CheckBox.tsx
@@ -1,0 +1,79 @@
+import React from 'react'
+import styled from 'styled-components'
+import { StatusCheck } from '@repay/cactus-icons'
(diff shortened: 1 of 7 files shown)


modules/cactus-web/src/CheckBox/CheckBox.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
+import styled from 'styled-components'
+import { StatusCheck } from '@repay/cactus-icons'
+
+interface CheckBoxProps {
```
JamesNimlos: You could extend this with `React.HTMLProps<HTMLInputElement>`, then you don't need to add the onChange, onBlur, ... etc events manually?

Dhalton: Will do!

PR #30, opened 2019-04-17
feat(cactus-web): Add radio buttons

Dhalton:
Radio 📻 Buttons 🔘


modules/cactus-web/src/RadioButton/RadioButton.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  test('should trigger onChange event', () => {
+    const onChange = jest.fn()
+    const { getByTestId } = render(
+      <ThemeProvider theme={cactusTheme}>
+        <RadioButton name="test" id="radio" onChange={onChange} data-testid="will-change" />
+      </ThemeProvider>
+    )
+
+    fireEvent.click(getByTestId('will-change'))
```
JamesNimlos: We could do these tests using `aria-label` to ensure it shows up correctly on the radio button

PR #32, opened 2019-04-18
feat(cactus-web): Add toggles

Dhalton:
Toggles!


modules/cactus-web/src/Toggle/Toggle.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </ThemeProvider>
+    )
+
+    expect(toggle.asFragment()).toMatchSnapshot()
+  })
+
+  test('should render a disabled toggle', () => {
+    const toggle = render(
+      <ThemeProvider theme={cactusTheme}>
+        <Toggle val={false} />
```
JamesNimlos: I think we should keep the name as `value` and then we can omit it from extending the checkbox props and add a different prop to replace it, like `checkboxValue` since it's rarely used by most developers. What do you think?

PR #33, opened 2019-04-18
test(cactus-web): Add tests for disabled checkboxes

Dhalton:
Updating our checkbox tests to use `user-event`, which recently published a fix for the issue of triggering events even when the checkbox is disabled

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2355,6 +2355,11 @@
   resolved "https://registry.yarnpkg.com/@types/unist/-/unist-2.0.3.tgz#9c088679876f374eb5983f150d4787aa6fb32d7e"
   integrity sha512-FvUupuM3rlRsRtCN+fDudtmytGO6iHJuuRKS1Ss0pG5z8oX0diNEw94UEL7hgDbpN94rgaK5R7sWm6RrSkZuAQ==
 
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -4821,6 +4826,16 @@ dom-serializer@0:
...
+  version "3.19.3"
+  resolved "https://registry.yarnpkg.com/dom-testing-library/-/dom-testing-library-3.19.3.tgz#fba399987be1bdd57b07c4bc3ef46c3c084b26d9"
+  integrity sha512-oiI+oq91iO/Vpp+pt8PqfqLfBK074FH0eprhoFNvBCvJOk7vL4ozbe/yj/kEEGR6kiT4F3MAam19AX1fdGFjrA==
+  dependencies:
+    "@babel/runtime" "^7.3.4"
+    "@sheerun/mutationobserver-shim" "^0.3.2"
+    pretty-format "^24.5.0"
+    wait-for-expect "^1.1.0"
+
 dom-testing-library@^3.19.0:
```
JamesNimlos: Looks like another opportunity to resolve these to the same versions: I really feel like Yarn should do this for us. I wonder how hard it would be to put in a step that would 🤔

PR #34, opened 2019-04-23
feat(cactus-web): Add text inputs

Dhalton:
CACTUS-46


modules/cactus-web/src/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,125 @@
...
+    return statusMap[props.status]
+  }
+}
+
+const StyledCheck = styled(StatusCheck)<StyledCheckProps>`
+  width: 20px;
+  height: 20px;
+  position: relative;
+  left: -28px;
+  opacity: ${p => (p.status === 'success' && !p.disabled ? '1' : '0')};
```
JamesNimlos: Instead of using opacity to show or hide these, let's actually remove them from the DOM when the status doesn't match by using a component map or something

Dhalton:
@JamesNimlos I see that the CI pipeline is failing on this, I'll fix that, I'm just currently battling with the CI pipeline for the icon library 🙃

PR #37, opened 2019-04-24
feat(cactus-icons): Add new icons & handle new svg folder structure

Dhalton:
This PR adds support for the new svg folder structure, using the `svgr` API rather than the CLI. Because of this, I also had to make some changes to the `gen-docs` script. 

Let me know what you think!


modules/cactus-icons/scripts/gen-docs.js
```diff
@@ -1,31 +1,36 @@
...
     markdown += `  <figure
     style="display:flex;justify-content:center;flex-direction:column;align-items:center;"
   >
-    <img src="../../modules/cactus-icons/svgs/${filename}.svg"
+    <img src="../../modules/cactus-icons/svgs/${filePath}.svg"
       style="width:100px;"
       alt="svg for ${componentName}" />
     <figcaption style="min-width: 150px;text-align:center;font-size:14px;">
-      <code style="white-space:pre-wrap;font-size:10px;">/i/${filename}</code>
+      <code style="white-space:pre-wrap;font-size:10px;">/i/${filePath}</code>
```
JamesNimlos: This should represent how someone would import an individual icon from the library so it should be the kebab case, not the original file path I think.

Dhalton: Ah, yep, you're totally right. My bad, will update!

docs/Icons/Available Icons.md
```diff
@@ -4,1013 +4,1167 @@
...
-      <div>ActionsAdd</div>
+      <code style="white-space:pre-wrap;font-size:10px;">/i/actions/coachmarks</code>
+      <div>ActionsCoachmarks</div>
     </figcaption>
   </figure>
   <figure
     style="display:flex;justify-content:center;flex-direction:column;align-items:center;"
   >
-    <img src="../../modules/cactus-icons/svgs/actions-coachmarks.svg"
+    <img src="../../modules/cactus-icons/svgs/actions/add.svg"
```
JamesNimlos: I don't know how it decides to order these, but maybe we should alphabetize them or something so that they show up in the same order every time. Technically, changing the path of an icon or the name will be a Breaking change so this will be a useful check. In fact, maybe we should do another couple test files and do a snapshot of the svg urls and what is exported from root so that if it ever removes an icon or changes its import path it's obvious.

JamesNimlos: We can chat more about this tomorrow 😄

modules/cactus-icons/scripts/gen-docs.js
Dhalton: Updated!

docs/Icons/Available Icons.md
Dhalton: Yeah, that's a good point. I agree, because I noticed that the order seemed off in some places too. Sounds good we can discuss it in the AM!

PR #55, opened 2019-05-01
Styled system spacing

Dhalton:
Adds support for the margin spacing props that styled system offers. There are many, many, many changes here, so I *highly* recommend (for my own sanity) that you check out this branch and check out the stories to make sure you agree that nothing is broken and that the margin props are properly supported.

### DIFF
--- a/modules/cactus-web/src/Button/Button.test.tsx
+++ b/modules/cactus-web/src/Button/Button.test.tsx
@@ -86,7 +86,7 @@ describe('component: Button', () => {
   test('should support margin space props', () => {
     const button = render(
       <ThemeProvider theme={cactusTheme}>
-        <Button mt={5}>}>I have margins!</Button>
+        <Button mt={5}>I have margins!</Button>
       </ThemeProvider>
     )
 
--- a/modules/cactus-web/src/Button/Button.test.tsx
+++ b/modules/cactus-web/src/Button/Button.test.tsx
@@ -83,6 +83,32 @@ describe('component: Button', () => {
     expect(button.asFragment()).toMatchSnapshot()
   })
 
+  test('should support margin space props', () => {
+    const button = render(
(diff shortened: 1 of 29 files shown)


modules/cactus-web/src/Button/Button.test.tsx
```diff
@@ -83,6 +83,32 @@ describe('component: Button', () => {
     expect(button.asFragment()).toMatchSnapshot()
   })
 
+  test('should support margin space props', () => {
+    const button = render(
+      <ThemeProvider theme={cactusTheme}>
+        <Button mt={5}>}>I have margins!</Button>
```
JamesNimlos:
```suggestion
        <Button mt={5}>I have margins!</Button>
```

modules/cactus-web/src/CheckBoxField/CheckBoxField.tsx
```diff
@@ -1,41 +1,46 @@
...
   display: flex;
   flex-direction: row;
   align-items: center;
   padding-top: 4px;
   padding-bottom: 4px;
 
-  ${Label} {
-    position: relative;
-    top: 2px;
-    left: 8px;
```
JamesNimlos: What was the reason for moving this outside of here?

JamesNimlos:
Though if I hadn't thought of my previous comment I would have approved this! 😶

modules/cactus-web/src/CheckBoxField/CheckBoxField.tsx
Dhalton: The reason was that `Label` was turned into a React component instead of being a styled component and you can't reference a regular React component inside a styled component's css this way. I realized that there's another, cleaner way I could've done this a little later, but it seems like we won't be using this anyway 😄

Dhalton:
Seems like we're going to do this a different way so I'm closing this

PR #56, opened 2019-05-02
Custom space props

Dhalton:
Temporary `styled-system` types in `custom.d.ts`, as there is a bug with their type definitions. Will make a PR to that repo ASAP

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1628,10 +1628,36 @@
     "@babel/preset-typescript" "^7.3.3"
     babel-plugin-dynamic-import-node "^2.2.0"
 
+"@repay/cactus-fwk@^1.0.0":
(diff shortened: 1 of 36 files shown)


yarn.lock
```diff
@@ -1628,11 +1628,37 @@
     "@babel/preset-typescript" "^7.3.3"
     babel-plugin-dynamic-import-node "^2.2.0"
 
+"@repay/cactus-fwk@^1.0.0":
```
JamesNimlos: We should revert these to v0.3 versions

Dhalton: Ah, right! Good call. I totally forgot

JamesNimlos:
🙌

PR #57, opened 2019-05-02
feat(cactus-web): introduce ToggleField

JamesNimlos:
[CACTUS-51]

breaking(cactus-web): CheckBoxField onChange handler conforms to standard

### DIFF
--- a/modules/cactus-web/src/CheckBox/CheckBox.tsx
+++ b/modules/cactus-web/src/CheckBox/CheckBox.tsx
@@ -18,9 +18,17 @@ interface StyledCheckBoxProps extends React.HTMLProps<HTMLDivElement> {
   disabled?: boolean
 }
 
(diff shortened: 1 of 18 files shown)


modules/cactus-web/src/CheckBox/CheckBox.tsx
```diff
@@ -31,7 +39,7 @@ const HiddenCheckBox = styled.input.attrs({ type: 'checkbox' as string })`
   position: absolute;
 `
 
-const StyledCheckBox = styled.div<StyledCheckBoxProps>`
+const StyledCheckBox = styled.span<StyledCheckBoxProps>`
```
Dhalton: One thing I noticed earlier this morning is that `StyledCheckBoxProps` still extends from `HTMLDivElement`. Looks like it doesn't break anything but we still might want to change that just to be consistent

PR #59, opened 2019-05-03
Move docs

Dhalton:
Includes docs moved from their old location, new docs for new components, and refactored components so we're exporting styled components directly instead of wrapping them in React components. 😜


modules/cactus-web/src/IconButton/__snapshots__/IconButton.test.tsx.snap
```diff
@@ -37,6 +37,7 @@ exports[`component: IconButton should default to standard medium icon button 1`]
 
 <button
     class="c0"
+    display="inline-flex"
```
JamesNimlos: We should pull the display prop off of the buttonProps so it doesn't get spread with the others.

modules/cactus-web/src/Label/Label.mdx
```diff
@@ -9,11 +9,13 @@ import cactusTheme from '@repay/cactus-theme'
...
 
-## Best practices
-
 ## Basic usage
 
-TODO
+```jsx
+import { Label } from '@repay/cactus-web'
+...
+<Label>You know nothing, Jon Snow</Label>
```
JamesNimlos: 😂

modules/cactus-web/src/ToggleField/ToggleField.mdx
```diff
@@ -52,7 +52,7 @@ export default function Form() {
         onChange={handleChange}
         value={values.enabled}
       />
-      <Button type="submit" variant="callToAction">
+      <Button type="submit" variant="action">
```
JamesNimlos: thanks for this!

Dhalton: Np!

modules/cactus-web/src/Label/Label.mdx
Dhalton: Yeah...I couldn't help myself from sprinkling some GoT references in there. Hope that's alright! 😄

PR #62, opened 2019-05-06
feat(cactus-web): Add RadioButtonField component

Dhalton:
I put some fixes in here too. I set the cursor to `not-allowed` on a `CheckBoxField` label when it's disabled, and I adjusted the `RadioButton` styles to show that a radio button is selected even if it's disabled. Other than that this adds the `RadioButtonField` component!


modules/cactus-web/src/RadioButtonField/RadioButtonField.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const RadioButtonFieldBase = (props: RadioButtonFieldProps) => {
+  const { label, labelProps, id, className, name, onChange, ...radioButtonProps } = props
+  const radioButtonId = useId(id, name)
+  const handleChange = React.useCallback(
+    (event: React.FormEvent<HTMLInputElement>) => {
+      if (typeof onChange === 'function') {
+        const target = (event.target as unknown) as HTMLInputElement
+        onChange(name, target.checked)
```
JamesNimlos: Actually, value matters for radio buttons. It's used like a `<select>` and the onChange should be onChange(name, target.value). And then while rendering, `checked={values[name] === value}` so that only one radio button is selected at a time.

modules/cactus-web/src/RadioButtonField/RadioButtonField.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  test('should trigger onChange event', () => {
+    const onChange = jest.fn()
+    const { getByLabelText } = render(
+      <ThemeProvider theme={cactusTheme}>
+        <RadioButtonField id="Ackermann" name="rbf" label="Levi" onChange={onChange} />
+      </ThemeProvider>
+    )
+
+    fireEvent.click(getByLabelText('Levi'))
+    expect(onChange).toHaveBeenCalled()
```
JamesNimlos: I would validate there that it's called with the expected values as well

modules/cactus-web/src/RadioButtonField/RadioButtonField.mdx
```diff
@@ -0,0 +1,60 @@
...
+    [values]
+  )
+
+  return (
+    <form onSubmit={onSubmit}>
+      <RadioButtonField
+        name="radiobuttonfield"
+        label="Is Selected"
+        onChange={handleChange}
+        checked={values.selected}
```
JamesNimlos: A realistic example would be more like:
```jsx
export default function Form() {
  const [values, setValues] = useState({ radiobuttonfield: 'option_a' })
  const handleChange = useCallback(
    (name, value) => {
      setValues(state => setValues({ ...state, [name]: value }))
    },
    [setValues]
  )
  const onSubmit = useCallback(
    event => {
      event.preventDefault()
      const data = { ...values }
      // send data to api
    },
    [values]
  )
  return (
    <form onSubmit={onSubmit}>
      <RadioButtonField
        name="radiobuttonfield"
        label="option a"
        value="option_a"
        onChange={handleChange}
        checked={values.radiobuttonfield === 'option_a'}
      />
      <RadioButtonField
        name="radiobuttonfield"
        label="option b"
        value="option_b"
        onChange={handleChange}
        checked={values.radiobuttonfield === 'option_b'}
      />
      <Button type="submit" variant="action">
        Submit
      </Button>
    </form>
  )
}
```

PR #64, opened 2019-05-07
feat(cactus-icons): Use icon sizes from cactus theme

Dhalton:
Adds icon sizes to the theme and then makes use of them in both `cactus-web` (icon buttons) and `cactus-icons`

I had to move a copy of `custom.d.ts` to `cactus-icons` because of another error with their definitions. I think I'm going to try to make a PR to get those fixed today.


modules/cactus-icons/tests/__snapshots__/svgs.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`renders ActionsAdd 1`] = `
 <DocumentFragment>
   <svg
-    class="sc-bdVaJa bnWwrs"
+    class="sc-bdVaJa cqUrtM"
```
JamesNimlos: It's curious that these are consistent but not `c0`, `c1`, ...etc like this other snapshots. I wonder if the babel config is not being applied normally or something

Dhalton: Fixed!!

PR #71, opened 2019-05-13
feat(cactus-web): Add TextInputField component

Dhalton:
A few things to note:

- To make this possible with Mariana's new designs, I had to add some new transparent status colors to the theme.
- `StyleProvider` was added because we will need the global styles for the tooltip on the new component (and ones to come). Plus, we are just going to need the global styles eventually regardless.
- I renamed the `invalid` status on `TextInput` to `warning`, because that makes a ton more sense.
- The tooltip on `TextInputField` is not done yet. I'll need to make a new component for that and I wanted to wait on that because the changes/additions necessary in this PR just kept growing.


modules/cactus-web/src/StyleProvider/StyleProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+  font-style: normal;
+  font-stretch: normal;
+  line-height: 1.56;
+  letter-spacing: normal;
+}
+`
+
+interface StyleProviderProps extends styledComponents.ThemeProviderProps<any, any> {
+  global?: boolean
+}
```
JamesNimlos: Should we make `theme` optional on ours since we can know the default will be the exported cactus default?

modules/cactus-web/src/StyleProvider/StyleProvider.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  test('should provide the theme to children', () => {
+    const { container } = render(
+      <StyleProvider theme={cactusTheme}>
+        <TestComponent />
+      </StyleProvider>
+    )
+
+    expect(container).toMatchSnapshot()
+  })
+})
```
JamesNimlos: We should test the `global` prop also

modules/cactus-web/src/StyleProvider/StyleProvider.tsx
Dhalton: Good call, I can do that.

modules/cactus-web/src/StyleProvider/StyleProvider.test.tsx
Dhalton: Do you know how we can test it given that the global styles don't show up in the snapshot?

PR #72, opened 2019-05-15
Adds spinner and loading button state

JamesNimlos:
Also includes the cactus-addon for storybook that allows manipulating the theme.

### DIFF
--- a/modules/cactus-web/src/Button/Button.test.tsx
+++ b/modules/cactus-web/src/Button/Button.test.tsx
@@ -105,6 +105,15 @@ describe('component: Button', () => {
     )
   })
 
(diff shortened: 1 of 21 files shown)


modules/cactus-web/src/Button/Button.test.tsx
```diff
@@ -105,6 +105,15 @@ describe('component: Button', () => {
     )
   })
 
+  test('should render Spinner when loading is true', () => {
+    const { asFragment } = render(
+      <ThemeProvider theme={cactusTheme}>
```
Dhalton: Looks like this is where the CI pipeline is getting mad

PR #74, opened 2019-05-17
feat(cactus-web): Add textStyles

Dhalton:
Adds `textStyles` to the theme

## Breaking changes:
- Button (height has changed from 35px to 36px)
- TextInput (height changed from 32px to 36px to match Button)

I had to modify the styles of some other components as well, but they should still look/function the same so I wouldn't consider them breaking changes.

I also took out our custom typedefs for `styled-system`, since my changes were merged in 🎉


modules/cactus-theme/index.ts
```diff
@@ -149,6 +163,36 @@ export function generateTheme({ primaryHue }: GeneratorOptions = repayOptions):
...
+        fontSize: '12.5px',
+        lineHeight: '0.833em',
+      },
+      small: {
+        fontSize: '15px',
+        lineHeight: '0.833em',
+      },
+      medium: {
+        fontSize: '18px',
+        lineHeight: '1em',
```
JamesNimlos: Use the pixel sizes for line-height in the style-guide here, https://repaygithub.github.io/cactus/design-system/typography/

modules/cactus-theme/index.ts
```diff
@@ -149,6 +163,36 @@ export function generateTheme({ primaryHue }: GeneratorOptions = repayOptions):
...
+    textStyles: {
+      tiny: {
+        fontSize: '12.5px',
+        lineHeight: '0.833em',
+      },
+      small: {
+        fontSize: '15px',
+        lineHeight: '0.833em',
+      },
+      medium: {
```
JamesNimlos: also, let's rename this to "body"

PR #75, opened 2019-05-20
feat(cactus-web): Add TextArea component

Dhalton:
I'll need to go back and use the text styles defined in the theme when #74 is merged in, but I wanted to just get this PR up and keep it moving


modules/cactus-web/src/TextArea/TextArea.mdx
```diff
@@ -0,0 +1,50 @@
...
+    event => {
+      event.preventDefault()
+      const data = { ...values }
+      // send data to api
+    }
+  )
+  return (
+    <form onSubmit={handleSubmit}>
+      <Label id="label-id">Enter some text</Label>
+      <TextArea aria-labelledby="label-id" name="textArea" onChange={handleChange} />
```
JamesNimlos: Let's change this to recommend id on the TextArea and htmlFor on the Label

modules/cactus-web/src/TextArea/TextArea.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import userEvent from 'user-event'
+
+afterEach(cleanup)
+
+describe('component: TextArea', () => {
+  test('should render a textarea', () => {
+    const { container } = render(
+      <StyleProvider>
+        <TextArea />
+      </StyleProvider>
```
JamesNimlos: I'm wondering if we should make a custom render for react-testing-library since we always need to wrap them in this provider anyways. The benefit is that we wouldn't need to write `<StyleProvider>` every time, but the disadvantage is that it adds a level of abstraction that isn't obvious at first glance.

Just something for us to think about

modules/cactus-web/src/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,141 @@
...
+  box-sizing: border-box;
+  font-size: 18px;
+  padding: 8px 16px;
+  outline: none;
+  background-color: ${p => (p.disabled ? p.theme.colors.lightGray : p.theme.colors.white)};
+  height: ${p => p.height || 'auto'};
+  width: ${p => p.width || 'auto'};
+  resize: none;
+
+  &:focus {
```
JamesNimlos: I think this is supposed to be a `:hover` as well

modules/cactus-web/src/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,141 @@
...
+  width: 20px;
+  height: 20px;
+  position: absolute;
+  right: 8px;
+  top: 8px;
+  color: ${p => p.theme.colors.darkestContrast};
+`
+
+const TextAreaBase = (props: TextAreaProps) => {
+  const { disabled, status, className } = props
```
JamesNimlos: We should make sure className doesn't get spread onto the `<Area />` along with the others. Maybe do a `...rest` and then manually assign `disabled={disabled}`?

modules/cactus-web/src/TextArea/TextArea.test.tsx
Dhalton: I think that's a good idea, I've thought about doing that too. I'll make a ticket to explore this idea

modules/cactus-web/src/TextArea/TextArea.tsx
Dhalton: Nice catch!

modules/cactus-web/src/TextArea/TextArea.tsx
JamesNimlos: Per our conversation, I was wrong: carry on!

modules/cactus-web/scripts/make-component.js
```diff
@@ -80,7 +80,7 @@ function componentTemplate(componentName) {
 import styled from 'styled-components'
 import { margins, MarginProps } from '../helpers/margins'
 
-interface ${componentName}Props extends SpaceProps {}
+interface ${componentName}Props extends MarginProps {}
```
JamesNimlos: Good fix!

Dhalton:
@JamesNimlos you cool with me merging this?

PR #76, opened 2019-05-21
 feat(cactus-web): add Select component

JamesNimlos:
Also added a snapshot test of the export names of the library


modules/cactus-web/src/Tooltip/Tooltip.story.tsx
```diff
@@ -7,8 +7,18 @@ import Tooltip from './Tooltip'
...
-    <NotificationInfo />
-  </Tooltip>
-))
+tooltipStories
+  .add('Basic Usage', () => (
+    <Tooltip label={text('label', 'Some tooltip text here')}>
+      <NotificationInfo />
+    </Tooltip>
+  ))
+  .add(
```
Dhalton: Had no idea you could chain together `.add()`s like this. Doh! 🤦‍♂

modules/cactus-web/tests/__setup__/setup.ts
```diff
@@ -1 +1,10 @@
 import 'jest-styled-components'
+
+// filter out @reach style warnings from console during tests
+const __REACH_STYLES__ = /@reach\/[\S]+\sstyles not found/
+const _warn = console.warn
+console.warn = function filteredWarn(...args) {
```
Dhalton: Genius. Wish I thought of something like this

PR #78, opened 2019-05-28
 feat(cactus-web): Add TextAreaField component

Dhalton:
Adds `TextAreaField` component along with some other small fixes.

### Fixes
1. I noticed that the enclosing div had a bigger height than the textarea in the `TextArea` component. This was throwing off the positioning of the status message in `TextAreaField`, so I added a ref to the textarea and set the height of the div based on that.
2. I also noticed that the color of the tooltip should be call to action, so I fixed that too.
3. For some reason, the label on `TextInputField` was set to use 15px left padding instead of 16, so I fixed that as well.

Let me know if you have any questions!


JamesNimlos:
lol, I ALSO fixed the tooltip color. I'll revert it on mine after we merge yours

modules/cactus-web/src/TextArea/TextArea.tsx
```diff
@@ -114,17 +119,27 @@ const StyledError = styled(NotificationError)<StyledErrorProps>`
...
+`
+
 const TextAreaBase = (props: TextAreaProps) => {
   const { disabled, status, className, ...rest } = props
   const statusProps = { disabled, status }
 
+  const ref = useRef<HTMLTextAreaElement | null>(null)
+  let textAreaHeight = undefined
+  if (ref.current) {
+    textAreaHeight = `${ref.current.getBoundingClientRect().height}px`
```
JamesNimlos: I'm going to guess the reason this is happening is due to `textarea` being `display: inline` by default. Try changing the display to `block` instead and see how that goes

PR #82, opened 2019-05-31
feat(cactus-web): Add Link component

Dhalton:
I left the `:visited` color out of the theme for now until we can go over it with Mariana. I also changed our exports since we can export the named exports directly instead of the default exports now. Let me know if you have any questions!


modules/cactus-web/src/Link/Link.story.tsx
```diff
@@ -0,0 +1,8 @@
+import { storiesOf } from '@storybook/react'
+import { text } from '@storybook/addon-knobs'
+import Link from './Link'
+import React from 'react'
+
+storiesOf('Link', module).add('Basic Usage', () => (
+  <Link to={text('to', 'https://repaygithub.github.io/cactus/')}>{text('text', 'Click me!')}</Link>
+))
```
JamesNimlos: There should be another story of the `Link` within a block of text.

modules/cactus-web/src/Link/Link.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  :focus {
+    padding: 1px 2px 1px 1px;
+    border-radius: 4px;
+    color: ${p => p.theme.colors.base};
+    border: 2px solid ${p => p.theme.colors.callToAction};
+  }
+
+  ${p => p.theme.textStyles.body};
```
JamesNimlos: We should let it inherit whatever font-size and line-height it's inside since this is a text element

modules/cactus-web/src/Link/Link.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  :hover {
+    color: ${p => p.theme.colors.base};
+  }
+
+  :focus {
+    padding: 1px 2px 1px 1px;
+    border-radius: 4px;
+    color: ${p => p.theme.colors.base};
+    border: 2px solid ${p => p.theme.colors.callToAction};
```
JamesNimlos: Ideally, the link should not change the layout of a page or text when it receives focus but gaining padding here will probably cause some shifting. We may need to do something like what's being done with the button focus states and try an absolutely positioned element.

Dhalton:
@JamesNimlos I updated this PR with your suggestions

modules/cactus-web/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+      </StyleProvider>
+    )
+
+    expect(container).toMatchSnapshot()
+  })
+
+  test('should contain the correct href', () => {
+    const { getByText } = render(
+      <StyleProvider>
+        <Link to="https://throatpunch.com">cough cough</Link>
```
JamesNimlos: ❤️

modules/cactus-web/src/Link/Link.tsx
```diff
@@ -0,0 +1,61 @@
...
+  extends MarginProps,
+    Omit<
+      React.DetailedHTMLProps<React.LinkHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>,
+      'href'
+    > {
+  to: string
+}
+
+const LinkBase = (props: LinkProps) => {
+  const { className, to, children, ...rest } = props
```
JamesNimlos: probably should split the margin props out.

PR #86, opened 2019-06-05
feat(cactus-icons): Update to newest icons

Dhalton:
Adds the new icons that Mariana provided for us. I had to set the `cactus-icons` version to 0.5.0 to prevent packages that depend on icons from needing to be updated. It will actually be 0.4.2 when it's released. There are lots and lots of "changes" here so probably the most important things to look at are the name/export tests. Let me know what y'all think 💯


modules/cactus-icons/tests/__snapshots__/exports.test.tsx.snap
```diff
@@ -56,25 +57,32 @@ Array [
...
+  "NavigationCircleRight",
+  "NavigationCircleUp",
   "NavigationClose",
+  "NavigationDragDrop",
+  "NavigationHamburger",
   "NavigationLess",
   "NavigationLogout",
   "NavigationMenuDots",
-  "NavigationMenuLines",
-  "NavigationMenu",
```
JamesNimlos: I'm thinking the only places that would be using these icons would be our own website, so if that's the case we should be good to merge.

Dhalton: I don't see them being used anywhere in repay api so I'm going to merge this

PR #87, opened 2019-06-06
feat(cactus-web): Apply new link designs

Dhalton:
Adds a new color to the theme and updates the link styles


modules/cactus-web/src/StyleProvider/StyleProvider.tsx
```diff
@@ -29,13 +29,23 @@ body {
 }
 `
 
+const checkThemeProperties = (theme: CactusTheme) => {
+  if (theme.colors.transparentCTA === undefined) {
```
JamesNimlos: We should check `process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test'` before logging this as well,

Also, we should ensure we only run this function once.

PR #93, opened 2019-06-10
feat(cactus-web): Add Grid component

Dhalton:
CACTUS-82


modules/cactus-web/src/Grid/Grid.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+| **Breakpoint** | **Max. Number of Columns**       |
+| -------------- | -------------------------------- |
+| `tiny`         | 4                                |
+| `small`        | 6                                |
+| `medium`       | 8                                |
+| `large`        | 12                               |
+| `extraLarge`   | 12                               |
+
+
```
JamesNimlos: Can you add to the best practices or basic usage how it will use the nearest breakpoint that is less than the current viewport size, and an example describing what would happen if they set two values.

modules/cactus-web/src/Grid/Grid.tsx
```diff
@@ -0,0 +1,98 @@
...
+    React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> {}
+
+interface ItemProps
+  extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> {
+  tiny: ColumnNum
+  small?: ColumnNum
+  medium?: ColumnNum
+  large?: ColumnNum
+  extraLarge?: ColumnNum
+  debug?: boolean
```
JamesNimlos: Can you add a description of what the debug prop does or maybe we don't need it anymore after you added the boxes in storybook?

modules/cactus-web/src/Grid/Grid.mdx
Dhalton: Fa sho! I meant to explain that, I must've forgot.

modules/cactus-web/src/Grid/Grid.tsx
Dhalton: Yeah I debated with myself about whether or not I should take it out. I *suppose* it's not really necessary since I'm using boxes in the storybook. You don't think the devs would ever use it?

PR #94, opened 2019-06-11
feat(cactus-fwk): Add error boundary component

Dhalton:
CACTUS-39

### DIFF
--- a/modules/cactus-fwk/src/ErrorBoundary.tsx
+++ b/modules/cactus-fwk/src/ErrorBoundary.tsx
@@ -1,7 +1,12 @@
-import React, { ErrorInfo, Fragment, ReactElement } from 'react'
+import React, { ComponentType, ErrorInfo, Fragment } from 'react'
 
-export type ErrorView = (error: Error, info: ErrorInfo) => ReactElement
+export type ErrorView = ComponentType<ErrorViewProps>
 export type OnError = (error: Error, info: ErrorInfo) => void
+
+interface ErrorViewProps {
+  error: Error
+  info: ErrorInfo
+}
 
 interface ErrorBoundaryProps {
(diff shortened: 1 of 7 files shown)


JamesNimlos:
Let's add a `withErrorBoundary` higher order component for them to use also

Dhalton:
Closing until I can refactor into a HOC

modules/cactus-fwk/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+      return <Fragment>{children}</Fragment>
+    }
+  }
+}
+
+const noop = (error: Error, info: React.ErrorInfo) => {}
+
+const withErrorBoundary = <BaseProps extends any>(
+  BaseComponent: ComponentType<BaseProps>,
+  onError?: OnError,
```
JamesNimlos: What do you think about making this mandatory and throwing if you don't provide an `onError` here instead?

modules/cactus-fwk/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+interface ErrorBoundaryState {
+  error: Error | undefined
+  errorInfo: ErrorInfo | undefined
+}
+
+const initialState: ErrorBoundaryState = { error: undefined, errorInfo: undefined }
+
+export class ErrorBoundary extends React.Component<ErrorBoundaryProps, ErrorBoundaryState> {
+  readonly state: ErrorBoundaryState = initialState
```
JamesNimlos: We could also potentially throw if they don't provide an `onError` onto the component as well to ensure developers are adding it? Or we could start adding propTypes definitions now?

PR #95, opened 2019-06-12
feat(cactus-web): Add new icon button focus state

Dhalton:
Alright, so I also added the `justify` prop to `Grid` and that will give the devs some more control over the content inside grids. I noticed that would be a nice feature when I was using the grid to lay out the new story for the icon buttons. Other than that this is the new icon button focus state that we talked about this morning


JamesNimlos:
🎉 icon button focus states!

PR #97, opened 2019-06-13
website(docs): Updated shared-styles to different table layout in mob…

vvyverman1:
…ile view. CACTUS-112


website/src/components/Color.tsx
```diff
@@ -273,64 +358,50 @@ export function AccessibilityBox({ color, title, isDark }: AccessibilityBoxProps
   let thisColor = new Color(hslaStr)
   const contrastToWhite = thisColor.contrast(whiteColor)
   const contrastToDarkestContrast = thisColor.contrast(darkestContrastColor)
+  let t = title
```
JamesNimlos: do we still need this variable assignment or can we delete it?

website/src/components/Color.tsx
```diff
@@ -162,27 +162,112 @@ Cell.defaultProps = {
...
 const AccessibilityFlex = styled(Flex)`
-  width: device-width;
+  max-width: 500px;
+  margin: 16px;
+  @media only screen and (max-width: 500px) {
+    max-width: 95%;
+  }
+`
+const SpanRatio = styled(Span)`
+  font-size: 95%;
```
JamesNimlos: Instead of referencing percentages, let's use the fontSizes that are defined in the theme

website/src/pages/design-system/color.tsx
```diff
@@ -186,13 +186,7 @@ export default () => {
...
         illustrations.
       </Text>
-      <Flex
-        display="flex"
-        justifyContent="center"
-        flexWrap="wrap"
-        maxWidth="1200px"
-        margin="0 auto"
-      >
+      <Flex display="flex" justifyContent="center" flexWrap="wrap" maxWidth="1200" margin="0 ">
```
JamesNimlos: extra space on the margin

website/src/pages/design-system/language.tsx
```diff
@@ -10,7 +10,11 @@ const Em = styled.em`
   font-weight: 600;
   color: ${p => p.theme.colors.callToAction};
 `
-
+const P = styled.p`
+  @media only screen and (max-width: 500px) {
+    min-width: 300px;
```
JamesNimlos: instead of forcing a min-width on these paragraph tags, let's try converting this page to use the new `Grid` component Dhalton made and have them be full width at the tiny size

vvyverman1: Good idea!

PR #103, opened 2019-06-19
website(docs): update props table for mobile view.

vvyverman1:
CACTUS-101

### DIFF
--- a/website/src/components/PropsTable.tsx
+++ b/website/src/components/PropsTable.tsx
@@ -9,7 +9,8 @@ import styled from 'styled-components'
 
 const Table = styled.table`
   border-radius: 8px;
-  width: 100%;
+  max-width: 100%;
+
   padding: 0;
   border-collapse: collapse;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


website/src/components/PropsTable.tsx
```diff
@@ -47,6 +48,80 @@ const Table = styled.table`
...
+    }
+
+    td {
+      position: relative;
+      padding: 0 10% 10% 35%;
+      border-bottom: 1px solid #eee;
+      overflow-x: auto;
+    }
+
+    td:before {
```
JamesNimlos: for convention purposes, I'd recommend using `::` for pseudo elements, but that's not a blocker.

website/src/components/PropsTable.tsx
```diff
@@ -47,6 +48,80 @@ const Table = styled.table`
...
+    }
+      td:nth-of-type(2):before {
+      content: ' Required';
+    }
+
+    td:nth-of-type(3):before {
+      content: 'Type';
+    }
+    td:nth-of-type(4):before {
+      content: 'Description';
```
JamesNimlos: You could try adding these titles using the `attr(data-title)` or something similar so you don't need to extend the tables separately

PR #105, opened 2019-06-21
feat(cactus-web): Created avatar component.

vvyverman1:
CACTUS-124


modules/cactus-web/src/Avatars/Avatars.tsx
```diff
@@ -0,0 +1,141 @@
+import { CactusTheme } from '@repay/cactus-theme'
+import { MarginProps, margins } from '../helpers/margins'
+import styled, { css, FlattenInterpolation, ThemeProps } from 'styled-components'
+
+export type AvatarUsage = 'alert' | 'feedBack' | 'steps'
+export type AvatarColor =
+  | 'NotificationError'
+  | 'NotificationAlert'
+  | 'NotificationInfo'
+  | 'StatusCheck'
```
JamesNimlos: I think we should make these options `'error' | 'warning' | 'info' | 'success'`

modules/cactus-web/src/Avatars/Avatars.tsx
```diff
@@ -0,0 +1,141 @@
...
+export type AvatarStep = 'notDone' | 'inProcess' | 'done'
+export type CurrentStep = '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
+
+interface AvatarProps extends MarginProps {
+  avatarUsage?: AvatarUsage
+  avatarColor?: AvatarColor
+  avatarStep?: AvatarStep
+  stepNumber?: CurrentStep
```
JamesNimlos: I'm wondering if maybe it would make more sense to have the Avatar and the Steps be different components since their props are so different?

modules/cactus-web/src/Avatars/Avatars.tsx
vvyverman1: I think that would be better too, they are currently labeled liked that because that is the actual name of the icon that they render.

modules/cactus-web/src/Avatars/Avatars.tsx
vvyverman1: Good idea!

modules/cactus-web/src/Avatars/Avatars.tsx
```diff
@@ -0,0 +1,108 @@
...
+  info: css`
+    background: ${avaColor};
+  `,
+  success: css`
+    background: ${avaColor};
+  `,
+}
+
+const variant = (props: AvatarProps): FlattenInterpolation<ThemeProps<CactusTheme>> | undefined => {
+  const { avatarType, avatarUsage } = props
```
Dhalton: I'm thinking we don't need to pull `avatarUsage` here. It looks like it's not being used.

modules/cactus-web/src/StepAvatar/StepAvatar.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return stepColorMap[stepType]
+}
+
+export const StepAvatar = styled.div<StepAvatarProps>`
+  border-radius: 50%;
+  width: 48px;
+  height: 48px;
+
+  font-size: ${p => p.theme.textStyles.h2.fontSize};
+  font-weight: 400px;
```
JamesNimlos: font-weight doesn't have a unit
```suggestion
  font-weight: 400;
```

modules/cactus-web/src/StepAvatar/StepAvatar.tsx
```diff
@@ -0,0 +1,53 @@
...
+const variant = (
+  props: StepAvatarProps
+): FlattenInterpolation<ThemeProps<CactusTheme>> | undefined => {
+  const { stepType } = props
+  //@ts-ignore
+  return stepColorMap[stepType]
+}
+
+export const StepAvatar = styled.div<StepAvatarProps>`
+  border-radius: 50%;
```
JamesNimlos: We need to add `box-sizing: border-box;` style because we learned previously that box-sizing is being set globally in the application.

modules/cactus-web/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+const avaColor = (props: AvatarProps & ThemeProps<CactusTheme>) => {
+  const { avatarUsage, avatarType } = props
+
+  if (avatarUsage === 'alert') {
+    switch (avatarType) {
+      case 'error':
+        return `hsla(353, 84%, 44%, 0.3)`
```
JamesNimlos: I think you're waiting on Dhalton's PR to get these from the theme, but maybe then we should remove these components from the main export file until the theme is updated?

modules/cactus-web/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,145 @@
...
+} from '@repay/cactus-icons'
+import React from 'react'
+import styled, { css, FlattenInterpolation, ThemeProps } from 'styled-components'
+
+export type AvatarUsage = 'alert' | 'feedBack'
+export type AvatarType = 'error' | 'warning' | 'info' | 'success'
+
+interface AvatarProps extends MarginProps {
+  avatarUsage?: AvatarUsage
+  avatarType?: AvatarType
```
JamesNimlos: Let's name these `usage` and `status` or `type` and `status`. Prepending avatar feels redundant.

modules/cactus-web/src/StepAvatar/StepAvatar.tsx
```diff
@@ -0,0 +1,53 @@
+import { CactusTheme } from '@repay/cactus-theme'
+import { MarginProps, margins } from '../helpers/margins'
+import styled, { css, FlattenInterpolation, ThemeProps } from 'styled-components'
+
+export type AvatarStep = 'notDone' | 'inProcess' | 'done'
+
+interface StepAvatarProps extends MarginProps {
+  stepType?: AvatarStep
```
JamesNimlos: Can we just call this `status` also?

modules/cactus-web/src/StepAvatar/StepAvatar.tsx
JamesNimlos:
```css
  line-height: 49px;
```
Add line-height to fix weird cross browser font rendering inconsistency.

PR #108, opened 2019-06-24
 feat(cactus-web): Add file input component

Dhalton:
Adds file input to `cactus-web` & the avatar colors to `cactus-theme`


modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,624 @@
...
+              onRetry={handleRetry}
+              status={file.status}
+              errorMsg={file.errorMsg}
+              labels={labels}
+              boxRef={index === 0 ? topFileBox : undefined}
+            />
+          ))}
+          <TextButton variant="action" onClick={handleOpenFileSelect}>
+            <BatchstatusOpen iconSize="small" />
+            Select Files...
```
JamesNimlos: This needs to be translatable.

modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,624 @@
...
+  | 'EncodingError'
+  | 'UnknownError'
+type FileStatus = 'loading' | 'loaded' | 'error'
+
+interface FileInputProps
+  extends MarginProps,
+    MaxWidthProps,
+    WidthProps,
+    Omit<
+      React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>,
```
JamesNimlos: An input will ideally accept `onFocus` and `onBlur` as a form element. Probably on the main button would be sufficient.

modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,624 @@
...
+
+interface State {
+  files: FileObject[]
+  inputKey: string
+}
+
+const EmptyPromptsBase = (props: EmptyPromptsProps) => (
+  <div className={props.className}>
+    <ActionsUpload iconSize="large" />
+    <span>Drag files here or</span>
```
JamesNimlos: More text to make translatable.

modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,624 @@
...
+          new Promise<FileObject>(resolve => {
+            resolve({ fileName: file.name, contents: null, status: 'loading' })
+          })
+      )
+
+      Promise.all(loadingPromises).then(results => {
+        if (isMounted.current === true) {
+          setState(state => ({ ...state, files: results }))
+          if (typeof onChange === 'function') {
+            onChange(name, results)
```
JamesNimlos: Let's leave this call to `onChange` out and if they want access to the state at this point we can add a different api for that. Instead we can just set the state synchronously with the loading files.

modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,624 @@
...
+              fileName: file.name,
+              contents: null,
+              status: 'error',
+              errorMsg: errorMsg,
+            })
+          })
+        }
+
+        if (rawFiles) {
+          return new Promise<FileObject>(resolve => {
```
JamesNimlos:
```suggestion
          return Promise.resolve({
            fileName: file.name,
            contents: file,
            status: 'loaded'
          })
```
You can actually just use Promise.resolve here, and it's the same just a little cleaner.

modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,673 @@
...
+    }
+
+    ${ActionsRefresh} {
+      height: 12px;
+      width: 12px;
+    }
+  }
+`
+
+const defaultErrorHandler = (errorType: ErrorType, accept: string[] | undefined = []): string => {
```
JamesNimlos: Let's add this as a static property on the exported component so people can use it when they only want to override a single error type.

PR #113, opened 2019-06-27
feat(cactus-i18n): allow section language override

JamesNimlos:
[CACTUS-128]


examples/standard/src/locales/es/global.js
```diff
@@ -3,4 +3,5 @@ welcome-message = Bienvenido a la aplicación estándar usando \`@repay/cactus-i
   demostrando los usos básicos
 home-link = Vete a página de inicio
 feature-carrots-label = Incluir zanahorias en bocadillos
+terms-title = Terms and Conditions
```
Dhalton: Shouldn't this be the Spanish translation?

PR #114, opened 2019-06-27
feat(cactus-web): Addded media queries and breakpoints scale to the s…

vvyverman1:
…tyle provider. Created a story as proof of concept for the breaking points.

CACTUS-125


modules/cactus-theme/index.ts
```diff
@@ -42,6 +42,8 @@ export type TextStyle = {
 }
 
 export interface CactusTheme {
+  breakpoints?: Array<String>
+  mediaQueries?: {}
```
JamesNimlos: I think we should name the media queries we set, so small, medium, large, extraLarge

modules/cactus-web/src/StyleProvider/StyleProvider.mdx
```diff
@@ -30,6 +31,40 @@ const App = () => {
 }
 ```
 
-## Properties
+# Breaking Points
```
JamesNimlos: Just "Breakpoints"

modules/cactus-web/src/StyleProvider/StyleProvider.mdx
```diff
@@ -30,6 +31,40 @@ const App = () => {
...
+
+The `StyleProvider` has a media queries scale which is based on the provided breaking points, and is added to the given theme. The media queries have screen sizes small, medium, large, and extralarge. The default media query is a screen size less than or equal to 320px.
```
JamesNimlos: The default is actually anything and then each media query describes screens _bigger_ than itself in a "mobile first" approach. I.e. the default should describe a mobile screen and then any screen that breaks from that would add custom media queries to change the layout for larger sizes.

modules/cactus-web/src/StyleProvider/StyleProvider.mdx
```diff
@@ -30,6 +31,40 @@ const App = () => {
...
 
-<PropsTable of={StyleProvider} />
+const App = () => {
+  return (
+    <StyleProvider theme={cactusTheme} global={true}>
+      <BreakpointBox backgroundColor={backgroundColors} width="100px" height="100px" />
+    </StyleProvider>
+  )
+}
+```
```
JamesNimlos: Can you also provide an example using the array prop and then provide a link to the styled-system documentation from the story?

PR #115, opened 2019-06-28
chore(): Set maxWorkers on CI test pipeline

Dhalton:
This **should** fix the problems with memory that we've been seeing in our CI pipeline the last couple weeks. I added `--maxWorkers=2` to the `test:ci` commands wherever appropriate and added a `test:local` command without the restriction on max workers to each package that should be used for simulating the CI testing locally


package.json
```diff
@@ -21,6 +21,7 @@
     "fmt": "yarn lint --fix",
     "test": "yarn test:ci",
```
JamesNimlos: Let's change the normal test command to call `test:local` instead of ci

Dhalton: Yeah must've missed that! Whoops!

PR #116, opened 2019-07-01
Alert component

vvyverman1:
CACTUS-96


modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,177 @@
...
+            <PushAvatarBase type={status} usage="alert" />
+          </Grid.Item>
+          <Grid.Item tiny={8}>
+            <PushTextAlign>{props.children}</PushTextAlign>
+          </Grid.Item>
+
+          <Grid.Item tiny={2}>
+            <CloseIcon>
+              <Button>
+                <NavigationClose onClick={onClick} iconSize="small" />
```
JamesNimlos: It's correct that this is wrapped in a `button` but the `onClick` needs to be on the `Button` instead of the icon.

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+const GeneralAvatarBase = styled(Avatar)`
+  margin-top: 24px;
+  margin-left: 16px;
+`
+const PushAvatarBase = styled(Avatar)`
+  margin-top: 10px;
+  margin-left: 16px;
+`
+const GeneralTextAlign = styled(Box)`
```
JamesNimlos: `Box` should be used when we want access to the styled-system props it has, if we are just adding static styles like this we should just use `styled.div`

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,177 @@
...
+  margin-bottom: 16px;
+`
+
+const CloseIcon = styled(Box)`
+  margin-top: 15px;
+  justify: center;
+`
+const Button = styled('button')`
+  background: none;
+  border: none;
```
JamesNimlos: Also, add `appearance: none` which is required to remove button specific styles in some browsers.

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+import { NavigationClose } from '@repay/cactus-icons'
+
+export type Status = 'error' | 'warning' | 'info' | 'success'
+export type Type = 'general' | 'push'
+
+interface AlertProps extends MarginProps {
+  status?: Status
+  type?: Type
+  onClick?: any
```
JamesNimlos: Let's call this `onClose` so that it's clear why this will be called versus when the alert is generically called. Also, in the type definition we should specify more exactly that it's a function.

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+const AlertBase = (props: AlertProps) => {
+  const { className, type: type, status: status, onClick: onClick } = props
+
+  if (type === 'push') {
+    return (
+      <div className={className}>
+        <Grid>
+          <Grid.Item tiny={2}>
+            <PushAvatarBase type={status} usage="alert" />
```
JamesNimlos: The fact that we are providing `type={status}` is a "code smell" to me; I would think that `status` would be referring to a consistent property and when it has similar values to another component it would be called the same thing.

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,177 @@
...
+              </Button>
+            </CloseIcon>
+          </Grid.Item>
+        </Grid>
+      </div>
+    )
+  } else {
+    return (
+      <div className={className}>
+        <Box>
```
JamesNimlos: This Box is unnecessary here; if it's adding some benefit, we should figure out what style is being applied that we need to add at the top level of this component instead.

modules/cactus-web/src/StyleProvider/StyleProvider.mdx
```diff
@@ -30,6 +31,40 @@ const App = () => {
...
+  ${p => p.theme.mediaQueries.large} {
+    border-radius: 45%;
+  }
+  ${p => p.theme.mediaQueries.extralarge} {
+    border-radius: 50%;
+  }
+`
+const backgroundColors = ['springgreen', 'mediumturquoise', 'cornflowerblue', 'thistle', 'pink']
 
-<PropsTable of={StyleProvider} />
```
JamesNimlos: We shouldn't delete this `PropsTable`

modules/cactus-web/src/Alert/Alert.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+TODO - brief description
+
+```jsx
+import React from 'react'
+
+;<Alert />
+```
+
+```tsx
```
JamesNimlos: Since there is nothing special about the usage in a typescript environment I'd delete this example.

modules/cactus-theme/index.ts
```diff
@@ -42,6 +42,8 @@ export type TextStyle = {
 }
 
 export interface CactusTheme {
+  breakpoints?: Array<String>
+  mediaQueries?: {}
```
JamesNimlos: I think this is a hold over from the other commits. Let's work to remove this change from these commits.

modules/cactus-web/src/StyleProvider/StyleProvider.tsx
```diff
@@ -77,11 +86,14 @@ interface StyleProviderProps extends Omit<styledComponents.ThemeProviderProps<an
 }
 
 export const StyleProvider: React.FC<StyleProviderProps> = props => {
-  const { global, children, theme, ...themeProviderProps } = props
-  shouldCheckTheme && checkThemeProperties(theme || cactusTheme)
+  const { global, children, theme: providedTheme, ...themeProviderProps } = props
+  shouldCheckTheme && checkThemeProperties(providedTheme || cactusTheme)
+  const theme = providedTheme ? providedTheme : cactusTheme
+  theme.breakpoints = breakpoints.slice()
+  theme.mediaQueries = queries
```
JamesNimlos: Again, I think these are getting pulled in incorrectly because there are duplicate commits between this branch and the other.

Dhalton:
Wait, I'm confused. I just reviewed this code in a different PR. These two branches seem to be overlapping or something

vvyverman1:
> Wait, I'm confused. I just reviewed this code in a different PR. These two branches seem to be overlapping or something

I am also confused

modules/cactus-web/src/Alert/Alert.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+The `Alert` component is designed to provide user feedback. The alert is a rectangular shape providing the user a current status based on the icon and background color.
+
+## Best practices
+
+The `Alert` component should be used to provide users with feedback based on the user's status. The status can be success, error, warning, and info. The `Alert` component should be set to a general alert or a push notification alert. The general alert is larger in size than the push notifcation. The push notification alert has a button, on the top right side, and an onClick action should be set to handle the click event when implementing a push notification.
+
+## Basic usage
+
+TODO - brief description
```
Dhalton: Looks like there's a TODO left here that we could get rid of

JamesNimlos: I think we can delete this `TODO` because the code example is self explanatory

PR #118, opened 2019-07-01
Define side effects

JamesNimlos:
[Enables tree-shaking](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) for webpack users of the libraries.


JamesNimlos:
Closing to upgrade `@repay/scripts` as well

PR #119, opened 2019-07-01
Fix component docs

vvyverman1:


modules/cactus-web/src/Alert/Alert.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+The `Alert` component is designed to provide user feedback. The alert is a rectangular shape providing the user a current status based on the icon and background color.
+
+## Best practices
+
+The `Alert` component should be used to provide users with feedback based on the user's status. The status can be success, error, warning, and info. The `Alert` component should be set to a general alert or a push notification alert. The general alert is larger in size than the push notifcation. The push notification alert has a button, on the top right side, and an onClick action should be set to handle the click event when implementing a push notification.
+
+## Basic usage
+
+TODO - brief description
```
Dhalton: We should probably get rid of this TODO

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,176 @@
...
+  justify: center;
+`
+const Button = styled('button')`
+  background: none;
+  border: none;
+  appearance: none;
+`
+
+const AlertBase = (props: AlertProps) => {
+  const { className, type: type, status: status, onClose: onClose } = props
```
Dhalton: You don't need to use the colon notation when you're destructuring these props. `const { className, type, status, onClose } = props` should work fine.

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,176 @@
...
+          <Grid.Item tiny={2}>
+            <PushAvatarBase status={status} type="alert" />
+          </Grid.Item>
+          <Grid.Item tiny={8}>
+            <PushTextAlign>{props.children}</PushTextAlign>
+          </Grid.Item>
+
+          <Grid.Item tiny={2}>
+            <CloseIcon>
+              <Button onClick={onClose}>
```
Dhalton: We have an `IconButton` component that can be used here instead of creating your own `Button`

PR #125, opened 2019-07-10
fix(cactus-web): Fix Accordion not closing bug

Dhalton:
CACTUS-84

### DIFF
--- a/modules/cactus-web/src/Accordion/Accordion.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.tsx
@@ -101,20 +101,28 @@ AccordionHeader.propTypes = {
   closeLabel: PropTypes.string,
 }
 
+const BODY_PADDING_TOP = 24
+const BODY_PADDING_BOTTOM = 40
+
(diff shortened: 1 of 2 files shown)


modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -104,17 +104,17 @@ AccordionHeader.propTypes = {
...
+  const [state, setState] = useState({ height: 0, paddingTop: 0, paddingBottom: 0 })
   const containerRef = useRef<HTMLDivElement | null>(null)
 
   useEffect(() => {
     if (isOpen && containerRef.current) {
-      const newHeight = containerRef.current.getBoundingClientRect().height
+      const newHeight = containerRef.current.getBoundingClientRect().height + 64
       if (newHeight !== state.height) {
-        setState({ height: newHeight })
+        setState({ height: newHeight, paddingTop: 24, paddingBottom: 40 })
```
JamesNimlos: let's make these variables to ensure we don't accidentally change one and not the addition above

PR #126, opened 2019-07-10
feat(examples): Add example app as a mock ebpp. Implements various ca…

vvyverman1:
…ctus components.

CACTUS-144


examples/mock-ebpp/src/containers/home.tsx
```diff
@@ -0,0 +1,64 @@
...
+const Home = (props: HomeProps) => {
+  return (
+    <div>
+      <Helmet>
+        <title> Home</title>
+      </Helmet>
+
+      <Flex justifyContent="center">
+        <Flex width="90%" backgroundColor="base" alignItems="center" paddingLeft="10px">
+          <Text color="white" textStyle="h2">
```
JamesNimlos: There is also a `colorStyles` property on the theme that should be able to be used the `colors="base"`. This should set both the background color and the color styles so that you don't have to define it twice.

examples/mock-ebpp/src/containers/payment-history-report.tsx
```diff
@@ -0,0 +1,89 @@
...
+            }}
+          >
+            <thead style={{ backgroundColor: 'grey', fontSize: '26px', border: '2px solid grey' }}>
+              <tr>
+                <th> First Name</th>
+                <th> Last Name</th>
+                <th> Pay Ref</th>
+                <th> Date</th>
+                <th> Last Four</th>
+                <th> ID</th>
```
JamesNimlos: The spaces here shouldn't affect the layout of the rendered text because it's a `display: table-cell` I think, but it may be best practice to remove them.

examples/mock-ebpp/src/containers/ui-config.tsx
```diff
@@ -0,0 +1,274 @@
...
+  const handleReset = () => {
+    console.log(state.formData.file_input)
+    setState({ ...getInitialState() })
+    console.log(state.formData.file_input)
+  }
+
+  const arr = ['vvyverman@repay.com', 'dhuber@repay.com']
+
+  return (
+    <div>
```
JamesNimlos: Since everything else is already wrapped in a `Flex` this would be a good use for `React.Fragment`

examples/mock-ebpp/src/containers/ui-config.tsx
```diff
@@ -0,0 +1,274 @@
...
+                value={state.formData.display_name}
+                tooltip="Enter your merchant display name"
+              />
+
+              <TextInputField
+                onChange={handleChange}
+                name="merchant_name"
+                label="Merchant Name"
+                value={state.formData.merchant_name}
+                my={3}
```
JamesNimlos: I think we should build this style into a `Field` abstraction so that we don't need to add it always.

examples/mock-ebpp/src/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    </div>
+  )
+}
+
+const App = () => {
+  return (
+    <div>
+      <Helmet>
+        <title> Home </title>
+        <link rel="icon" type="image/png" sizes="16x16" href="src/assets/favicon.ico" />
```
JamesNimlos: I totally forgot, that a `*.ico` file won't be picked up by the loaders we have either way so getting this to work is a little more involved. We can try and fix it later when we come back to this.

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -75,7 +75,7 @@ const AccordionHeaderBase = (props: AccordionHeaderProps) => {
 export const AccordionHeader = styled(AccordionHeaderBase)`
   box-sizing: border-box;
   width: 100%;
-  height: 48px;
+  min-height: 48px;
```
JamesNimlos: A change in cactus-web like this should be made as a separate commit so we can reference it in the change log. Let's walk through how to update this.

PR #130, opened 2019-07-24
fix(cactus-web): Make Alert width responsive, and adjust layout of pu…

vvyverman1:
…sh to work with any width.

### DIFF
--- a/modules/cactus-web/src/Alert/Alert.story.tsx
+++ b/modules/cactus-web/src/Alert/Alert.story.tsx
@@ -4,10 +4,16 @@ import { select, text } from '@storybook/addon-knobs/react'
 import { storiesOf } from '@storybook/react'
 import Alert, { Status, Type } from './Alert'
 import React from 'react'
+import styled from 'styled-components'
 
 const status: Status[] = ['error', 'warning', 'info', 'success']
 const type: Type[] = ['general', 'push']
 const eventLoggers = actions('onClose')
+
+const ShadowFlex = styled(Flex)`
+  box-shadow: 0 9px 24px ${p => p.theme.colors.callToAction};
+  border-radius: 8px;
+`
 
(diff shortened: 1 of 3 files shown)


modules/cactus-web/src/Alert/Alert.story.tsx
```diff
@@ -1,21 +1,56 @@
...
 import { storiesOf } from '@storybook/react'
 import Alert, { Status, Type } from './Alert'
 import React from 'react'
+import styled from 'styled-components'
 
 const status: Status[] = ['error', 'warning', 'info', 'success']
 const type: Type[] = ['general', 'push']
 const eventLoggers = actions('onClose')
 
+const ShadowFlex = styled(Flex)`
```
Dhalton: What's the need for this `ShadowFlex` component?

Dhalton: If the alerts are going to need shadows, we should probably make that a part of the alert component itself

vvyverman1: Yeah I guess we will be using the shadow, I added it in the component instead.

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -167,12 +176,15 @@ export const Alert = styled(AlertBase)<AlertProps>`
   ${margins}
   ${backgroundVariant}
   ${sizeVariant}
+  ${boxShadow}
```
Dhalton: Would you be opposed to just adding something like ```box-shadow: ${p => p.shadow && `0 9px 24px ${p.theme.colors.callToAction};`}``` instead of making it a function? Just doesn't seem necessary

vvyverman1: I think you're right, that seems better.

PR #132, opened 2019-08-01
Refactor code and add close button if "onClose" is set

vvyverman1:


modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -99,68 +99,50 @@ const sizeVariant = (
...
   appearance: none;
   margin-right: 16px;
 `
 
+const closeButton = (onClose) => {
+  return(
+    <Flex justifyContent="flex-end">
+      <Grid.Item tiny={1}>
+        <div style={{marginTop:"12px"}}>
+          <Button onClick={onClose}>
```
Dhalton: I think we should probably just use the `IconButton` component here and then we wouldn't have a need for this `Button`. Then you can give it an aria label with the `aria-label` prop, and the cursor will change and such

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -99,68 +99,50 @@ const sizeVariant = (
...
-  margin-top: 12px;
-`
 const Button = styled('button')`
   background: none;
   border: none;
   appearance: none;
   margin-right: 16px;
 `
 
+const closeButton = (onClose: any) => {
```
Dhalton: Also, if you want to get really wild, you could make this into a component with an `onClose` prop instead of a function. Then you could use it like this: `<CloseButton onClose={onClose} />`. Just a suggestion

modules/cactus-web/src/Alert/Alert.tsx
vvyverman1: Good idea 🙂

PR #140, opened 2019-08-14
fix(cactus-web): Remove special characters in select ids

Dhalton:
When I was adding the select fields in channels, I noticed that we were getting some console errors when the query selector to find select options contained a url in its value. Turns out it was because of some special characters in the url (colon, slash, period). This PR just removes them, but if you want to replace them with something other than empty string I'm all for it.

### DIFF
--- a/modules/cactus-web/src/Select/Select.tsx
+++ b/modules/cactus-web/src/Select/Select.tsx
@@ -163,7 +163,7 @@ function asOption(opt: string | OptionType): OptionType {
 }
 
 function getOptionId(option: OptionType) {
-  return `${option.label}-${option.value}`.replace(/\s/g, '-').replace(/:|\.|\//g, '')
+  return `${option.label}-${option.value}`.replace(/[\s:.\/]/g, '')
 }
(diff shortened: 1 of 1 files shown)


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -163,7 +163,7 @@ function asOption(opt: string | OptionType): OptionType {
 }
 
 function getOptionId(option: OptionType) {
-  return `${option.label}-${option.value}`.replace(/\s/g, '-')
+  return `${option.label}-${option.value}`.replace(/\s/g, '-').replace(/:|\.|\//g, '')
```
JamesNimlos: We could write this regex a little more efficiently if we combine them 
```suggestion
  return `${option.label}-${option.value}`.replace(/[\s:.\/]/g, '')
```

Dhalton: So you're okay with just removing the spaces altogether instead of replacing them with dashes then?

PR #144, opened 2019-08-19
fix(cactus-web): adjust accordion height when open

JamesNimlos:
Fixes the issue where the height of the accordion body wouldn't grow with it's content. Also implements a dynamic length of animation based on how far the animation is. If the accordion is already open it no longer animates.

### DIFF
--- a/modules/cactus-web/src/Accordion/Accordion.test.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.test.tsx
@@ -52,7 +52,7 @@ describe('component: Accordion', () => {
   })
 
   describe('Provider', () => {
-    test('Should manage Accordion state', async () => {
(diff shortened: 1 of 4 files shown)


modules/cactus-web/src/Accordion/Accordion.test.tsx
```diff
@@ -37,15 +46,13 @@ describe('component: Accordion', () => {
...
-        jest.runAllTimers()
       })
       expect(container).toHaveTextContent('Test Body')
-      expect(container).toMatchSnapshot()
     })
   })
 
   describe('Provider', () => {
-    test('Should manage Accordion state', () => {
+    test('Should manage Accordion state', async () => {
```
Dhalton: This shouldn't need to be an async function since we're not going with the new async `act()`, right?

PR #149, opened 2019-08-29
Upgrade styled system

JamesNimlos:
Also includes: style prop features, perf upgrades, security patching


modules/cactus-icons/scripts/svgr.js
```diff
@@ -8,13 +8,22 @@ const prettier = require('prettier')
...
   try {
     await fs.access(path.resolve(__dirname, '../ts'))
   } catch {
     await fs.mkdir(path.resolve(__dirname, '../ts'))
   }
 
+  await fg(['./src/**/*.ts']).then(extraSources =>
+    Promise.all(
+      extraSources.map(f =>
+        fs.copyFile(path.join(CWD, f), path.join(CWD, 'ts', f.replace('./src/', '')))
```
Dhalton: Just trying to understand what's going on here...so this is finding src files like `iconSizes.ts` & `types.ts` and copying them under a `ts` folder? What's this for?

JamesNimlos: That's all it is doing; so that iconSizes can be used by all the Icons instead of all creating their own and then exported to be used by IconButton as well

PR #158, opened 2019-09-18
test(cactus-web): Add integration tests to our test app

Dhalton:
Adds integration tests for our form elements to the mock-ebpp test app. Admittedly, these tests aren't as fleshed out as I'd like them to be, but I've been working on this for several days and I'd like to get at least this into the pipeline and get back to working on Channels, and this is at least a good starting point.

### DIFF
--- a/examples/mock-ebpp/tsconfig.json
+++ b/examples/mock-ebpp/tsconfig.json
@@ -15,5 +15,5 @@
     "noEmit": true,
     "jsx": "preserve"
   },
-  "include": ["src", "node_modules/@types/puppeteer/index.d.ts"]
+  "include": ["src", "puppeteer"]
 }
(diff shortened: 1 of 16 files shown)


examples/mock-ebpp/tsconfig.json
```diff
@@ -15,5 +15,5 @@
     "noEmit": true,
     "jsx": "preserve"
   },
-  "include": ["src"]
+  "include": ["src", "node_modules/@types/puppeteer/index.d.ts"]
```
JamesNimlos: does just "puppeteer" not work for this?

Dhalton: let me give it a try

Dhalton: Looks like that did work!

PR #163, opened 2019-10-02
Danger Buttons

Dhalton:
This adds the styles for the danger variants for `Button` and `IconButton`. We already have a danger variant for `TextButton` implemented.

I didn't get any designs for the inverse versions of these variants,  but I just tried to follow the same patterns we've done for the other ones. If we end up using them later on down the line, I'm sure Mariana will stub out some designs for them.

I also fixed a bug in the `IconButton` story that was caused by trying to render all values inside `icons`. Apparently, the last value is a parse function and not an icon, so I sliced it out before rendering.

### DIFF
--- a/modules/cactus-theme/index.ts
+++ b/modules/cactus-theme/index.ts
@@ -76,6 +76,7 @@ export interface CactusTheme {
     transparentSuccess: string
     transparentWarning: string
     transparentError: string
+    errorDark: string
 
     status: StatusColors
   }
@@ -141,6 +142,7 @@ export function generateTheme({ primaryHue }: GeneratorOptions = repayOptions): 
   let transparentSuccess = `hsla(145, 89%, 28%, 0.2)`
   let transparentError = `hsla(353, 84%, 44%, 0.2)`
   let transparentWarning = `hsla(47, 82%, 47%, 0.2)`
+  let errorDark = `hsl(353, 96%, 11%)`
 
   const status: StatusColors = {
     background: {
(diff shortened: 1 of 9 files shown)


modules/cactus-theme/index.ts
```diff
@@ -141,6 +142,7 @@ export function generateTheme({ primaryHue }: GeneratorOptions = repayOptions):
   let transparentSuccess = `hsla(145, 89%, 28%, 0.2)`
   let transparentError = `hsla(353, 84%, 44%, 0.2)`
   let transparentWarning = `hsla(47, 82%, 47%, 0.2)`
+  let errorDark = `hsl(353, 96%, 11%)`
```
JamesNimlos: Long term, I think we'll move to save the error hue as a variable from which to build these error colors. For now, this works well.

modules/cactus-web/src/IconButton/IconButton.story.tsx
```diff
@@ -34,13 +34,15 @@ storiesOf('IconButton', module)
...
       <Grid justify="center">
-        {Object.values(icons).map(Icon => (
-          <Grid.Item tiny={3} medium={2} large={1}>
-            <IconButton variant={variantSelection}>
-              <Icon />
-            </IconButton>
-          </Grid.Item>
-        ))}
+        {Object.values(icons)
+          .slice(0, Object.keys(icons).length - 2)
```
JamesNimlos: Part of me wants to remove iconSizes from the main export file since it's been so problematic 😖

JamesNimlos:
Great job breaking this up into commits for the theme and web so it's easier for the release log 💯

PR #165, opened 2019-10-08
feat(cactus-web): Add more mobile optimized select dropdown

Dhalton:
Keeps multiselect open when you click options, adds mobile version of the select's dropdown. Honestly, I'm not very happy with the way I ended up having to do the fade in/out on the dropdown, but I've been trying anything I can think of for days now and there just doesn't seem to be a good way to make the gradients appear only when you have somewhere to scroll, so I had to just add some extra vertical padding. I also don't think we can test this outside integration tests since we don't have a viewport for the unit tests. Anyway, I'm open to any feedback on this so let me know what you think.

### DIFF
--- a/modules/cactus-web/src/StyleProvider/StyleProvider.tsx
+++ b/modules/cactus-web/src/StyleProvider/StyleProvider.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 
 import * as styledComponents from 'styled-components'
+import { breakpoints } from '../helpers/constants'
 import { Omit } from '../types'
 import cactusTheme, { CactusTheme } from '@repay/cactus-theme'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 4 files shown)


modules/cactus-web/src/StyleProvider/StyleProvider.tsx
```diff
@@ -89,7 +88,7 @@ export const StyleProvider: React.FC<StyleProviderProps> = props => {
   const { global, children, theme: providedTheme, ...themeProviderProps } = props
   shouldCheckTheme && checkThemeProperties(providedTheme || cactusTheme)
   const theme = providedTheme ? providedTheme : cactusTheme
-  theme.breakpoints = breakpoints.slice()
+  theme.breakpoints = Object.values(breakpoints).map(bp => bp.toString())
```
JamesNimlos: We should create this array as a constant outside the render, and then because we support IE11 we can't guarantee iteration order so we should generate it from an array of the keys in the order we want.

PR #168, opened 2019-10-22
Adding DateInput and DateInputField

JamesNimlos:


modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,1558 @@
...
+interface DateInputState {
+  value: PartialDate
+  locale: string
+  isOpen: false | 'month-year' | 'calendar'
+  // YYYY-MM-dd
+  focusDay?: string
+}
+
+/**
+ * TODO
```
Dhalton: I think we can get rid of this TODO

PR #170, opened 2019-11-01
feat(cactus-web): Add combobox functionality to select

Dhalton:
Adds the option to render a select as a combobox


examples/mock-ebpp/tests/form-elements.test.ts
```diff
@@ -49,39 +49,73 @@ describe('UI Config Form', () => {
     await server.close()
   })
 
-  it('should be titled "UI Config"', async () => {
+  test('should be testled "UI Config"', async () => {
```
JamesNimlos: lol, "testled"

examples/mock-ebpp/tests/helpers/form-actions.ts
```diff
@@ -1,49 +1,99 @@
...
+    if (typeof optionOrOptions === 'string') {
+      const option = await getByText(listbox, optionOrOptions)
+      await option.tap()
+    } else if (Array.isArray(optionOrOptions)) {
+      for (let i = 0; i < optionOrOptions.length; i++) {
+        const option = await getByText(listbox, optionOrOptions[i])
+        await option.tap()
+      }
+      // haxxors
+      await this.page.keyboard.press('Escape')
```
JamesNimlos: Maybe we should be clicking our own "Done" button instead actually?

JamesNimlos:
That last commit could actually be referenced as a fix in cactus-web since the dialog should have had that role all along

Dhalton:
> That last commit could actually be referenced as a fix in cactus-web since the dialog should have had that role all along

You're right, let me reset the commit and separate them

PR #171, opened 2019-11-04
chore(): adds release and changelog script

JamesNimlos:
Now you can execute `yarn release [module] [new version]` at the root and it will run you through the script to clean, build, publish, and commit the new release; then it will generate and commit the updated change log.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -8,6 +8,7 @@ const exec = util.promisify(require('child_process').exec)
 const prompts = require('prompts')
 
 /**
+ * The parsed commit from conventional-commits-parser
  * type Commit = {
  *   type: string
  *   scope: null | string
@@ -157,7 +158,6 @@ async function main() {
             commit.header.includes('publish') &&
             commit.header.includes(scope)
         ) + 1
-    // const lastReleaseCommit = all[lastReleaseCommitIndex]
     let relevantCommits = all.slice(0, lastReleaseCommitIndex)
     for (const commit of relevantCommits) {
       let { scope: commitScope } = normalizeScope(commit.scope)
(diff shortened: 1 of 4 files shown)


scripts/release.js
```diff
@@ -0,0 +1,232 @@
...
+const path = require('path')
+const gitRawCommits = require('git-raw-commits')
+const commitsParser = require('conventional-commits-parser')
+const util = require('util')
+const { spawnSync } = require('child_process')
+const exec = util.promisify(require('child_process').exec)
+const prompts = require('prompts')
+
+/**
+ * type Commit = {
```
Dhalton: Do we need these commented out types?

scripts/release.js
```diff
@@ -0,0 +1,232 @@
...
+    const lastReleaseCommitIndex =
+      all
+        .slice(1)
+        .findIndex(
+          commit =>
+            commit.type === 'chore' &&
+            commit.header.includes('publish') &&
+            commit.header.includes(scope)
+        ) + 1
+    // const lastReleaseCommit = all[lastReleaseCommitIndex]
```
Dhalton: Looks like we don't need this line

scripts/release.js
```diff
@@ -0,0 +1,232 @@
...
+
+    content = generateReleaseHeader(pkg, releaseVersion, all[0]) + content + '\n'
+
+    let where = CHANGELOG.indexOf('## ')
+    CHANGELOG = CHANGELOG.substring(0, where) + content + CHANGELOG.substring(where)
+
+    await fsp.writeFile(path.join(process.cwd(), 'CHANGELOG.md'), CHANGELOG)
+
+    await exec(`git add CHANGELOG.md`)
+    let gitCommitResult = await exec(`git commit -m "chore(): update changelog"`)
```
Dhalton: What do you think about including the module name here? I think it would be something like this:
```await exec(`git commit -m "chore(${scope}): update changelog"`)```. Because I'm pretty sure I usually include the module name in these commits

PR #173, opened 2019-11-05
fix(cactus-web): Detect values that need to be added to options

Dhalton:
Detects when values are passed that don't exist in options, and then adds them

### DIFF
--- a/modules/cactus-web/src/Select/Select.test.tsx
+++ b/modules/cactus-web/src/Select/Select.test.tsx
@@ -1024,6 +1024,36 @@ describe('component: Select', () => {
       trigger = getByRole('button')
       expect(document.activeElement).toBe(trigger)
     })
+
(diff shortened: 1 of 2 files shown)


modules/cactus-web/src/Select/Select.test.tsx
```diff
@@ -1024,6 +1024,36 @@ describe('component: Select', () => {
...
+            name="city"
+            options={['phoenix', 'tucson', 'flagstaff']}
+            value="superior"
+            comboBox
+          />
+        </StyleProvider>
+      )
+      await animationRender()
+      expect(getByRole('listbox')).toHaveTextContent('superior')
+    })
```
JamesNimlos: Can you add another test where you set the value to different things in consecutive renders and then confirm that both of them are in the listbox?

PR #177, opened 2019-11-08
Dedupe @reach/rect by upgrading @reach/tooltip

Dhalton:
Upgrades @reach dependencies and modifies affected components so they work with the newest versions


modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -33,34 +29,22 @@ interface TooltipProps
...
+    top: triggerRect.top + triggerRect.height + scrollY,
+    right: 0,
+    bottom: 0,
+    height: 0,
+    width: 0,
+    x: 0,
+    y: 0,
+    toJSON: () => {
+      return null
+    },
```
JamesNimlos: Let's override this def with `as unknown as DOMRect` or something similar instead of adding code to fix the incompatibilities.

PR #179, opened 2019-11-12
fix(cactus-web): Fix how we decide to make an option the create option

Dhalton:
When I built the combobox, I looked at the ids on the options to determine if the id contained "create", in which case I gave it the "create" `data-role`. However, option ids are generated in part using the option's label, so this breaks if the option contains the string "create".  /facepalm

This fixes that issue, though, and I already tested it locally on channels

### DIFF
--- a/modules/cactus-web/src/Select/Select.tsx
+++ b/modules/cactus-web/src/Select/Select.tsx
@@ -763,6 +763,7 @@ class List extends React.Component<ListProps, ListState> {
                     let isSelected = opt.isSelected
                     let ariaSelected: boolean | 'true' | 'false' | undefined =
                       isSelected || undefined
+                    let isCreateNewOption = comboBox && optId === `create-${this.state.searchValue}`
                     // multiselectable should have aria-select©ed on all options
                     if (multiple) {
                       ariaSelected = isSelected ? 'true' : 'false'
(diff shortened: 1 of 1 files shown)


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -774,13 +775,13 @@ class List extends React.Component<ListProps, ListState> {
...
-                        data-role={optId.includes('create') ? 'create' : 'option'}
+                        data-role={isCreateNewOption ? 'create' : 'option'}
                         aria-selected={ariaSelected}
                         onMouseEnter={this.handleOptionMouseEnter}
                       >
-                        {optId.includes('create') ? (
+                        {isCreateNewOption ? (
                           <ActionsAdd mr={2} mb={2} />
-                        ) : multiple && !optId.includes('create') ? (
+                        ) : multiple && !isCreateNewOption ? (
```
abunsen: why are you checking for `!isCreateNewOption` isn't that already the case since we're after the `:` in the ternary operator?

Dhalton: Yeah you're right was probably just being extra careful when I wrote this and then just didn't think about it when I was switching them. Will update

Dhalton: Updated

PR #180, opened 2019-11-13
Accordion Accessibility

Dhalton:
Updates the accordion to adhere to the accessibility standards defined here: https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html

It's a breaking change because I removed the `openLabel` & `closeLabel` props since they aren't necessary if you're using `aria-expanded` correctly.

~Had to populate the provider's state with all of the accordions & their ids using a somewhat hacky solution, so I'm totally hoping you have a better idea for how we can do that. If not, this does seem to work as far as I can tell.~

Storing the ids of the accordions in an array using a ref worked great, and I applied `tabIndex={0}` to the body in the component if `children` is a string. I also refactored so that we generate the labels for the header and body inside the accordion and pass them along using the context

11/18/2019 update: I was having problems with the integration tests when I added another test file, so I renamed the file we've been using from `form-elements.test.ts` to `integration.test.ts`. Admittedly I don't like that all of the integration tests are in the one file but I couldn't fix the issue when I had multiple test files, so maybe we can work that out later on.


modules/cactus-web/src/Accordion/Accordion.mdx
```diff
@@ -9,10 +9,6 @@ import cactusTheme from '@repay/cactus-theme'
 
 # Accordion
 
-## Best practices
-
-- Use descriptive labels for the `openLabel` and `closeLabel` props on `Accordion.Header`. These labels are used for keyboard/screen-reader accessibility.
```
JamesNimlos: I think it would be best to mention here that it's an anti-pattern to have a series of accordions alongside each other which are not wrapped in an Accordion.Provider because they will not have the connection and keyboard functionality.

examples/mock-ebpp/tests/integration.test.ts
```diff
@@ -0,0 +1,314 @@
...
+
+async function getActiveAccessibility(page: puppeteer.Page) {
+  await sleep(0.2)
+  let activeElHandle = await getActiveElement(page)
+  let activeEl = await activeElHandle.asElement()
+  if (activeEl === null) {
+    return
+  }
+  return page.accessibility.snapshot({ root: activeEl })
+}
```
JamesNimlos: This function could maybe be an action now to make it universalized.

modules/cactus-web/src/Accordion/Accordion.test.tsx
```diff
@@ -18,10 +19,10 @@ afterEach(() => {
 
 describe('component: Accordion', () => {
   describe('Component', () => {
-    test('Should render Accordion', () => {
+    test.skip('Should render Accordion', () => {
```
JamesNimlos: 👾

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -39,33 +41,98 @@ interface AccordionProviderContext {
...
+    }
+  }
 
   return (
-    <div className={className} tabIndex={0} aria-label={ariaLabel}>
-      <header>{children}</header>
-      <IconButton onClick={handleToggle} label={buttonLabel} iconSize="small" mx="16px">
-        {isOpen ? <NavigationChevronDown /> : <NavigationChevronLeft />}
-      </IconButton>
-    </div>
```
JamesNimlos: Funny how we are probably going to go back to the original way for this with your next updates.

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -208,6 +288,39 @@ export const AccordionProvider = (props: AccordionProviderProps) => {
...
+  }
+
+  const focusLast = () => {
+    if (managedAccordions.current.length > 1) {
+      focusById(managedAccordions.current[managedAccordions.current.length - 1])
+    }
+  }
+
+  const focusById = (id: string) => {
+    const focusAccordion = document.querySelector(`[data-id="${id}"]`)
```
JamesNimlos:
```suggestion
    const focusAccordion = document.querySelector(`#${id}`)
```
You should be able to query id directly with just #, or was there a different purpose for adding it also as a data-* attribute?

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -243,32 +364,81 @@ interface AccordionState {
...
+    if (isManaged && typeof focusLast === 'function') {
+      focusLast()
     }
   }
 
   const rest = omitMargins(props, 'width', 'maxWidth')
 
   return (
-    <div {...rest}>
+    <div id={props.id || id} data-id={id} {...rest}>
```
JamesNimlos:
```suggestion
    <div id={props.id || id} {...rest}>
```
see comment about querying id directly with querySelector

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -208,6 +288,39 @@ export const AccordionProvider = (props: AccordionProviderProps) => {
...
+      const currentIndex = managedAccordions.current.findIndex(accordionId => accordionId === id)
+      let nextIndex = currentIndex + offset
+      if (nextIndex < 0) {
+        nextIndex = managedAccordions.current.length - 1
+      } else if (nextIndex >= managedAccordions.current.length) {
+        nextIndex = 0
+      }
+      focusById(managedAccordions.current[nextIndex])
+    }
+  }
```
JamesNimlos:
```tsx
const manageFocus = (id: string, offset: number) => {
     if (managedAccordions.current.length > 1) {
      const orderedIds = Array.from(document.querySelectorAll('#' + managedAccordions.current.join(',#'))).map(el => el.id)
       const currentIndex = orderedIds.indexOf(id)
       let nextIndex = currentIndex + offset
       if (nextIndex < 0) {
         nextIndex = orderedIds.length - 1
       } else if (nextIndex >= orderedIds.length) {
         nextIndex = 0
       }
       focusById(orderedIds[nextIndex])
     }
   }
```

modules/cactus-web/src/Accordion/Accordion.tsx
Dhalton: Yeah kind of sucks but just the way it goes sometimes! 😛

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -208,6 +288,47 @@ export const AccordionProvider = (props: AccordionProviderProps) => {
...
+    if (managedAccordions.current.length > 1) {
+      const orderedIds = getOrderedIds()
+      focusById(orderedIds[orderedIds.length - 1])
+    }
+  }
+
+  const focusById = (id: string) => {
+    const focusAccordion = document.getElementById(id)
+    if (focusAccordion && focusAccordion.children[0]) {
+      const toFocus = focusAccordion.children[0] as HTMLElement
```
JamesNimlos: I'd recommend using a data-* attribute to select this element to focus instead. It'll be more stable since we aren't restricting exactly how these items are rendered.

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -243,32 +372,81 @@ interface AccordionState {
...
+    accordions,
+    manageOpen,
+    manageFocus,
+    focusFirst,
+    focusLast,
+    isManaged,
+    managedAccordions,
+  } = useContext(ProviderContext)
   const id = useId()
+  const headerId = props.id ? `${props.id}-header` : `${id}-header`
```
JamesNimlos: isn't one of the arguments to `useId` the id so if it's provided it just uses that?

Dhalton: Ah yeah you're right duh. Updating

PR #183, opened 2019-11-21
feat(cactus-web): Allow user to initially render accordions open

Dhalton:
Allows devs to initially render an open accordion using the `open` prop.

To do this, I had to refactor the structure of the `managedAccordions` ref that I added in my last PR, and use that for both focus management (same as before) and for managing whether or not accordions are open (new behavior). Previously, I was using the `useState` hook in the provider to manage whether or not the accordions were open, but now I need that information to be available from the first render. 

Let me know if you have any questions/comments!

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1767,6 +1767,22 @@
     "@babel/preset-typescript" "^7.3.3"
     babel-plugin-dynamic-import-node "^2.3.0"
 
+"@repay/cactus-web@^0.5.2":
+  version "0.5.8"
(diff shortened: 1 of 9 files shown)


yarn.lock
```diff
@@ -1767,6 +1767,22 @@
     "@babel/preset-typescript" "^7.3.3"
     babel-plugin-dynamic-import-node "^2.3.0"
 
+"@repay/cactus-web@^0.5.2":
```
JamesNimlos: Can you fix this everywhere?

PR #187, opened 2019-11-25
fix(cactus-web): Provide more space for elements in accordion header

Dhalton:
Converted the accordion headers to `div` rather than `button` so we can include whatever kinds of headers we want as children. This is a breaking change because, previously, the accordion headers automatically had h3 styles, whereas now you must explicitly render your accordion header with whatever text styles you want

### DIFF
--- a/modules/cactus-web/src/Accordion/Accordion.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.tsx
@@ -5,6 +5,7 @@ import { CactusTheme } from '@repay/cactus-theme'
 import { margin, MarginProps, maxWidth, MaxWidthProps, width, WidthProps } from 'styled-system'
 import { NavigationChevronDown, NavigationChevronLeft } from '@repay/cactus-icons'
 import { omitMargins } from '../helpers/omit'
+import IconButton from '../IconButton/IconButton'
 import KeyCodes from '../helpers/keyCodes'
 import PropTypes from 'prop-types'
 import Rect from '@reach/rect'
@@ -20,7 +21,7 @@ interface AccordionProps
 }
 
 interface AccordionHeaderProps
-  extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLButtonElement>, HTMLButtonElement> {}
+  extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> {}
 
 interface AccordionBodyProps
   extends MarginProps,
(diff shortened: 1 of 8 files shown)


modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -130,7 +130,7 @@ const AccordionHeaderBase = (props: AccordionHeaderProps) => {
       aria-expanded={isOpen}
       aria-controls={bodyId}
     >
-      <span>{children}</span>
+      <span style={{ width: '100%', textAlign: 'left' }}>{children}</span>
```
JamesNimlos: I thought we were going to just make this a `div` instead?

Dhalton: Yeah I'll rework this on Monday. I'm going to close this until then. I'll convert it to a `div` and remove the enforced h3 styles

modules/cactus-web/src/IconButton/IconButton.tsx
```diff
@@ -132,7 +132,31 @@ IconButton.propTypes = {
...
   variant: PropTypes.oneOf(['standard', 'action', 'danger']),
   disabled: PropTypes.bool,
-  label: PropTypes.string.isRequired,
+  label: (props: IconButtonProps, propName: string, componentName: string) => {
+    if (!props.label && !props['aria-labelledby']) {
+      return new Error(
+        `One of props 'label' or 'aria-labelledby' was not specified in ${componentName}.`
+      )
+    } else if (props.label !== undefined && typeof props.label !== 'string') {
+      console.log(props.label)
```
JamesNimlos: 👾

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -117,26 +118,23 @@ const AccordionHeaderBase = (props: AccordionHeaderProps) => {
...
+        onKeyDown={handleHeaderKeyDown}
+        onKeyUp={handleHeaderKeyUp}
+        data-role="accordion-button"
+        type="button"
+        role="button"
+        aria-expanded={isOpen}
+        aria-controls={bodyId}
+        aria-labelledby={headerId}
+      >
+        {isOpen ? <NavigationChevronDown /> : <NavigationChevronLeft />}
```
JamesNimlos: Let's add `aria-hidden="true"` onto these icons.

PR #190, opened 2019-12-03
Feature and fix to finish the form elements

JamesNimlos:
* feat(cactus-web): allow refs on TextInput for focus control
* fix(cactus-web): stabilize time on large year changes in DateInput

### DIFF
--- a/modules/cactus-web/src/TextInput/TextInput.tsx
+++ b/modules/cactus-web/src/TextInput/TextInput.tsx
@@ -47,15 +47,17 @@ const displayStatus = (props: TextInputProps) => {
   }
 }
 
-const TextInputBase = (props: TextInputProps) => {
-  const { className, ...rest } = omitMargins(props)
+const TextInputBase = React.forwardRef<HTMLInputElement, TextInputProps>(
+  (props: TextInputProps, forwardRef) => {
+    const { className, ...rest } = omitMargins(props)
 
-  return (
-    <div className={className}>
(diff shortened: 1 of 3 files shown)


modules/cactus-web/src/TextInput/TextInput.tsx
```diff
@@ -47,15 +47,17 @@ const displayStatus = (props: TextInputProps) => {
...
-const TextInputBase = (props: TextInputProps) => {
-  const { className, ...rest } = omitMargins(props)
-
-  return (
-    <div className={className}>
-      <Input {...rest} />
-    </div>
-  )
-}
+const TextInputBase = React.forwardRef<HTMLInputElement, TextInputProps>(
```
Dhalton: Looks good just curious why you made this a `forwardRef` component?

Dhalton: I mean, I can see from your commit message that it was done so you can control the focus, I just don't see that being done anywhere

JamesNimlos: It's to use it with an [autosuggest library](https://github.com/moroshko/react-autosuggest) in EBPP

JamesNimlos: I'll point it out when I make the PR, but it was using a normal `input[type=text]` before so not even a semantic-ui component

PR #193, opened 2019-12-10
fix(cactus-web): firefox differences in timezone offsets

JamesNimlos:
Timezones are slightly different in Firefox vs Safari/Chrome so in order to prevent the time slippage we need to just let the UTC time be different in different browsers ☹️

### DIFF
--- a/modules/cactus-web/src/helpers/dates.ts
+++ b/modules/cactus-web/src/helpers/dates.ts
@@ -588,9 +588,8 @@ export class PartialDate implements FormatTokenMap {
 
   toDate(): Date {
     let date = new Date()
-    const modernOffset = date.getTimezoneOffset()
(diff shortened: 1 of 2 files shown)


wilysword:
Is there a way to indicate that it shouldn't convert to UTC before sending to the API (or before being used however)? For example in the webhook history filters it makes sense to convert to UTC, but in the Create Scheduled Task page the datetime is supposed to be naive, literally sending whatever is entered regardless of the local time zone.

JamesNimlos:
@wilysword Yeah, by default it keeps everything as timezone less strings but I set it to use `Date` when it's a datetime type in Channels. It is a more secure way to keep it stable though so I can change that default and have only the DatetimeRange use the Date type.

PR #197, opened 2019-12-20
fix(cactus-web): Fix buttons in header toggling accordion

Dhalton:
Quick bug fix to restrict the accordion from toggling its open/closed state based on what target was clicked.


JamesNimlos:
I think we should use a while loop and `element = element.parentElement` to make sure whatever was clicked was not a button or descendant of a button instead of specifically singling out IconButtons. The problem is if we ever put another svg in a header it cannot be clicked to toggle the accordion

Dhalton:
> I think we should use a while loop and `element = element.parentElement` to make sure whatever was clicked was not a button or descendant of a button instead of specifically singling out IconButtons. The problem is if we ever put another svg in a header it cannot be clicked to toggle the accordion

Good idea, I'll do that!

PR #203, opened 2020-01-28
I18n Upgrades

JamesNimlos:
* docs(examples): use more idiomatic fluent translations
* feat(cactus-i18n): merge sections into a single bundle per language
* feat(cactus-i18n): only load supported languages
* feat(cactus-i18n): make i18n library platform agnostic

BREAKING: 
* all keys in a given section must be prefixed with the pattern:
`<section>__<id>` so that there are no conflicts when merging the resources into
the single bundle.
* The I18nProvider will no longer use navigator.language internally to
ensure cross platform usage. Therefore, any DOM usage should provide the
navigator.language as the initial value for I18nProvider lang prop.

### DIFF
--- a/docs/Internationalization/API Documentation.md
+++ b/docs/Internationalization/API Documentation.md
@@ -96,6 +96,29 @@ return (
 )
 ```
 
+In the browser, it is recommended to use the user provided language by default but also allow the user to change as desired.
+
+```jsx
+// App.js
+import React, { useState } from 'react'
+import I18nProvider from '@repay/cactus-i18n'
+import i18nController from './I18nController'
+
+function App(props) {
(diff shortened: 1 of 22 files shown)


Dhalton:
Everything looks solid as far as I can tell, I'm just not sure why the CI pipeline isn't running on this...

JamesNimlos:
Is it because the branch has `wip` in the name?

Dhalton:
> Is it because the branch has `wip` in the name?

It looks like this branch was just so old that it still had the circleci config, not the GHA one. I rebased this onto master and that triggered the build. Unfortunately, that also associated me with your commits 😦

PR #208, opened 2020-03-03
feat(cactus-theme): adds option to generate theme from hex colors

JamesNimlos:


modules/cactus-theme/src/theme.ts
```diff
@@ -0,0 +1,613 @@
...
+  let transparentCTA = `hsla(${primaryHue}, 96%, 35%, 0.25)`
+
+  /** Contrasts */
+  let lightContrast = `hsl(${primaryHue}, 29%, 90%)`
+  let mediumContrast = `hsl(${primaryHue}, 18%, 45%)`
+  let darkContrast = `hsl(${primaryHue}, 9%, 35%)`
+  let darkestContrast = `hsl(${primaryHue}, 10%, 20%)`
+
+  /** Neutrals */
+  let white = `hsl(0, 0%, 100%)`
```
Dhalton: I know you were a little bit concerned about having so much duplicate code in this file. Couldn't we move constant color definitions like this up to the module level since they don't depend on the hue or the primary/secondary hex values? It's not a big deal but it could clean up at least a little bit of the duplication

JamesNimlos: Oh right, because they're in both functions. I'll do that

PR #209, opened 2020-03-11
feat(cactus-i18n): restrict logging to debugMode only

JamesNimlos:
* removes all logging when not in debug mode
* allows extra properties to be passed via `I18nSection` to the controller.load function
* allows adding dependencies to an `I18nSection`

[CACTUS-232]
[CACTUS-233]

[CACTUS-232]: https://repayonline.atlassian.net/browse/CACTUS-232
[CACTUS-233]: https://repayonline.atlassian.net/browse/CACTUS-233

### DIFF
--- a/modules/cactus-i18n/src/BaseI18nController.ts
+++ b/modules/cactus-i18n/src/BaseI18nController.ts
@@ -111,14 +111,14 @@ export default abstract class BaseI18nController {
       )
     }
     const bundle = this._getDict()[lang]
-    if (!bundle) {
+    if (this._debugMode && !bundle) {
       console.error(
         `Attempting to set dictionary for unsupported language: ${lang} and section: ${section}`
       )
       return
     }
     const errors = bundle.addResource(new FluentResource(ftl))
-    if (Array.isArray(errors) && errors.length) {
+    if (this._debugMode && Array.isArray(errors) && errors.length) {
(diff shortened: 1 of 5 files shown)


modules/cactus-i18n/src/BaseI18nController.ts
```diff
@@ -234,6 +240,24 @@ export default abstract class BaseI18nController {
       })
   }
 
+  hasText({
```
Dhalton: Was a hasText function something that was asked for by the Channels guys? Just curious what their use-case is

Dhalton:
What do you think about adding an extra test where you pass a dependency as an object with its own "props"?

modules/cactus-i18n/src/BaseI18nController.ts
JamesNimlos: Yeah, we're going to use it to decide whether to render the I18nText component or pull text from a config

Dhalton: Ah, right okay that makes sense. Thanks!

JamesNimlos:
@Dhalton I've added the test you mentioned and I've also split the tests to run in different steps in the CI so hopefully that mitigates some issues we've seen with inconsistent test runs.

PR #214, opened 2020-03-19
feat(cactus-web): Added rendered components to docs pages. [CACTUS-236]

NicolasSimmonds:


modules/cactus-web/src/CheckBoxField/CheckBoxField.mdx
```diff
@@ -5,12 +5,20 @@ menu: Components
...
 import cactusTheme from '@repay/cactus-theme'
 
 # CheckBoxField
 
 The `CheckBoxField` is a combination of `CheckBox` and `Label`, along with some accessibility features that are taken care of for you.
 
+### How it looks 
+
+<CheckBoxField name="checkboxfield" label="Is Checked"/>
+<Button type="submit" variant="action">
```
Dhalton: Why the submit button here?

modules/cactus-web/src/FileInput/FileInput.mdx
```diff
@@ -11,6 +11,21 @@ import cactusTheme from '@repay/cactus-theme'
...
+<FileInput
+  name="file-input"
+  accept={['.txt', '.doc']}
+  labels={{ delete: 'delete file', retry: 'retry file upload', loading: 'loading', loaded: 'successful' }}
+  prompt="Drag files here or"
+  buttonText="Select Files..."
+  rawFiles={false}
+  multiple={true}
+/>
+<Button type="submit" variant="action">
```
Dhalton: Same thing here, what's the need for the button?

modules/cactus-web/src/Grid/Grid.mdx
```diff
@@ -5,30 +5,53 @@ menu: Components
...
 import cactusTheme from '@repay/cactus-theme'
 
 # Grid
 
 A responsive grid component designed to help lay out items on a page based on screen size breakpoints.
 
+### How it looks
+
+<Grid>
+  <Grid.Item tiny={3} large={3}>
```
Dhalton: I think it might be useful to use different values on the `tiny` and `large` props here. That way someone looking at the docs could resize the screen to see the effect. Maybe we could add the `medium` prop too. Let's try setting tiny to 2, medium to 4, and large to 6 on all of these `Grid.Item`s. And then maybe include a little message telling the viewer to try resizing the screen to see the grid in action.

modules/cactus-web/src/IconButton/IconButton.mdx
```diff
@@ -6,11 +6,19 @@ menu: Components
 import PropsTable from 'website-src/components/PropsTable'
 import IconButton from './IconButton'
 import cactusTheme from '@repay/cactus-theme'
+import { ActionsAdd }from '@repay/cactus-icons'
+console.log("icons", Icons)
```
Dhalton: We should remove this console log. I think this would cause an error anyway, since `Icons` doesn't look like it's defined. Also, on the line above, we're missing a space.

modules/cactus-web/src/StepAvatar/StepAvatar.mdx
```diff
@@ -11,6 +11,10 @@ import cactusTheme from '@repay/cactus-theme'
 
 The `Step Avatar` is an Avatar used to convey the amount of steps in a task.
 
+### How it looks
+
+<StepAvatar status="inProcess">3</StepAvatar>
```
Dhalton: Let's try adding a few of these step avatars next to each other. Maybe steps 1, 2, and 3. And we can say step 2's status is `inProcess`

modules/cactus-web/src/TextArea/TextArea.mdx
```diff
@@ -6,9 +6,16 @@ menu: Components
...
 import TextArea from './TextArea'
 import cactusTheme from '@repay/cactus-theme'
+import Label from '../Label'
+import Button from '../Button'
 
 # TextArea
 
+### How it looks
+
+<Label htmlFor="my-id">Enter some text</Label>
```
Dhalton: Let's remove this label so the differences between `TextArea` and `TextAreaField` are a little more obvious

modules/cactus-web/src/Toggle/Toggle.mdx
```diff
@@ -5,12 +5,27 @@ menu: Components
...
+
 # Toggle
 
 The `Toggle` component can be useful for boolean fields.
 
+### How it loooks
+
+<Grid>
+  <Grid.Item tiny={12}>
+    <Toggle value={true} />
```
Dhalton: What do you think about maybe instead of rendering the `true` and `false` versions separately, we render 1 toggle inside of a component that can manage the toggle's state? Then it'd be a little more interactive for the user. I can help you with this is you're unsure of how to do it

modules/cactus-web/src/ToggleField/ToggleField.mdx
```diff
@@ -9,6 +9,19 @@ import cactusTheme from '@repay/cactus-theme'
 
 # ToggleField
 
+### How it looks
+
+<ToggleField
```
Dhalton: Same thing for these toggle fields

modules/cactus-web/src/Tooltip/Tooltip.mdx
```diff
@@ -5,10 +5,20 @@ menu: Components
...
 import cactusTheme from '@repay/cactus-theme'
 
 # Tooltip
 
+
+### How it loooks
+
+<div>
+  <Label>Custom Input</Label>
+  <Tooltip label="Type something!" />
```
Dhalton: Let's just render the tooltip by itself here

modules/cactus-web/src/StepAvatar/StepAvatar.mdx
```diff
@@ -11,6 +11,12 @@ import cactusTheme from '@repay/cactus-theme'
 
 The `Step Avatar` is an Avatar used to convey the amount of steps in a task.
 
+### How it looks
+
+<StepAvatar status="notDone" margin={3}>1</StepAvatar>
```
Dhalton: Would you mind switching the statuses for step 1 and 3 so that 1 is "done" and 3 is "notDone"? That's how these would behave in a real-world setting

modules/cactus-web/src/Toggle/Toggle.mdx
```diff
@@ -6,11 +6,22 @@ menu: Components
...
 import cactusTheme from '@repay/cactus-theme'
+import React, { useState } from 'react'
 
 # Toggle
 
 The `Toggle` component can be useful for boolean fields.
 
+### How it loooks
+
+export const Wrapper = () => {
```
Dhalton: Excellent!

PR #219, opened 2020-04-16
Breadcrumb component

NicolasSimmonds:


modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,75 @@
...
+const BreadCrumbBase = (props: BreadcrumbProps) => {
+  const { ariaLabel, children, className } = props
+  return (
+    <nav aria-label={ariaLabel} className={className}>
+      <ul>{children}</ul>
+    </nav>
+  )
+}
+
+interface BreacdrumbComponent extends StyledComponentBase<'ul', CactusTheme, BreadcrumbProps> {
```
Dhalton, line 39: Typo in the interface name

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,75 @@
...
+interface BreadcrumbItemProps {
+  label: string
+  linkTo: string
+  className?: string
+  active?: boolean
+}
+interface BreadcrumbProps {
+  children?: React.ReactNode
+  className?: string
+  ariaLabel?: string
```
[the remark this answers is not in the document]
NicolasSimmonds: Ok got it, but should we remove the aria for the wrapper. In the documentation is encouraged to write an aria-label for the nav tag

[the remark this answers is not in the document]
Dhalton: I was just looking at the docs (assuming you're referring to [these](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html) docs) and it looks like the label on the `nav` should always be set to "Breadcrumb". I'm guessing that provides the screen-reader with knowledge that there should be child links there in the form of a breadcrumb component.

So we should add one to the `nav` in our own code, but not allow developers to modify it. Whereas the aria-label for each item should be able to be set by the dev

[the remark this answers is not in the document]
NicolasSimmonds: Sounds good, I'll make those changes then.

modules/cactus-web/src/Breadcrumb/Breadcrumb.mdx
```diff
@@ -0,0 +1,54 @@
...
+name: Breadcrumb
+menu: Components
+---
+
+import PropsTable from 'website-src/components/PropsTable'
+import Breadcrumb from './Breadcrumb'
+
+# Breadcrumb
+
+The `Breadcrumb` comopnent allows you to render a navigation of your previusly visited sites, inlcuding the current one
```
Dhalton: Just noticed a typo in the word "component" here

modules/cactus-web/src/Breadcrumb/Breadcrumb.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+### How it looks
+
+<Breadcrumb ariaLabel="page-navigaton">
+  <Breadcrumb.Item label={text('Label 1', 'Account')} linkTo="/" />
+  <Breadcrumb.Item label={text('Label 2', 'Make a Payment')} linkTo="/" active />
+</Breadcrumb>
+## Best practices
+
+The Breadcrumb component consists of two parts: The `Breadcrum` which works as a wrapper and should incloude a descriptive aria label.
```
Dhalton: Looks like we're missing the last letter of "Breadcrumb" here too

modules/cactus-web/src/Breadcrumb/Breadcrumb.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+### How it looks
+
+<Breadcrumb ariaLabel="page-navigaton">
+  <Breadcrumb.Item label={text('Label 1', 'Account')} linkTo="/" />
+  <Breadcrumb.Item label={text('Label 2', 'Make a Payment')} linkTo="/" active />
+</Breadcrumb>
+## Best practices
+
+The Breadcrumb component consists of two parts: The `Breadcrum` which works as a wrapper for the navigation links.
```
Dhalton, line 21: It looks like we're missing the last letter of "Breadcrumb" here

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const Breadcrumb = styled(BreadCrumbBase)`
+  > ul {
+    display: flex;
+    flex-direction: row;
+    list-style: none;
+  }
+` as any
+
+Breadcrumb.defaultProps = {
+  ariaLabel: 'Breadcrumb item',
```
Dhalton, line 70: Are these default props still necessary since we're not passing the aria label through the props on the `Breadcrumb` component anymore?

PR #221, opened 2020-04-21
test(): Jack up jest timeout for integration tests

Dhalton:
I'm not 100% sure that this will fix the problem, but I did manage to get the integration tests to pass four times in a row with this timeout. We'll have to see how things go with this over the next couple of weeks or so.


NicolasSimmonds:
> I'm not 100% sure that this will fix the problem, but I did manage to get the integration tests to pass four times in a row with this timeout. We'll have to see how things go with this over the next couple of weeks or so.

PR #224, opened 2020-05-06
chore(): Revert menu button styles import

Dhalton:
A few days ago I imported the menu button styles because I thought it would fix the issue in we're seeing on the SSO login page, but it didn't. Since that's the case, these changes aren't necessary and I'd like to remove them b/c I don't want to have to use the css loader.


luisniebla:
I can see my opinion isn't valued around here.

PR #227, opened 2020-05-18
feat(cactus-web): Component theme changes phase one

NicolasSimmonds:
[CACTUS-244]

[CACTUS-244]: https://repayonline.atlassian.net/browse/CACTUS-244

### DIFF
--- a/modules/cactus-web/src/Button/Button.test.tsx
+++ b/modules/cactus-web/src/Button/Button.test.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 
 import { ActionsDelete } from '@repay/cactus-icons'
 import { cleanup, fireEvent, render } from '@testing-library/react'
+import { generateTheme } from '@repay/cactus-theme'
 import { StyleProvider } from '../StyleProvider/StyleProvider'
 import Button from './Button'
 
@@ -161,5 +162,28 @@ describe('component: Button', () => {
 
     fireEvent.click(getByTestId('not-clicked'))
     expect(onClick).not.toHaveBeenCalled()
+  })
+  test
+})
+
+describe('With theme changes ', () => {
+  test('should have 2px border', () => {
+    const theme = generateTheme({ primaryHue: 200, border: 'thick' })
+    const { asFragment } = render(
+      <StyleProvider theme={theme}>
+        <Button>Click me!</Button>
+      </StyleProvider>
+    )
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  test('Should have intermediate border radius', () => {
+    const theme = generateTheme({ primaryHue: 200, shape: 'intermediate' })
+    const { asFragment } = render(
+      <StyleProvider theme={theme}>
+        <Button>Click me!</Button>
+      </StyleProvider>
+    )
+    expect(asFragment()).toMatchSnapshot()
   })
 })
--- a/modules/cactus-web/src/Button/Button.test.tsx
+++ b/modules/cactus-web/src/Button/Button.test.tsx
@@ -163,12 +163,11 @@ describe('component: Button', () => {
     fireEvent.click(getByTestId('not-clicked'))
     expect(onClick).not.toHaveBeenCalled()
   })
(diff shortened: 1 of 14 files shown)


Dhalton:
I'm going to hold off on finishing my review on this until you make the changes we talked about for the MenuButton. Just to recap those for you, we wanted to handle that component in the following ways:

- Make the outline around the button for the focus state match the shape of the button
- If box shadows are turned off, we add a border to the dropdown and use the `lightContrast` color from the theme, and set the border width to the border size from the theme.

I opened my PR for the form elements, so if you want to check out what I did for the DateInput feel free to have a look at #228

modules/cactus-web/src/MenuButton/MenuButton.tsx
```diff
@@ -88,9 +144,10 @@ type MenuButtonType = StyledComponent<typeof MenuButtonBase, any, {}, never> & {
 const MenuButton = styled(MenuButtonBase)`
   position: relative;
   box-sizing: border-box;
-  border-radius: 20px;
+  ${p => getShape(p.theme.shape)};
+  ${p => getBorder(p.theme.border)};
+  box-shadow: ${p => getBoxShadow(p.theme)};
```
Dhalton, line 149: There was no box shadow on the button itself before; only on the dropdown, so let's remove this

modules/cactus-web/src/TextButton/TextButton.tsx
```diff
@@ -89,7 +100,8 @@ export const TextButton = styled.button<TextButtonProps>`
       width: 100%;
       top: 0px;
       left: 0px;
-      border: 2px solid ${p => p.theme.colors.callToAction};
+      ${p => getBorder(p.theme.border)};
+      border-color: ${p => p.theme.colors.callToAction};
       border-radius: 20px;
```
Dhalton, line 105: I wonder if we want to apply the shape from the theme here too. I think we should. Maybe check with Mariana/Hector before doing that

mikesoltow: I think we should too, but it doesn't hurt to check.

modules/cactus-web/src/Button/Button.tsx
```diff
@@ -158,9 +182,9 @@ export const Button = styled(ButtonBase)`
       width: calc(100% + 10px);
       top: -5px;
       left: -5px;
-      border: 2px solid ${p => p.theme.colors.callToAction};
-      border-radius: 20px;
+      border-color: ${p => p.theme.colors.callToAction};
       box-sizing: border-box;
+      ${p => getShape(p.theme.shape)}
```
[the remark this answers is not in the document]
Dhalton: Nice catch, I didn't notice this

modules/cactus-web/src/Button/Button.test.tsx
```diff
@@ -162,4 +163,27 @@ describe('component: Button', () => {
     fireEvent.click(getByTestId('not-clicked'))
     expect(onClick).not.toHaveBeenCalled()
   })
+  test
+})
+
+describe('With theme changes ', () => {
+  test('should have 2px border', () => {
+    const theme = generateTheme({ primaryHue: 200, border: 'thick' })
```
Dhalton, line 171: The border defaults to `thick`, so I'd argue that this is already inadvertently being tested in other snapshots. Let's cover the bases a little more by testing `thin` borders instead

modules/cactus-web/src/MenuButton/MenuButton.test.tsx
```diff
@@ -74,3 +75,47 @@ describe('component: MenuButton', () => {
     })
   })
 })
+
+describe('With theme changes ', () => {
+  test('Should have rounded borders', () => {
```
Dhalton, line 80: Same thing goes for the three new tests in this file too...the shape defaults to `round`, border defaults to `thick`, and `boxShadows` is turned on by default, so these aren't really testing anything new. It's more important to test that the new styles are being applied

PR #228, opened 2020-05-18
Form Element Theming

Dhalton:
This applies the new theme customization features to form element components. Probably a good idea to check this branch out and play around with the new features in Storybook. You can change the border, shape, and box shadows under the "Cactus Theme" knob section.

Let me know if you have any questions or anything

### DIFF
--- a/modules/cactus-web/src/DateInput/DateInput.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.tsx
@@ -1,5 +1,6 @@
 import React, { Component, Fragment, MouseEventHandler, useMemo } from 'react'
 
+import { BorderSize, CactusTheme, Shape } from '@repay/cactus-theme'
 import { compose, margin, MarginProps, width, WidthProps } from 'styled-system'
 import {
   DateType,
@@ -28,7 +29,7 @@ import Portal from '@reach/portal'
 import positionPortal from '../helpers/positionPortal'
 import PropTypes from 'prop-types'
 import Rect from '@reach/rect'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 
 export type IsValidDateFunc = (date: Date) => boolean
 export type ParseDateFunc = (dateStr: string) => Date
@@ -206,6 +207,30 @@ const marginAndWidth = compose(
   width
 )
 
+const borderMap: { [K in BorderSize]: ReturnType<typeof css> } = {
+  thin: css`
+    border: 1px solid;
+  `,
+  thick: css`
+    border: 2px solid;
+  `,
+}
+
+const inputShapeMap: { [K in Shape]: ReturnType<typeof css> } = {
+  square: css`
+    border-radius: 1px;
+  `,
+  intermediate: css`
+    border-radius: 8px;
+  `,
+  round: css`
+    border-radius: 20px;
+  `,
+}
+
+const getBorder = (borderSize: BorderSize) => borderMap[borderSize]
+const getInputShape = (shape: Shape) => inputShapeMap[shape]
+
 const InputWrapper = styled.div`
   position: relative;
   color: ${p => p.theme.colors.darkContrast};
@@ -213,8 +238,9 @@ const InputWrapper = styled.div`
   display: inline-flex;
   flex-wrap: nowrap;
   align-items: center;
-  border: 1px solid ${p => p.theme.colors.darkestContrast};
-  border-radius: 20px;
+  ${p => getBorder(p.theme.border)}
+  border-color: ${p => p.theme.colors.darkestContrast};
+  ${p => getInputShape(p.theme.shape)}
   background-color: ${p => p.theme.colors.white};
   height: 36px;
   outline: none;
@@ -283,13 +309,37 @@ const InputWrapper = styled.div`
   }
 `
 
+const popupShapeMap: { [K in Shape]: ReturnType<typeof css> } = {
+  square: css`
+    border-radius: 0 8px 4px 4px;
+  `,
+  intermediate: css`
+    border-radius: 0 20px 10px 10px;
+  `,
+  round: css`
+    border-radius: 0 32px 16px 16px;
+  `,
+}
+
+const getPopupShape = (shape: Shape) => popupShapeMap[shape]
+const getPopupBoxShadowStyles = (theme: CactusTheme) => {
+  return theme.boxShadows
+    ? css`
+        box-shadow: 0 3px 9px 0 ${theme.colors.callToAction};
+      `
+    : css`
+        ${borderMap[theme.border]}
+        border-color: ${theme.colors.lightContrast};
+      `
+}
+
 const CalendarPopup = styled.div`
   box-sizing: border-box;
   position: absolute;
   z-index: 1000;
   background-color: ${p => p.theme.colors.white};
-  border-radius: 0 32px 16px 16px;
-  box-shadow: 0 3px 9px 0 ${p => p.theme.colors.callToAction};
+  ${p => getPopupShape(p.theme.shape)}
+  ${p => getPopupBoxShadowStyles(p.theme)}
   overflow: hidden;
 `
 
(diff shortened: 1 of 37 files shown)


modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -283,13 +309,37 @@ const InputWrapper = styled.div`
   }
 `
 
+const popupShapeMap: { [K in Shape]: ReturnType<typeof css> } = {
+  square: css`
+    border-radius: 0 8px 4px 4px;
```
mikesoltow, line 314: Why is this still a bit rounded?  Is this what Mariana wanted?  (I'm not super fussed, just curious!)

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -206,15 +207,40 @@ const marginAndWidth = compose(
...
+const getInputShape = (shape: Shape) => inputShapeMap[shape]
+
 const InputWrapper = styled.div`
   position: relative;
   color: ${p => p.theme.colors.darkContrast};
   box-sizing: border-box;
   display: inline-flex;
   flex-wrap: nowrap;
   align-items: center;
-  border: 1px solid ${p => p.theme.colors.darkestContrast};
```
mikesoltow, line 216: It looks like the border used to be 1px?  Are we okay increasing the default setting to 2px?

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: The pop up is taller/wider than a lot of the other bits and pieces we're doing this too, so it looks mostly square I think...but now that I'm looking at it a little closer, I think we might be better off with something like `0 4px 2px 2px` or even `0 2px 1px 1px`.

I came up with the styles myself, but I did pass them on to Mariana and she seemed to be fine with them. I think it might be worth re-visiting though. I'll bring it back up to her tomorrow

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: I'm pretty sure it was left that way by mistake. I did the same thing to the select, I believe. When we originally built these components, everything had 1ps border, but we chose to bump them to 2px and I think these just got missed. Even if it wasn't an accident, I think in the interest of consistency, they should be set to 2px like most everything else. I'll confirm this w/ Mariana though

Dhalton:
@mikesoltow In regards to your first question, I didn't notice the styles on the month/year select in the date picker before. I'm not sure what we want to do with that honestly. I'll bring that up to Mariana when I'm asking her about the other stuff too

modules/cactus-web/src/DateInput/DateInput.tsx
mikesoltow: Sounds good, thanks!

modules/cactus-web/src/DateInput/DateInput.tsx
mikesoltow: Makes perfect sense to me...I'm fine with what you did.  Just wanted to check 😄

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: After talking with Mariana, this has been updated to `0 2px 1px 1px`

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: So I discussed this with Mariana today. She said she thought everything already was set to have a 1px border and that that's what the theme should default to. So I just updated the theme to default to 1px and updated the tests/snapshots for all of the other components. I thought it was the other way around, but apparently we want 1px border defaults

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -381,11 +431,10 @@ const ListWrapper = styled.div`
   position: absolute;
   z-index: 1000;
   box-sizing: border-box;
-  border-radius: 8px;
+  ${p => getListShape(p.theme.shape)}
   max-height: 400px;
   max-width: 100vw;
-  box-shadow: 0 3px 9px 0 rgba(7, 61, 232, 0.28), 0 3px 8px 0 rgba(17, 51, 159, 0.07),
-    0 3px 9px 0 rgba(7, 61, 232, 0.28), 0 3px 8px 0 rgba(17, 51, 159, 0.07);
+  ${p => p.theme.boxShadows && `box-shadow: 0 3px 9px 0 ${p.theme.colors.callToAction};`}
```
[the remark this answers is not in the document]
Dhalton: Yeah no idea why it was hard-coded to a blue color -- it would never change with the theme that way

modules/cactus-web/src/RadioButton/RadioButton.tsx
```diff
@@ -81,7 +81,7 @@ export const RadioButton = styled(RadioButtonBase)`
   }
 
   input:focus ~ span {
-    box-shadow: 0 0 8px ${p => p.theme.colors.callToAction};
+    ${p => p.theme.boxShadows && `box-shadow: 0 0 8px ${p.theme.colors.callToAction};`}
```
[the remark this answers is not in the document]
Dhalton: I hadn't thought about it, let me ask Mariana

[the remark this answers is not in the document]
Dhalton: She said she doesn't think it's necessary

Dhalton:
@mikesoltow 

- I can change the status message to match the theme shape
- I'll update the `ComboInput`
- Mariana said she did end up updating the value tags shape so I'll do that too

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -98,8 +99,9 @@ const ValueTag = styled(ValueTagBase)`
   box-sizing: border-box;
   ${p => p.theme.textStyles.small};
   padding: 0 8px 0 8px;
-  border: 1px solid ${p => p.theme.colors.lightGray};
-  border-radius: 7px;
+  border: 1px solid ${p => p.theme.colors.lightContrast};
+  border-radius: ${p =>
+    p.theme.shape === 'square' ? '1px' : p.theme.shape === 'intermediate' ? '4px' : '8px'};
```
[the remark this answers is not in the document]
Dhalton: Yeah absolutely, I don't like this either. I was having some issues with typescript with the map when I did it before. It was breaking the build. I can keep looking into it.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -543,7 +607,7 @@ const MonthYearListWrapper = styled.div`
     background-color: white;
 
     ul {
-      border-bottom-right-radius: 16px;
+      ${p => getYearShape(p.theme.shape)}
```
mikesoltow, line 610: Do we need a corresponding change to the `border-bottom-left-radius` on line 600?

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -234,6 +249,30 @@ const Placeholder = styled.span`
   font-size: ${p => p.theme.fontSizes.p}px;
 `
 
+const borderMap: { [K in BorderSize]: ReturnType<typeof css> } = {
+  thin: css`
+    border-width: 1px;
+  `,
+  thick: css`
+    border-width: 2px;
```
mikesoltow, line 257: I think this needs to be `border: 1px solid`.  As it currently stands, I'm getting a weird border where half of it is one color and half of it is another color after I click it (use the basic ComboBox story under Select).

Screen Shot 2020-05-22 at 3 10 49 PM

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: No I didn't need to on the other side. I did this because I noticed that the shape of the container to select the year wasn't matching up with the shape of the outline of the calendar dropdown. I didn't see the issue on the bottom left corner because it has the same background color as the border

modules/cactus-web/src/Select/Select.tsx
Dhalton: Whoa very weird. Nice catch I'll fix this!

PR #230, opened 2020-05-21
Add commitizen

mikesoltow:
I was about to make my first real commit, and I noticed that we use a standard commit message template.  In the past, for projects where this was the case, I used a tool called commitizen to make it easier to conform to this template.  I thought I'd give it a whirl here and create a separate PR rather than wrap it up in my other stuff.

What do you guys think?

### How to Test

Pull down the branch and run `yarn` in the root.  To make a commit, run `yarn commit` instead of `git commit`.  It will walk you through the template.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "docs:publish": "gh-pages -d ./website/public",
     "cleanup": "yarn ws run cleanup",
     "run-full": "yarn cleanup && yarn build && yarn test:ci",
-    "release": "node scripts/release.js"
+    "release": "node scripts/release.js",
+    "commit": "git-cz"
   },
   "license": "MIT",
   "private": true,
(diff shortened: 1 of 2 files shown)


Dhalton:
I like the idea, but do you know if there's a way to make this work with a monorepo? We typically provide the package name in parentheses so it's obvious which package the commit is modifying. We also parse that type of info out of the commit messages to auto-generate the changelog

mikesoltow:
> I like the idea, but do you know if there's a way to make this work with a monorepo? We typically provide the package name in parentheses so it's obvious which package the commit is modifying. We also parse that type of info out of the commit messages to auto-generate the changelog

Yup!  The second prompt is a scope, which we could use to put the package name.  If you provide it, it should put it in parentheses like you normally do.  If you skip it (like I did for these, since they applied globally), then it just omits the parens.

PR #231, opened 2020-05-22
Icon, Card, & Accordion Theming

Dhalton:
I looked at the icon buttons but I really don't think we should change anything about them.

To test, just check out this branch and have a look at the card & accordion components and play around with the theming options. Make sure to look at both the `simple` and `outline` accordion variants

### DIFF
--- a/modules/cactus-web/src/Accordion/Accordion.test.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.test.tsx
@@ -546,6 +546,22 @@ describe('component: Accordion', () => {
       expect(asFragment()).toMatchSnapshot()
     })
 
+    test('outline accordion should not have box shadows when open', () => {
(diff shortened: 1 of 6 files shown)


modules/cactus-web/src/Accordion/Accordion.test.tsx
```diff
@@ -527,4 +528,70 @@ describe('component: Accordion', () => {
       expect(container).not.toHaveTextContent('Should not show')
     })
   })
+
```
mikesoltow, line 531: Might want to add in a no box-shadow test here

Dhalton:
> For the Accordion, should the bottom border for the header (see line 211) also change thickness?

Ah yeah I suppose it should. That's the line that separates the header and the body when the accordion is open; I didn't notice it before. I'll update that, nice catch!

PR #232, opened 2020-06-01
CACTUS-230 :: IE11 Integration tests

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-230

This PR replaces the Puppeteer tests in the example apps with TestCafe tests.  These tests can be run on local browsers or on remote browsers via BrowserStack.

### Testing

1. Run `yarn install` to pull in new packages/dump the old ones
2. Create two `local.env` files: one in `examples/mock-ebpp`, the other in `examples/standard`.  Slack me for the contents (they contain browserstack creds)
3. Run `yarn w mock-ebpp test:local -b "Chrome -incognito"`.  You can switch out `mock-ebpp` for `standard`, and you can also use `-b firefox` or `-b safari` to test on other browsers, if you want.


examples/mock-ebpp/tests/testing-library-client-scripts/core.min.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * core-js 3.6.4
```
Dhalton, line 2: I don't have a problem with doing this necessarily, I'm just a little curious why we need to?

mikesoltow:
@Dhalton Thanks for checking out those other browsers!  It's definitely not a quick fix, as I think there might be something wrong with the TestCafe adapters for those browsers (essentially, click isn't working sometimes, which means the selectors don't work because the page isn't in the selected state).

I'm going to merge this since it fixes the primary issue, but I'll add a ticket to the backlog to look into this.  Not sure I'll prioritize it in the short term, however, given the other stuff on our plate, but I'm definitely open to feedback about that.

PR #234, opened 2020-06-02
CACTUS-259 Fix Modal Render Issue

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-259

The main issue, it turns out, was simply that it's a bad idea to specify the height of the modal. It works much better if the modal auto-sizes, and then the user can scroll the entire page if it's too tall: most of the changes were to make the modal look better when there's no height specified. Still, I didn't want to remove the option entirely, so I also added overflow scrollbars. That's really not recommended, though, since the scrollbars turn the nice rounded corners into right angles, and I don't know of any way to prevent that.

There are a quite a few changes here:
* Updated Storybook so there is no default height.
* Added `nowrap` to the primary modal content, so it will always be rendered in a single column.
* Hide the `children` div if there are no children.
* Added a specific margin on the row of buttons.
* Changed the `min-height` to 20% to reduce chance of excess whitespace when there's no title/description.
* Added `overflow: auto` so that if all the other changes fail, the contents will still render inside the modal.
* Added a `min-height` on the icon so it won't get squished.

### Testing
Go to the storybook for `Modal` and play around with it. Try different screen sizes, and resizing the browser window. Try adding an explicit height (as a percentage), and try it all again. You may get an ugly scrollbar, but you should never see the the content outside the modal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/node_modules
 *.log
+*.swp
 **/dist
 **/coverage
 .cache
--- a/modules/cactus-web/src/Modal/Modal.mdx
+++ b/modules/cactus-web/src/Modal/Modal.mdx
@@ -16,7 +16,7 @@ import { livePreviewStyle } from '../helpers/constants'
 
 # Modal
 
-The `Modal` component is usefull when an action from the user is required, and no other interaction
+The `Modal` component is useful when an action from the user is required, and no other interaction
 on the web page is permited until the modal is closed.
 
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 **/node_modules
 *.log
+*.swp
```
Dhalton, line 3: What's this for?

wilysword: Vim, of course.

PR #235, opened 2020-06-02
Table component 99

NicolasSimmonds:


modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,203 @@
...
+  }
+  ${p => p.theme.mediaQueries.medium} {
+    width: calc(160px * 0.7125);
+  }
+  ${p => p.theme.mediaQueries.large} {
+    width: calc(160px * 0.875);
+  }
+
+  ${p => p.theme.mediaQueries.extraLarge} {
+    width: 160px;
```
wilysword, line 134: These should probably be `min-width` instead of `width`.

wilysword:
I kind of wonder if it wouldn't be better to do it like Griddle does, where you just define the header configuration and then pass in the data, but that might be too high-level for this component.

mikesoltow:
> I kind of wonder if it wouldn't be better to do it like Griddle does, where you just define the header configuration and then pass in the data, but that might be too high-level for this component.

If I'm understanding you correctly, that's what [this ticket](https://repayonline.atlassian.net/browse/CACTUS-132) should accomplish.  This is just the low-level piece that will be a part of that higher-level component.

wilysword:
Fair enough. I was mostly speculating because of the table-to-card-carousel conversion they want: that will require attaching information from the header row to every other row, which is going to be crazy (the `Table` component will have to somehow get its grandchildren from the `Table.Header` and then smash them together with its grandchildren from the `Table.Row`s). I linked a CSS-only way of doing it in my other answer, but that actually won't work for us because we can't guarantee the header cells only have text in them; so we'll probably need something like [this](https://medium.com/better-programming/how-to-use-media-queries-programmatically-in-react-4d6562c3bc97) that allows us to programatically swap out the components we're using on small screens. That data-grid thing raises new questions in my mind, though: where are the sorting & pagination controls going to be on the card-carousel view? (I probably shouldn't be cluttering up this PR with tangents, though.)

modules/cactus-web/src/Table/Table.tsx
NicolasSimmonds: Did this but  it ended up filling all the space available so the mediaquery wasn't doing anything.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,214 @@
...
+const getDoubleBorder = (shape: Shape) => doubleBorder[shape]
+const getBorder = (size: BorderSize) => borderMap[size]
+
+const Tablebase: FunctionComponent<TableProps> = props => {
+  const { children, className } = props
+
+  return <table className={className}>{children}</table>
+}
+
+const TableCellBase: FunctionComponent<TableCellProps> = props => {
```
Dhalton, line 72: I'd like to propose that we split this component into two components: `TableHeader` and `TableData`, and set those both as sub-components of table. Then, if we changed the name of `Table.Header` to `Table.Head`, our API could look something like this:
```
<Table>
  <Table.Head>
    <Table.Header>Header</Table.Header>
  </Table.Head>
  <Table.Body>
    <Table.Row>
      <Table.Data>Data</Table.Data>
    </Table.Row>
  </Table.Body>
</Table>
```

That API would be closer to the native table API and wouldn't require passing `cellType`.

Does anyone else have any opinions on this?

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,214 @@
...
+  :focus {
+    ::after {
+      content: '';
+      display: block;
+      position: absolute;
+      height: calc(100% + 10px);
+      width: calc(100% + 10px);
+      top: -5px;
+      left: -5px;
+      border: 1px solid;
```
mikesoltow, line 187: Should this thickness change when the theme's thickness changes?

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  ${p => p.theme.mediaQueries.extraLarge} {
+    width: 160px;
+  }
+`
+
+export const TableRow = styled(TableRowBase)`
+  position: relative;
+  margin: 4px 0;
```
mikesoltow, line 150: I think you'll need some horizontal margin here.  Without it, the focus border gets cut off.  See the top row in the screenshot.
Screen Shot 2020-06-05 at 1 41 34 PM

modules/cactus-web/src/Table/Table.tsx
mikesoltow: I'm fine with this.

wilysword: I don't necessarily mind the switch to `Table.Head`, but I think the other changes are confusing. When I hear "header" I think the whole row; similarly when I hear "data" I think _all_ the data. `Cell` is nice and unambiguous. We could do `HeaderCell` and `DataCell`, but those names are getting kind of long. I think ideally we'd support the `as` parameter and default it to `th` in the header and `td` in the body, but that would be pretty tricky.

NicolasSimmonds: In the latest commit, I did something like the suggestion from Glen. The table cell accepts a cellType prop, (which can be renamed to 'as') and it can be either th or td. So in the header it'd render the th and in the body the td. But I'm fine with whatever you prefer to do.

mikesoltow: Good call.  I think we could support what you're suggesting without getting too crazy.  One option would be to have the `Table.Head` and `Table.Body` components set a default element type in a context that `Table.Cell` makes use of.

modules/cactus-web/src/index.ts
```diff
@@ -38,3 +38,4 @@ export { DateInputField } from './DateInputField/DateInputField'
 export { MenuButton } from './MenuButton/MenuButton'
 export { Modal } from './Modal/Modal'
 export { Breadcrumb } from './Breadcrumb/Breadcrumb'
+export { Table } from './Table/Table'
```
Dhalton, line 41: We should probably refrain from exporting this component until the other ticket is done & it's fully built out.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+type cellAlignment = 'center' | 'right' | 'left'
+
+type cellType = 'th' | 'td'
+
+interface TableContext {
+  cellType: cellType
+}
+
+interface TableProps {
```
Dhalton, line 14: For each of these props interfaces, should we extend from the HTML type definitions? Would look something like this, for the table at least:
```
interface test
  extends React.DetailedHTMLProps<React.TableHTMLAttributes<HTMLTableElement>, HTMLTableElement> {
  // any extra props here...
}
```
That way, we could also destructure any remaining props in the component and spread them onto the element, so all of the attributes that are normally available on the semantic table elements are still available in ours. We do this type of thing with pretty much every other cactus component; I don't see why we wouldn't want to do that with these.

I'm open to discuss this if anyone has any objections. cc @mikesoltow @wilysword

wilysword: I suspect there are a couple where we could just directly style the HTML element, no intermediate component needed.

mikesoltow: Assuming we look into eliminating any extra intermediates, as Glen said, I don't have any objections.

modules/cactus-web/src/index.ts
mikesoltow: Agreed

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,234 @@
...
+  thick: css`
+    border: 2px solid;
+  `,
+}
+
+const getShape = (shape: Shape) => shapeMap[shape]
+const getDoubleBorder = (shape: Shape) => doubleBorder[shape]
+const getBorder = (size: BorderSize) => borderMap[size]
+
+const Tablebase: FunctionComponent<TableProps> = props => {
```
mikesoltow, line 73: Sorry to nitpick, but this should probably be `TableBase`

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,234 @@
...
+}
+
+interface TableRowProps {
+  className?: string
+}
+interface TableBodyProps {
+  className?: string
+}
+
+const TableContext = createContext<TableContext>({
```
mikesoltow, line 33: Should we avoid naming the instance of the context the same thing as the interface?  Or is that not a big deal?

modules/cactus-web/src/Table/Table.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { StyleProvider } from '../StyleProvider/StyleProvider'
+import Table from './Table'
+
+afterEach(cleanup)
+
+describe('component: Table', () => {
+  test('snapshot', () => {
+    const { container } = render(
+      <StyleProvider>
+        <Table />
```
mikesoltow, line 13: We should probably add a header and some rows to this test so that the snapshot has a fully rendered table.

modules/cactus-web/src/Table/Table.mdx
```diff
@@ -0,0 +1,39 @@
...
+name: Table
+menu: Components
+---
+
+import PropsTable from 'website-src/components/PropsTable'
+import Table from './Table'
+
+# Table
+
+TODO - describe component here
```
mikesoltow, line 11: Are we leaving this as a todo until the next ticket?

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,234 @@
...
+  & small {
+    color: ${p => p.theme.colors.darkestContrast};
+  }
+  :hover {
+    fill: ${p => p.theme.colors.callToAction};
+    cursor: pointer;
+    background-color: #d5e8f2;
+    ${p => getBorder(p.theme.border)}
+    border-color: ${p => p.theme.colors.callToAction};
+    box-shadow: -1px 6px 4px 0px rgba(3, 118, 176, 0.35);
```
mikesoltow, line 194: This bit of box shadow is still not responding to the theme choices

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+export const TableRow = styled(TableRowBase)`
+  position: relative;
+  margin: 4px 4px;
+  outline: 0;
+  display: flex;
+  background-color: ${p => p.theme.colors.white};
+  ${p => getShape(p.theme.shape)};
+  ${p => getBorder(p.theme.border)};
+  border-color: #dee8ed;
```
mikesoltow, line 173: Should this border color (and the one on line 178) be a function of the theme instead of hard-coded?

modules/cactus-web/src/Table/Table.tsx
Dhalton: Yeah, nice catch, I second this

modules/cactus-web/src/Table/Table.mdx
NicolasSimmonds: Since this is not the final component, I don't know if its worth it to document it since there may be some changes along the way. But I'm ok with whatever you guys think its better

mikesoltow:
@Dhalton @NicolasSimmonds If we merge from master and push, they should pass.  This branch doesn't have the latest changes from the IE integration tests.

@wilysword I'm going to wait for your feedback before merging

PR #241, opened 2020-06-08
CACTUS-255 :: StatusMessage Style Update

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-255

I'm sure I missed something, but hopefully I got everything in there.  When testing, I'd be sure to go into all of the `BlahBlahField` components (i.e. TextInputField, SelectField, ...) with the exception of `CheckboxField` and `RadioButtonField`.

Also, I ended up making some changes to the `FileInput` component, as the style guide did not have the `StatusMessage` taking up 100% of the width.  This required adjusting the placement of the refresh icon.  I think it looks fine if the message is on one line, but I'm not sure about two lines.  However, I'm unsure how to style it any differently.

### DIFF
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
@@ -112,9 +112,11 @@ function AccessibleFieldBase(props: AccessibleFieldProps) {
           status,
         })}
       {status !== undefined && (
-        <StatusMessage status={status} id={statusId}>
-          {statusMessage}
-        </StatusMessage>
+        <div>
+          <StatusMessage status={status} id={statusId}>
+            {statusMessage}
+          </StatusMessage>
+        </div>
       )}
     </FieldWrapper>
   )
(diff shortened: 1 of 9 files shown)


Dhalton:
> I don't know if this is what we want. But in the AccessibleField, the status message shows on the right of the input and it looks kind of odd to me.
Hmmm, weird, I didn't see this when I looked at that component 🤔

mikesoltow:
> I don't know if this is what we want. But in the AccessibleField, the status message shows on the right of the input and it looks kind of odd to me.
> Screen Shot 2020-06-09 at 3 27 10 PM

I noticed this and pushed up a new commit this morning.  Try pulling and restarting...let me know if you still see it after that.

mikesoltow:
Are you guys okay with me merging this without the file reload icon change?  I just looked at the currently deployed version, and you can't even see it (because the current version has a white background), so this is technically an improvement.

I can create another ticket and talk to Mariana about redesigning this, though.

PR #243, opened 2020-06-09
SplitButton Component

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-101

At long last, I give you: the SplitButton 😎 

I mentioned this in a few meetings, but I'll reiterate here that I had to upgrade `@reach/menu-button` in order to make use of the `position` function. This is because we need to position the dropdown based on the main button instead of the trigger that opens the dropdown. However, the `@reach` deps all work with & depend on each other, so upgrading `menu-button` meant I had to upgrade several other of their deps. As a result, we should be extra careful in testing components that use those dependencies. These include:

- `MenuButton`
- `Tooltip`
- `Select`
- `Modal`
- `Accordion`
- `DateInput`

Other than that, feel free to check this branch out locally and test out the new component & let me know what you think!


modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -24,15 +24,22 @@ import styled, {
...
   onSelectMainAction: (event: React.MouseEvent<HTMLButtonElement>) => void
   mainActionIcon?: React.FunctionComponent<IconProps>
   disabled?: boolean
+  // Aria label for the dropdown trigger. Defaults to "Action List"
+  'aria-label'?: string
 }
 
-interface SplitButtonActionProps extends MenuItemImplProps {
+interface SplitButtonActionProps extends Omit<MenuItemImplProps, 'onSelect'> {
+  // !important
```
wilysword, line 41: Is this the fix for the props table in the docs?

Dhalton: Yeah, part of it anyway. I wanted to make sure `onSelect` was visible in the table, but the `PropsTable` component throws some props out if it thinks they're not sueful. If you ned to force some props to show up, you can add `!important` in its description

Dhalton:
@mikesoltow I looked into what you mentioned about being able touse the escape key to exit the dropdown, but it was working for me. I know if you have VoiceOver on, you have to double-tap the escape key (if you have the touchpad, anyway). Could it be that you weren't double-tapping when you were testing that?

PR #246, opened 2020-06-17
ScreenSizeProvider

wilysword:
Let me know what you think about naming, utility, etc. Also, should we add `tiny` to the list of breakpoints on the theme so the keys of the two lists match?


Dhalton:
This is awesome! I can see how this will be a useful tool. I don't really have any changes I'd like to suggest; naming and utility seem good, but yeah I think we should add `tiny` to the breakpoints in the theme

mikesoltow:
I agree, this is awesome!  I'm not sure how `tiny` would work on the theme...as shouldn't it always be 0?  Or are you saying we just hardcode it to 0 always?

In any case, I don't think we should get the breakpoints from the constants file, as users can technically override that (unless you were just doing that for the quick illustration).  You can access it with `useContext` (https://styled-components.com/docs/advanced#via-usecontext-react-hook)

PR #247, opened 2020-06-19
CACTUS-273 :: SplitButton & MenuButton Integration Tests

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-273

Working on this made a bug that we hadn't noticed before pretty obvious: the background on the dropdown in both of these components was transparent! I guess we just hadn't rendered them on anything but a white background. So that should be fixed now.

## MenuButton

I swapped the language select in `examples/standard` to the `MenuButton` & updated the integration test to work with that component

## SplitButton

For this one, I added a column to the accounts table in `examples/mock-ebpp` and filled it up with SplitButtons with and some basic actions: view the account page or delete the account. Obviously the account page I added is very basic but hey, it's just an example app 🤷‍♂️ 

## Testing

Not really much to test here, but you should probably check out this branch and run the integration tests in the `mock-ebpp` and `standard` example apps. You can also run the `mock-ebpp` app and have a look at the new functionality I added, if you're interested in it.

### DIFF
--- a/examples/mock-ebpp/tests/account-integration.test.ts
+++ b/examples/mock-ebpp/tests/account-integration.test.ts
@@ -23,6 +23,6 @@ test('navigate to account page using SplitButton', async (t: TestController) => 
 
 test('delete account with using SplitButton', async (t: TestController) => {
   await t.click(Selector('button').withAttribute('data-reach-menu-button').nth(0))
-  await t.click(queryByTestId('delete-acct-85963'))
+  await t.click(queryByText('Delete Account'))
   await t.expect(queryByText('Account 85963 deleted successfully').exists).ok()
 })
--- a/examples/mock-ebpp/tests/account-integration.test.ts
+++ b/examples/mock-ebpp/tests/account-integration.test.ts
@@ -1,0 +1,28 @@
+import * as path from 'path'
+import { queryAllByText, queryByTestId, queryByText } from '@testing-library/testcafe'
+import { Selector } from 'testcafe'
+import startStaticServer from './helpers/static-server'
+
+fixture('Account Integration Tests')
+  .before(async ctx => {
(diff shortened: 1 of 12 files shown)


examples/mock-ebpp/tests/account-integration.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  .page('http://localhost:33567/accounts')
+
+test('navigate to account page using SplitButton', async (t: TestController) => {
+  await t.click(queryAllByText('View Account').nth(0))
+  await t.expect(queryByText('Account 85963').exists).ok()
+})
+
+test('delete account with using SplitButton', async (t: TestController) => {
+  await t.click(Selector('button').withAttribute('data-reach-menu-button').nth(0))
+  await t.click(queryByTestId('delete-acct-85963'))
```
mikesoltow, line 26: You probably don't need to use a test ID here.  If `queryByText` wasn't working for you, it's probably because there's a typo in the dropdown.  "Delete Acount" needs to be changed to "Delete Account" 😄

Dhalton: Haha yeah I didn't notice that at all, whoops! Nice catch

Dhalton:
@mikesoltow About the console error you were seeing...nice catch, I must not have had my console up when I was setting that up. I've been looking into it trying to figure out what's causing it, no luck so far though. I don't see anything in my code that looks like it'd cause that (I'm making sure to pass a function definition to `onSelect` and not directly calling `onLangChange`) so I guess it's possible that it's related to the out-of-date deps. I'm working on updating those now, so what if we merge this, I can check if the error goes away when everything is up-to-date and if not I can keep digging on it?

PR #249, opened 2020-06-23
CACTUS-275 :: Table Redesign

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-275

While I was working on this, I noticed that Styled Components was not type-checking the theme correctly within our comments.  It turns out that there's a way to fix this (see https://styled-components.com/docs/api#typescript), so I added that in, and type checking started working!  A few errors popped up as a result (mostly as a result of mediaQueries being optional), so I fixed those.  **It may be easier to look at those commits by themselves rather than with the rest of the PR.** I can also split them out into a separate PR if you all would prefer.

**Another thing to note while testing:** The Style Guide has hover/focus borders only on the top/bottom, but I added the borders all the way around.  Having borders only on the top/bottom did not look very good on the last row due to the border radius.  I'll talk to Mariana tomorrow and perhaps modify this, but I'd be interested in your feedback as well.

### Testing

* Remember to test on IE in BrowserStack!
* Check that the borders/backgrounds for the focus and hover states look good


mikesoltow:
> I immediately notice that you're still using `display: flex`; did you try and fail using the default table display, or did you not try? I ask because I was pretty sure the new design could be implemented without overriding the standard table behavior, though I could be wrong about that.

I definitely did try.  Most of the design was definitely doable with table styles, but the borders for hover/focus were what did me in (after spending almost a full day of trying to get them to work).  Long story short (okay, not that short): when using standard table styling, you can't set a border on a `tr` element (unless the table has `border-collapse: collapse`...but in that case the top border is always cut off because the borders are collapsed).  Therefore, it was necessary to try to set a row border by combining a bunch of `td`/`th` borders.  This almost worked, but IE would add random white space in between cells that had a border radius for that.  There's a fix for this, which is to add a 1px border on the left/right side to cover up that white space.  But when the side borders are non-zero, then it adds little divots to the top/bottom borders so that they're no longer continuous lines.

Maybe I could have pushed through to something, but at that point I had spent hours on everything, and using flex "just worked," so I went back to that route.

mikesoltow:
@Dhalton Good call.  I just added a knob for that on a few of the stories.

wilysword:
Oh ye of little faith, behold: https://github.com/repaygithub/cactus/compare/cactus-275-table-redesign...glen-table-style. Basically borders have to be done on the cells, not on the row or table. Kind of annoying, and it leads to rather complex CSS, but very doable. There is still at least one weird thing with this: if you have a header + empty tbody, the header doesn't get the rounded borders on bottom. Not sure if we care about that enough to try to fix it. I also added a much more useful story for seeing how the layout reacts to different numbers of rows & columns, long text, etc.

Now since you've already got two approvals, you can merge this if you want and we can talk about the table display later; I would very much like it if we could eventually use it, though.

I did indeed see a weird issue on IE due to border radius, but I was able to get rid of it by explicitly setting the border on the header cells.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -138,118 +119,144 @@ const TableBodyBase: FunctionComponent<TableBodyProps> = (props) => {
...
-      top: -5px;
-      left: -5px;
-      ${(p) => getBorder(p.theme.border)}
-      ${(p) => getDoubleBorder(p.theme.shape)};
-      border-color: ${(p) => p.theme.colors.callToAction};
-      box-sizing: border-box;
+
+  // first row
+  & > tr:first-of-type,
+  tbody > tr:first-child,
```
wilysword, line 208: Just noticed I missed something here, since tables are allowed to have more than one `tbody`.
```suggestion
  tbody:first-of-type > tr:first-child,
```

wilysword: Okay, that didn't fix it after all...

wilysword: This is kind of an uncommon use case, so I'd be fine if you leave it for now and I'll try to fix it in my PR.

wilysword:
I did a little research on the click focus/IE thing yesterday (https://nemisj.com/focusable/), and I think it's just how IE behaves. I guess we could add an `onClick` handler that explicitly calls `focus()`, but that seems like extreme overkill.

Dhalton:
@wilysword Ah, I see. Thanks for linking that article. I think I'll bookmark it for future reference. With that in mind, I'll go ahead and approve this. Again, this is looking great

wilysword:
Actually, considering the article is ten years old I'm surprised what it said about IE seems to still be true.

mikesoltow:
> Wow, this is looking _really_ good. I really like what you did with the Storybook. Much cleaner & easier to interact with this way.

That's all Glen (including the new story)!  He's becoming a CSS ninja 😄 

And I'm fine leaving out the focus thing from IE now.  I really don't anticipate it becoming an issue, and we can always add a click handler if it does.

mikesoltow:
> Actually, considering the article is ten years old I'm surprised what it said about IE seems to still be true.

Are you _really_ surprised? 😄

modules/cactus-web/src/Table/Table.tsx
mikesoltow: Yeah, I'm happy to leave it for now, but nice catch!

PR #250, opened 2020-06-23
CACTUS-277 :: Upgrade Reach Deps

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-277

## Some Notes
- The `MenuButton` & `SplitButton` will behave _slightly_ differently as a result of this upgrade; they updated the behavior when the trigger is clicked. It used to be that, when you click the trigger, the first option in your list will receive focus automatically, but they changed it so that if you click the trigger, it won't do this any more. It will only focus the first option if you open the dropdown using the keyboard. PR for reference: https://github.com/reach/reach-ui/pull/550
- You may notice that I had to use the `pressKey` function in integration tests that interact with `SplitButton` & `MenuButton`. It seems like there's some kind of incompatibility with `t.click` from TestCafe and the options in `@reach/menu-button` dropdowns. For some reason, I could not get the tests to pass when I used `t.click` on the MenuButton/SplitButton options, even though it works fine for me when I tested myself using:
  - Chrome on my Mac
  - Chrome on Windows
  - IE11
- @mikesoltow That error message you brought up when switching the language in the standard example app is gone, so it must've been due to the out-of-date deps.
- This fixes the error message you'd see in the SplitButton or the MenuButton if you opened the dropdown and then clicked outside of it w/o selecting an option.

## Testing
- Check out this branch and run `yarn install` at the root
- Bring up the Storybooks with `yarn start`
- Play around with the following components and make sure they're still working as expected:
  - Accordion
  - DateInput
  - MenuButton
  - Select
  - Modal
  - SplitButton
  - Tooltip


Dhalton:
@mikesoltow Updated that prop type

PR #254, opened 2020-06-29
CACTUS-282 :: Unbounded Pagination

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-282

Should be a pretty simple component to test.  Just make sure all the knobs work in Storybook and that accessibility look fine...and that you're happy with the API, of course.


modules/cactus-web/src/PrevNext/PrevNext.tsx
```diff
@@ -0,0 +1,120 @@
...
+      disabled={disableNext}
+      onClick={() => !disableNext && onNavigate('next')}
+    >
+      {nextText}
+    </PrevNextLink>
+  </>
+)
+
+export const PrevNext = styled(PrevNextBase)`
+  ${margin}
```
wilysword, line 97: This has no effect; you should probably either remove it, or you'll need to wrap the buttons in a div or something.

modules/cactus-web/src/index.ts
```diff
@@ -40,3 +40,4 @@ export { MenuButton } from './MenuButton/MenuButton'
 export { Modal } from './Modal/Modal'
 export { Breadcrumb } from './Breadcrumb/Breadcrumb'
 export { SplitButton } from './SplitButton/SplitButton'
+export { PrevNext } from './PrevNext/PrevNext'
```
wilysword, line 43: Should we be trying to keep this file alphabetized?

mikesoltow: Yeah, once we fix the TypeScript linting I'm hoping that will happen automatically (like it did in channels).  I'll just clean it up manually for now.

mikesoltow: Actually scratch that...linting doesn't work on the exports

modules/cactus-web/src/PrevNext/PrevNext.tsx
mikesoltow: Good catch, thanks!

PR #255, opened 2020-06-30
Modal redesign

NicolasSimmonds:
[CACTUS-265]

[CACTUS-265]: https://repayonline.atlassian.net/browse/CACTUS-265

### DIFF
--- a/modules/cactus-web/src/Modal/Modal.tsx
+++ b/modules/cactus-web/src/Modal/Modal.tsx
@@ -137,9 +137,9 @@ Modal.propTypes = {
 }
 
 Modal.defaultProps = {
-  closeLabel: 'Close Label',
+  closeLabel: 'Close Modal',
   isOpen: false,
-  modalLabel: 'Modal Label',
+  modalLabel: 'Modal',
   variant: 'default',
 }
 export default Modal
--- a/modules/cactus-web/src/Modal/Modal.tsx
+++ b/modules/cactus-web/src/Modal/Modal.tsx
@@ -93,6 +93,7 @@ export const ModalPopUp = styled(DialogOverlay)<ModalPopupProps>`
   position: fixed;
   right: 0;
   top: 0;
+  align-items: center;
   z-index: 52;
   > [data-reach-dialog-content] {
     ${(p) => getBorder(p.theme.border)};
--- a/modules/cactus-web/src/Modal/Modal.tsx
+++ b/modules/cactus-web/src/Modal/Modal.tsx
@@ -1,5 +1,4 @@
-import { BorderSize, Shape } from '@repay/cactus-theme'
-import { CactusTheme } from '@repay/cactus-theme'
+import { BorderSize, CactusTheme, Shape } from '@repay/cactus-theme'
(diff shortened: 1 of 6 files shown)


modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -1,192 +1,144 @@
...
-      width: 100%;
-    }
-  }
   > [data-reach-dialog-content] {
-    margin: 10vh auto;
+    ${(p) => getBorder(p.theme.border)};
+    ${(p) => getShape(p.theme.shape)};
     background: white;
+    border-color: ${baseColor};
+    box-shadow: ${(p) => p.theme.boxShadows && '0px 9px 24px rgba(3, 118, 176, 0.348704)'};
```
Dhalton, line 102: It looks like we were using a hard-coded color for the box-shadows before, so it's not _technically_ being introduced in this PR, but we really should use a color from the theme for the shadows. I think it should be either `callToAction` or `transparentCTA`.

modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -1,192 +1,144 @@
...
   return (
     <ModalPopUp
       aria-label={modalLabel}
       isOpen={isOpen}
-      onDismiss={closeModal}
+      onDismiss={onClose}
       variant={variant}
       {...rest}
     >
       <DialogContent aria-label={`${modalLabel}-content`}>
```
Dhalton, line 73: The `aria-label` should be both translatable and readable by a screen-reader. I'm not sure how screen-readers handle dashes, but in any case, I think we should get rid of the `-content` hard-coded portion of the label, since it wouldn't be translatable.

Dhalton, line 73: I was just doing some reading on modal accessibility and it looks like we should be setting `aria-modal="true"` on the dialog

NicolasSimmonds: Agree, but this is done automatically by reach. If you inspect the element you'll see that attribute!

Dhalton: Gotcha, I didn't know that. Cool!

modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -1,192 +1,145 @@
...
+  variant: PropTypes.oneOf(['default', 'danger', 'warning', 'success']),
 }
 
 Modal.defaultProps = {
-  variant: 'action',
-  modalTitle: 'Modal Title',
-  description: 'Modal description',
-  buttonText: 'Confirm',
-  modalLabel: 'Modal Label',
   closeLabel: 'Close Label',
```
mikesoltow, line 140: For modalLabel and closeLabel, I wonder if we want more descriptive strings as a default, as these don't seem as helpful for the screen reader (for example, when I close the modal using a screen reader, it currently reads "press Close Label, modal").  I think they'd make more sense as the following: `{ "modalLabel": "Modal", "closeLabel": "Close Modal" }`

PR #257, opened 2020-07-01
CACTUS-258 Table - Card Carousel

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-258

First commit is a bug fix I realized we needed when I was reading the the `styled-system` docs. Second is the `ScreenSizeProvider`, and third the table changes.

Added a couple of helpers, one for theme/box shadows, and the other for "variants" based on `styled-system`; I can explain that more in-depth if you guys want me to.

On the table storybook I added a new story, where instead of using the `Table.*` components, I used the raw HTML elements. Doing it like that loses some of the extra card-carousel functionality (specifically inserting the headers in each row), but I wanted to have an example of what it looks like, and ask you guys if you think we should move more of the styles to the `StyledTable`.

Speaking of `StyledTable`: aside from moving stuff around and adding the new styles for card carousel, I also switched up how the styling is done. I mentioned this in my pagination PR, but `styled-components` have some functionality that we really probably don't want to expose to outside devs, especially the `as` prop. Here I've fully exposed that prop on `Table`, and exposed it on `Table.Cell` with the values being restricted to only "td" or "th".

One note about accessibility: because in card mode the headers are inserted into each cell, the header row is hidden by moving it far to the left of the viewport. Since the inserted headers are marked with `aria-hidden`, screen readers _should_ read it exactly the same way in card mode as in table mode. (I'm not sure because Orca wouldn't read the table contents at all, and I haven't found a replacement yet.)

I haven't written tests for the `ScreenSizeProvider` yet since I'm still trying to figure out how I want to mock the media queries browser API; there's nothing visual so a snapshot is pointless.


modules/cactus-web/src/Table/Table.tsx
```diff
@@ -1,170 +1,293 @@
...
+  cellIndex: number
+  variant: TableVariant
 }
 
 interface TableProps
-  extends React.DetailedHTMLProps<React.TableHTMLAttributes<HTMLTableElement>, HTMLTableElement> {
-  className?: string
+  extends Omit<
+    React.DetailedHTMLProps<React.TableHTMLAttributes<HTMLTableElement>, HTMLTableElement>,
+    'ref'
```
mikesoltow, line 30: Out of curiosity, why are we omitting `ref`?

wilysword: I don't really know. Typescript was complaining about the type being wrong, but since I wasn't even using `ref` at all, I had no idea what it was talking about. Omitting it was the only way I could get it to stop complaining.

Dhalton:
This is looking awesome!

I few notes/questions for you:

- I was able to confirm that VoiceOver is reading the rows in card mode the same way they're being read in table mode, so it seems to be accessible.
- I would probably be fine with extracting more of the styles into `StyledTable` if I just understood the use-case a bit more. What I mean is that I just don't see why developers would want to use `tr` as opposed to `Table.Row` when they're using our table component. Could you go into more detail about why you think we need to cover that scenario?
- I noticed one thing on the table mode while testing that I think we might want to address: When I add more rows, I reach a point (~10+ on my screen size) at which I'm not longer able to see the headers, and as I add even more, those rows end up off-screen as well. I can't tell why this is happening, as there _is_ a scrollbar there. It just doesn't let me scroll as far as I might need to.
- This is kind of nit-picky, and by no means do we have to do this, but I think it'd be nice if, in the card view, when you tab to a new card that's currently off-screen, it's automatically moved on-screen, rather than having to use the arrow keys to make it show up. Just an idea.

wilysword:
> What I mean is that I just don't see why developers would want to use tr as opposed to Table.Row when they're using our table component.

Most likely they wouldn't. The only one that's at all likely is using `thead` instead of `Table.Header`, for instance if they wanted multiple header rows. There's no real reason beyond that, just that most of the styles are in the table, so might as well put more of them there 🤷.

> I'm not longer able to see the headers, and as I add even more, those rows end up off-screen as well.

I'm pretty sure this is something to do with storybook. Storybook wraps the component in a div that centers its child horizontally and vertically, and it behaves strangely on overflow. On Firefox, if I select the wrapping div and turn off all the `display` styles, I can see the whole table. Since that div is rendered by storybook, I'm not sure if there's a way to control the styles on just that story.

> I think it'd be nice if, in the card view, when you tab to a new card that's currently off-screen, it's automatically moved on-screen

That's a good idea, but I have no idea where to even start looking for how to do it.

modules/cactus-web/src/Table/Table.tsx
mikesoltow: I looked into this a bit, and the issue is that you were passing a ref that was expecting a React functional component into a styled component that styled an HTML element directly.  If you want to allow the user to pass in a ref to `Table` that can be used to reference the HTML `table` element directly, then you can use the `React.forwardRef` pattern to pass that down.  Otherwise, you can leave as-is.

Regardless of which way you go, I pushed up a branch that you might want to take a look at.  It has two commits: the first can be used to get rid of the `ts-ignore` statements and the TSWorkaround you use, while the second commit can be used to implement `React.forwardRef` if you wish.  https://github.com/repaygithub/cactus/compare/cactus-258-table-cards...fun-with-typescript?expand=1

wilysword: I wasn't all that concerned with allowing refs, since I can't imagine what anyone would do with a ref to a table, but on the other hand there's no real reason to prevent it (especially since you already did the work). Both those commits look good: if you rewrite the commit messages to the commitizen format, I'll just fast-forward my branch to include them. (Also, on [line 30](https://github.com/repaygithub/cactus/commit/f51289e2d67bd4d9013accdce2b3c2d74ea5a6af#diff-8f624f20ceeaf0b7d031e9191d78dbf3R30) I don't think there's supposed to be a trailing comma.)

mikesoltow: Works for me...done!

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -1,170 +1,293 @@
...
+//@ts-ignore
+Table.propTypes = {
+  fullWidth: PropTypes.bool,
+  cardBreakpoint: PropTypes.oneOf<Size>(['tiny', 'small', 'medium', 'large', 'extraLarge']),
+  variant: PropTypes.oneOf<TableVariant>(['table', 'card']),
+  as: PropTypes.elementType,
+}
+
+Table.defaultProps = {
+  cardBreakpoint: 'tiny',
```
mikesoltow, line 191: Should we match this up with the Style Guide, which says that the styles should switch at `small` by default?

wilysword: It does say that, but to me that seems really wide compared to the width of the cards. I think it's better to have it default to the smallest size, and then the dev can decide to use a higher breakpoint if the table is very large.

wilysword:
> the caption doesn't really show up correctly if you are in card mode

Technically we haven't done any styling at all for the caption in table _or_ card mode. I assumed that would be a separate effort (if we do it at all), since it's not mentioned in the style guide.

modules/cactus-web/src/Table/Table.tsx
mikesoltow: Fair point...I'd be curious how this looks in the wild.  I'm ultimately fine either way, but I'd like to get Hector's opinion too.

mikesoltow: Hector's good with this, so I'm good with this.

PR #265, opened 2020-07-08
Confirm modal

NicolasSimmonds:
Made a width adjustment on the Modal container. The max-width for Mobile went from 90% to 80% and desktop from 60% to 30%. This was discussed with Hector.

Also two more variants were added to the Button: "warning" and "success", this colors have to be double-checked with the styleguide when they are ready.


wilysword:
Haven't looked at the code yet, but the styling seems mostly pretty good; I wonder, though if we need another width for medium screens, maybe 50%?

Also, it seems like your word wrap settings are off somewhere:
modal-wrap

modules/cactus-web/src/index.ts
```diff
@@ -47,3 +47,4 @@ export { Toggle } from './Toggle/Toggle'
 export { ToggleField } from './ToggleField/ToggleField'
 export { Tooltip } from './Tooltip/Tooltip'
 export { default as useId } from './helpers/useId'
+export { ConfirmModal } from './ConfirmModal/ConfirmModal'
```
wilysword, line 50: I believe we're trying to keep the exports at least semi-alphabetical.

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+import Flex from '../Flex/Flex'
+import Modal, { ModalProps, ModalType } from '../Modal/Modal'
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled, { css, ThemeProps } from 'styled-components'
+import Text from '../Text/Text'
+import TextButton from '../TextButton/TextButton'
+
+export type IconSizes = 'medium' | 'large' | undefined
+export type IconNames = keyof typeof icons | undefined
```
wilysword, line 13: Still new to Typescript, but I don't think you need to include the `| undefined` in these types; it's enough that in the interface definitions you have the `?` marking them as optional. Unless there's another place they're used where it _is_ necessary.

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+interface IconProps {
+  iconSize: IconSizes
+  iconName: IconNames
+  className?: string
+  variant: string
+}
+
+const mapVariant = (variant: ModalType): ButtonVariants =>
+  variant === 'default' ? 'action' : variant
```
wilysword, line 35: I was wondering this with the basic modal as well, why not just have `action` be the default, instead of having to rename it?

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+    case 'default':
+      return theme.colors.callToAction
+    case 'danger':
+      return theme.colors.error
+    case 'success':
+      return theme.colors.success
+    case 'warning':
+      return theme.colors.warning
+  }
+}
```
wilysword, line 53: You can probably make this code cleaner by using `helpers/variant.ts`. There are examples in the `Table` styles, and if you're curious, the inspiration was taken from https://styled-system.com/variants/.

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  .children > :first-child {
+    word-break: break-all;
+    text-align: center;
+    margin: 0;
+  }
+  .children > div {
+    text-align: center;
+    width: 100%;
+    margin: 16px 0 0 0;
+  }
```
wilysword, line 121: I don't understand the purpose of these styles. I think the dev should be free to stick whatever they want in the modal without worrying about the modal itself changing the styling.
```suggestion
  .children {
    text-align: center;
    width: 100%;
  }
```

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  iconName: PropTypes.oneOf([...iconNames]),
+}
+
+ConfirmModal.defaultProps = {
+  cancelButtonText: 'Cancel',
+  confirmButtonText: 'Confirm',
+  description: undefined,
+  title: 'Modal Title',
+  iconName: undefined,
+  iconSize: 'medium',
```
wilysword, line 151: Should `variant` be here, instead of providing the default in the function definition?

modules/cactus-web/src/Button/Button.tsx
```diff
@@ -190,6 +190,7 @@ const ButtonBase: React.FC<ButtonProps> = ({
   loadingText,
   ...rest
 }) => {
+  console.log(variant)
```
wilysword, line 193: Missed.

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -69,10 +54,11 @@ const ConfirmModalBase: React.FunctionComponent<ConfirmModalProps> = ({
   onConfirm,
   iconSize,
   iconName,
-  variant = 'default',
+  variant,
   title,
   ...props
 }) => {
+  console.log(variant)
```
wilysword, line 61: And here.

PR #266, opened 2020-07-09
CACTUS-132 :: DataGrid (Not a Draft Anymore)

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-132

### Some Notes

- I used the React Context API to help achieve the API structure we defined. Since each column is a child of `DataGrid`, I passed functions in the context down to these children, which are used to push that data back up to the top level `DataGrid` component.
- Developers will be expected to update `paginationOptions` and `sortOptions` and pass them back in through the props; the `DataGrid` does not manage those values internally aside from updating one or two fields at a time and giving the data back to the `onPageChange` or `onSort` handlers respectively.
- I just went ahead and removed `DataGrid` from the exports since this won't be ready to go out until we implement the card view sorting.

### Testing

- Check out this branch & run `yarn web start`
- Make your way to the `DataGrid` storybook and test that the following features have been implemented correctly
    - Sorting based on a column. The sortable columns in the storybook are "Created" and "Active"
    - Pagination. You can toggle the "Provide Page Count" knob to switch between the `Pagination` and `PrevNext` components.
    - Toggle the "Provide Page Size Options" and "Show Results Count" knobs to see what the component looks like with/without those features
    - Change your viewport to test the card view
    - Make sure to test in other browsers such as IE as well!


modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,483 @@
...
+`
+
+const DataColumn = (props: DataColumnProps) => {
+  const { id, title, sortable = false, as: asComponent } = props
+  const { addDataColumn } = useContext(DataGridContext)
+  useEffect(() => {
+    if (addDataColumn && typeof addDataColumn === 'function') {
+      addDataColumn({ id, title, sortable, asComponent })
+    }
+  }, [asComponent, id, sortable, title]) // eslint-disable-line react-hooks/exhaustive-deps
```
mikesoltow, line 450: If this re-renders because one of these change, I think you'll need a way to maintain the column order.

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -141,7 +141,7 @@ const PageLinkBase: React.FC<PageLinkProps> = (props: PageLinkProps) => {
 }
 
 export const Pagination: React.FC<PaginationProps> = (props) => {
-  const { pageCount, currentPage, label, className, onPageChange, linkAs } = props
+  const { pageCount, currentPage, label, className, onPageChange, linkAs, ...rest } = props
```
wilysword, line 144: What's the reason for this change? You mentioned adding classname, but that's already a prop.

Dhalton:
@mikesoltow Thanks for the feedback on this...

> * When I view all 12 rows, I can't scroll all the way up to the top of the table

I saw that too, I figured it was just something with the viewport in storybook like what I had mentioned in Glenn's Table PR before, but I can try to confirm that. Maybe setting `overflow-y` to scroll could help.

> * I'm not sure how I feel about the sorting display in the card view, so I think we should confirm that with Hector.  My sense is that there should be a sorting control above the cards, but I don't know what that would look like 🤷

Yeah something like that would probably be better than leaving the sorting on the header in the card. I'll reach out to Hector today about this.

> * We should probably add some prop types

Totally! Just hadn't had time to polish this up yet 😄 

> * For some reason, the SplitButton dropdown icons are off-center for me (see screenshot)

I see that on my end too it must be from a last second refactor I did before I pushed this up last week where I moved some CSS for the arrow icon into the main component rather than a separate one. I'll just have to be more specific in the selector.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,483 @@
...
+                          if (sortOpt) {
+                            toUpdate.sortAscending = !toUpdate.sortAscending
+                          } else {
+                            sortOptionsCopy[0] = { id: key, sortAscending: false }
+                          }
+                          if (typeof onSort === 'function') {
+                            onSort(sortOptionsCopy)
+                          }
+                        }
+                      : undefined
```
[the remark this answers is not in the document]
Dhalton: Yeah I guess I was just trying to future-proof this for if we ever do end up wanting to support sorting on multiple columns. But we should probably just cross that bridge when we get there. I have a feeling that's part of the problem with the `aria-sort` property...

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,483 @@
...
+                  tabIndex={dataColumn.sortable ? 0 : undefined}
+                >
+                  {dataColumn.title}
+                  {sortOpt !== undefined && <NavigationChevronDown aria-hidden="true" />}
+                </Table.Cell>
+              )
+            })}
+            {Object.keys(columnFns).map((fnKey) => (
+              <Table.Cell key={`col-${fnKey}`}></Table.Cell>
+            ))}
```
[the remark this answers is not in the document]
Dhalton: Okay I should be able to merge them into a single array to loop through rather than keeping them separate

modules/cactus-web/src/Pagination/Pagination.tsx
Dhalton: I didn't see the classname prop, but I noticed that the classname I was setting wasn't being applied, at least on the component I was expecting it to be applied to, so I'll have to revisit this today.

wilysword:
Regarding the storybook centered div thing, there might be a way to fix that: https://github.com/repaygithub/cactus/blob/master/modules/cactus-web/src/SplitButton/SplitButton.story.tsx#L76. It looks like if we add
```
{ cactus: { overrides: { display: 'block' } } }
```
to those stories it doesn't center or clip things anymore.

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: I initially thought you were right about this, but I've tested this myself and the column order doesn't change when you update the props. It's because of how I save them in the component's state. They're stored in an object, so when I update a column's data, it updates the data under that key, but the order of the keys doesn't change.

mikesoltow: Ah cool!  While this does appear to work, I wonder if a `Map` would be a bit safer than a standard object.  After doing a little research, it appears that the order of `Object.keys` is technically not guaranteed (particularly in a cross-browser setting), while it _is_ guaranteed for a Map.  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

mikesoltow: Here's the relevant bit from the doc: `Since ECMAScript 2015, objects do preserve creation order for string and Symbol keys. In JavaScript engines that comply with the ECMAScript 2015 spec, iterating over an object with only string keys will yield the keys in order of insertion.`

I just don't trust IE 😄

Dhalton: Yeah understandable! I can try to do it with a `Map`. I've never tried to save a `Map` in a component's state before, but I bet it's doable

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+              ...paginationOptions,
+              currentPage:
+                direction === 'prev'
+                  ? paginationOptions.currentPage - 1
+                  : paginationOptions.currentPage + 1,
+            })
+          }}
+          disableNext={
+            prevNextProps
+              ? prevNextProps.disableNext || data.length < paginationOptions.pageSize
```
wilysword, line 249: Not sure why, but this doesn't seem to be working for me. I've got 11 items with a page size of 4, and on the third page the button is still enabled.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+        )}
+      </div>
+    )
+  }
+}
+
+const IEStyles = () => {
+  if (typeof window !== 'undefined' && /MSIE|Trident/.test(window.navigator.userAgent)) {
+    return css`
+      width: 100%;
```
wilysword, line 333: This doesn't appear to have a negative effect (at least in Firefox), is there a reason you're conditionally applying it?

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+          pageSizeOptions.map((pageSize) => {
+            const isCurrentPageSize = paginationOptions.pageSize === pageSize
+            return (
+              <li className="page-option" key={`page-size-option-${pageSize}`}>
+                <a
+                  role="link"
+                  aria-selected={isCurrentPageSize ? 'true' : 'false'}
+                  onClick={() => {
+                    onPageChange({ ...paginationOptions, pageSize: pageSize })
+                  }}
```
wilysword, line 379: Like the column headers, these don't respond to keyboard commands.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+      string,
+      { title: string; sortable: boolean; asComponent?: React.ComponentType<any> } | ColumnFn
+    >()
+  )
+
+  useEffect(() => {
+    if (sortOptions.length > 0 && typeof onSort === 'function') {
+      onSort(sortOptions)
+    }
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
Dhalton: I did this so that, if a sort option is passed in on the first render, `onSort` would be called to sort the data immediately. Otherwise, you could have a sort option for a column that says the data is sorted even though it's not.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+                    aria-sort={
+                      column.sortable
+                        ? sortOpt
+                          ? sortOpt.sortAscending
+                            ? 'ascending'
+                            : 'descending'
+                          : 'none'
+                        : undefined
+                    }
+                    tabIndex={!column.sortable ? -1 : undefined}
```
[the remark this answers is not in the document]
Dhalton: I was following the WAI-ARIA example for a data grid with sortable columns https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html

modules/cactus-web/src/DataGrid/DataGrid.tsx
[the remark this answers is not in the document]
wilysword: The dev already had access to that information (they just gave it to you), why would they give you incorrect information? My assumption would be that if they tell you the data's sorted a certain way, then that's how it's sorted.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+                  >
+                    {column.sortable ? (
+                      <span
+                        role="button"
```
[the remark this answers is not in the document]
Dhalton: This was also in the WAI-ARIA example. I'll use that a11y helper

modules/cactus-web/src/DataGrid/DataGrid.tsx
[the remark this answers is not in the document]
wilysword: Looks like they did that to enable traversing the table using the arrow keys, since you haven't done that I don't think it's needed.

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: Hmm, okay I'll look into it. I know I had it working at one point. It might be because of the order of the props. I'm spreading the `prevNextProps` on after this, so I'm thinking that the value from the storybook (default to false) is overriding it. Although that's how I'd want it to work anyway. There's just not an easy way to pass `undefined` for a knob in storybook 😢

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: I'm pretty sure it was causing weird behavior in Chrome, but I'll double check

modules/cactus-web/src/DataGrid/DataGrid.tsx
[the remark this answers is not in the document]
Dhalton: Actually, I'm going to use a button. Easier that way.

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: Yeah that was the issue. Had to do some weird things in the storybook but it'll be fixed when I push up

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: Yeah, setting the width to 100% in causes the page size selection to extend past the table and get cut off a bit:
image

Dhalton:
@wilysword Thanks for reviewing. I pushed up some commits to fix most of the things you mentioned.

> when I unchecked then re-checked "Provide Page Count", it retained the "Prev/Next" component instead of the Pagination component.
Seems like an issue with the storybook, not with the component. I don't know if it re-renders everything when you change knob values...I'll dig in to this more tomorrow and see if I can solve it.

> you can still click the headers in card mode and it will trigger the sort.
Yeah I planned on handling that in the next ticket. I'll remove the button if we're looking at the card view. To do that, I'll need the `ScreenSizeProvider`, which I wasn't planning on working into this until then.

> Using tab will also land on the "real" headers even though they're offscreen, which makes it look like focus has just vanished.
This is a nice catch, I hadn't realized this would happen. However, I'd say that could also be fixed with the next ticket. Since I'll be removing the buttons from the headers when we're in the card view, it should stop happening then.

> Also, if you're in card mode _without_ `fullWidth`, everything centers based on how wide the table is, which is probably not what we want.
I'm not sure what you mean by this...could you post a screenshot? It sounds like what was happening when I set the container to `display: block` in the storybook. But I'm not seeing that now, so I'm a little confused.

wilysword:
I stuck it all in one screenshot 😛. The focus style of the header button is a little weird, and the actual clickable area doesn't match the `cursor: pointer` style.

The card mode non-centered/fullWidth thing does indeed only happen when the container div is `display: block`, but frankly that's a more realistic scenario in my opinion.

datgrid

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,559 @@
...
+    <div className={className}>
+      <DataGridContext.Provider value={{ addDataColumn, addColumn }}>
+        {children}
+        <ResultsViewSection
+          onPageChange={onPageChange}
+          pageSizeSelectLabel={pageSizeSelectLabel}
+          resultsCountText={resultsCountText}
+          paginationOptions={paginationOptions}
+          makePageSizeLabel={makePageSizeLabel}
+        />
```
mikesoltow, line 147: Throwing this out there as food for thought.  Definitely not something I'd ask you to address in this PR...the question is just whether or not we'd want to make a follow-up ticket for this.

This component here `ResultsViewSection` and `Pagination`/`PrevNext` seem to be completely decoupled from the `Table` stuff.  Therefore, I wonder if we should have/let the user add those components themselves rather than bake them into the DataTable.  That could server to both clean up the props signature of `DataGrid` (which takes a ton of props right now) plus give the user the complete flexibility to customize how they display pagination, etc. if they need that.  I'm envisioning something like:

```
<DataGrid>
  <ResultsViewSection
    onPageChange={onPageChange}
    pageSizeSelectLabel={pageSizeSelectLabel}
    resultsCountText={resultsCountText}
    paginationOptions={paginationOptions}
    makePageSizeLabel={makePageSizeLabel}
  />
  <DataGrid.Table
    data={data}
    sortOptions={sortOptions}
    onSort={onSort}
    fullWidth={fullWidth}
  >
    <DataGrid.DataColumn id="name" title="Name" />
    <DataGrid.DataColumn id="created" title="Created" sortable={true} />
    <DataGrid.DataColumn id="active" title="Active" as={BoolComponent} sortable={true} />
  </DataGrid.Table>
  <Pagination
    currentPage={currentPage}
    pageCount={pageCount}
    onPageChange={onPageChange}
    {...otherPaginationPropsYouFeelLikeAdding}
  />
</DataGrid>
```  

`DataGrid` could just be the styled wrapper `div` while `DataGrid.Table` could render the context provider and the `Table` itself.  What do you think?

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+}
+
+DataGrid.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
+  paginationOptions: PropTypes.shape({
+    currentPage: PropTypes.number.isRequired,
+    pageSize: PropTypes.number.isRequired,
+    pageCount: PropTypes.number,
+    pageSizeOptions: PropTypes.arrayOf(PropTypes.number),
+  }).isRequired,
```
mikesoltow, line 513: Should these really be required?  What if there's only 10 rows and no need to paginate?

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,542 @@
...
+  sortOptions: PropTypes.arrayOf(
+    PropTypes.shape({ id: PropTypes.string.isRequired, sortAscending: PropTypes.bool.isRequired })
+  ).isRequired,
```
mikesoltow, line 516: Same thing...what if they don't care about sorting?

Dhalton:
@wilysword Alright, I fixed the issue with the DataGrid being rendered inside a `display: block` container, so the card view should be working correctly now, and I fixed the cursor pointer area. I'm waiting to hear back from Hector on the focus styling of the header buttons

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: Fixed!

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: Fixed!

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: I could go either way on it...
On the one hand, I like how we're doing it now simply because we absorb that complexity into the component so it's less work for other teams, and allowing them to render those components could result in inconsistencies either across applications or even within a single application.
On the other hand, it would obviously give other teams more freedom to lay things out in a way that may work better for them, and I agree that the DataGrid props have grown quite large and it'd be nice to shave them down a bit.

Are you thinking that we would export `ResultsViewSection` as its own component to be used for this case specifically or would that be a component built by another team? At the very least I think it'd be good to export the `PageSizeSelect` if we decide to go this route.

Like I said I could really go either way. If you think this would be a better way of doing it, then let's do it!

Dhalton:
@wilysword Ok I think I've got everything you mentioned check off the list! 🎉 Ready for your re-review whenever.
@mikesoltow Not sure if you want to re-review to check out that latest commit or not. Up to you!

PR #269, opened 2020-07-15
Theme cleanup

NicolasSimmonds:
The main changes are on the theme and Avatar and StatusMessage components.

Let me know if there are other components where you think the ColorStyles object can be used.


modules/cactus-web/src/Avatar/Avatar.tsx
```diff
@@ -83,13 +83,17 @@ const usageMap: UsageMap = {
 }
 
 const variant = (props: AvatarProps) => {
-  const { status } = props
+  const { status, type } = props
 
-  if (status !== undefined) {
+  if (status !== undefined && type === 'alert') {
     return css`
       background: ${avaColor};
```
wilysword, line 90: Pretty sure this is causing a problem, if you look at the snapshot it says things like `background: background-color:hsl(0,0%,90%);`, which is bad CSS. I think you can probably just change it to
```
return avaColor
```

modules/cactus-web/src/Avatar/Avatar.tsx
```diff
@@ -83,13 +83,17 @@ const usageMap: UsageMap = {
...
     `
-  }
+  } else if (status !== undefined && type === 'feedback') {
+    return css`
+      ${avaColor}
+    `
+  } else return
```
wilysword, line 96: I'm not sure if we care or not, but in storybook when I select "usage" = alert and check "disabled", the icon vanishes (color goes to white, no background color set).

modules/cactus-web/src/Avatar/Avatar.tsx
NicolasSimmonds: Yeah you are right, I don't know why thats still there, i thought i changed that. Thanks!

modules/cactus-web/src/Avatar/Avatar.tsx
NicolasSimmonds: Yup this can be corrected

modules/cactus-web/src/Avatar/Avatar.tsx
```diff
@@ -83,12 +83,19 @@ const usageMap: UsageMap = {
...
+    switch (type) {
+      case 'alert':
+        return css`
+          background: ${avaColor};
+        `
+      case 'feedback':
+        return avaColor
+    }
   }
 }
```
wilysword, line 100: This whole function duplicates some of the logic from the `avaColor` function. If you changed that function to include the `background` declaration where applicable, you could get rid of one of the two functions.
```
return `background: ${props.theme.colors.status.avatar.error};`
```

modules/cactus-theme/src/theme.ts
```diff
@@ -348,7 +349,7 @@ function fromTwoColor({
         },
         callToAction: {
           backgroundColor: callToAction,
-          color: white,
+          color: callToActionText,
```
mikesoltow, line 352: You probably want to change line 316 as well so they're both referencing the same thing

NicolasSimmonds: Done

PR #271, opened 2020-07-17
Textarea min width

NicolasSimmonds:


NicolasSimmonds:
I saw it happening in very tiny screens like Iphone 5. So thats why I applied the min-width to screens larger than tiny. But we could consider a even smaller mediaquery if you like.

Dhalton:
No I don't think we need a smaller media query. If it was happening on iphone 5 then that's good enough for me

PR #281, opened 2020-07-23
CACTUS-98 MenuBar Component

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-98

Most of what you need to know is on the ticket. In the end, I didn't manage to get any useful unit tests, since almost all the logic requires visible elements. The "sidebar" menu at smaller screen sizes has significantly different styles, which may require some structural changes as well, so I'm considering some changes that might alleviate that. (By the way, I want to be the one to work on that ticket.)

As a semi-interesting side note, I realized that a lot of our props interfaces are wrong, or at least over-complicated. We mostly use `React.DetailedHTMLProps`, but it would actually be sufficient to use `React.HTMLAttributes<T>` where T is the element type (e.g. `HTMLAnchorElement`). The detailed props only really adds two items: `key`, which I don't think actually makes it into the props; and `ref`, which most of the time we don't want to be implicitly added.


modules/cactus-web/src/MenuBar/scroll.ts
```diff
@@ -0,0 +1,335 @@
...
+          const parentRect = wrapper.getBoundingClientRect()
+          const correction = BUTTON_WIDTH - 0.1
+          const absBack = parentRect[back] + correction
+          const absFore = parentRect[fore] - (index + 1 === items.length ? 0 : correction)
+          if (itemRect[back] < absBack) {
+            next = index
+            if (next === 1 && items[0].getBoundingClientRect()[width] < BUTTON_WIDTH) {
+              next = 0
+            }
+          } else if (itemRect[fore] > absFore) {
```
mikesoltow, line 308: In the stories, when I look at a dropdown with 8 items and plenty of room to display them all, a scroll button appears after I scroll down to item number 7 (index 6) using the arrow keys.  It's hitting this block when presumably it shouldn't be.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -0,0 +1,340 @@
...
+const TextWrapper = styled.div`
+  flex-grow: 1;
+`
+
+const Menu: React.FC<MenuProps> = ({ children, expanded }) => {
+  const [scroll, menuRef] = useScrollButtons('vertical')
+  return (
+    <MenuWrapper expanded={expanded}>
+      <ScrollButton show={scroll.showBack} onClick={scroll.clickBack} onBlur={killEvent}>
+        <NavigationChevronLeft />
```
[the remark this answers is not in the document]
wilysword: I guess I mostly did it because then it parallels the structure of the horizontal menubar, but I wasn't really thinking deeply; you're right that up/down would be better.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -0,0 +1,340 @@
...
+  overflow: hidden;
+
+  & > li > [role='menuitem'] {
+    position: relative;
+    top: ${(p) => p.top || 0}px;
+    left: ${(p) => p.left || 0}px;
+  }
+`
+
+const MenuButton = styled.button.attrs({ tabIndex: -1, role: 'menuitem' })`
```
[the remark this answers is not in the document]
wilysword: You're not supposed to be able to. We [talked in Slack](https://repay.slack.com/archives/CABGTU4S3/p1594232111019300) about accessibility recommendations for this kind of component, and decided to follow the recommendation that the entire thing should be treated as a single tabbable entity. Use the arrow keys to navigate instead.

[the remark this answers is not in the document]
mikesoltow: Oh okay...I misunderstood that initial discussion and thought you were talking about a dropdown rather than the whole thing.  But since this is developed according to the guidelines, that's fine.

modules/cactus-web/src/MenuBar/scroll.ts
```diff
@@ -0,0 +1,329 @@
...
+export function useSubmenuToggle(): [boolean, ToggleSubmenu] {
+  const [expanded, setExpanded] = React.useState<boolean>(false)
+  const toggle = React.useCallback<ToggleSubmenu>(
+    (menuButton, open, focus = false) => {
+      setExpanded((isExpanded: boolean) => {
+        if (open !== isExpanded) {
+          isExpanded = open === undefined ? !isExpanded : open
+          // Was closed, now open
+          if (isExpanded && menuButton) {
+            setTimeout(() => {
```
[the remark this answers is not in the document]
wilysword: The `setTimeout` is necessary because the size calulations can't take place until the element is visible. Returning `isExpanded` will update the state, which will change the `display` from `none` to `flex`, and only then will we know how big the submenu is. My original plan was to have the elements start relative to their parent list item, which would result in a much smaller shift (or no shift at all), but because of a quirk in how `position: absolute` and `overflow: hidden` work, that kind of hierarchy wasn't possible.

The only way I could get around that would be to set `display: none` on the "off-screen" buttons, instead of `overflow: hidden`, but that would make calculations impossible since I wouldn't know how big the buttons are.

modules/cactus-web/src/MenuBar/scroll.ts
wilysword: Yeah, I've seen a number of quirks with that, so I guess I'll just have to keep tweaking the algorithm.

wilysword:
> I think we should probably make use of the keyPressAsClick helper so that when you press Enter on a menu item, the item is selected.

Was this not working for you? At least in Firefox, hitting {enter} or {space} on a `button` element runs the click handler, and I though that was common across browsers.

The divots are because I added a full border but made everything except the bottom transparent, so adding the focus style wouldn't shift everything. I can play around with other ways of getting the same effect.

wilysword:
Okay, regarding the enter thing I'll have to look into it a bit more. It appears maybe the event is getting canceled before it gets to the proper handler. (And space is just the opposite, causing two events instead of one.)

mikesoltow:
Quick thing that I noticed when reviewing Nicolas's layout PR: it seems like we should be extending the MenuBar to 100% of the width of its container, even if there aren't enough items to fill it (unless I am misinterpreting the style guide).  If we do indeed decide to do that, I'm happy to split that out into another ticket (so this PR can focus on just fixing the scrolling).

wilysword:
@mikesoltow The nav is already styled at 100% width, so I'm not clear on what you feel is lacking...

mikesoltow:
@wilysword Mainly a border across the bottom of the bar, as this is what it looks like when incorporated into the CLI
Screen Shot 2020-08-04 at 3 20 48 PM

wilysword:
@mikesoltow I can't actually think of any way to do that. The border changes color on hover, which means it has to be applied to the individual buttons. If I _also_ add a border on the nav, there's no way to get the button borders to overlap the nav border. Positioning won't work, because everything is overflow hidden. Using ::after had the same problem where the hover style wasn't visible. If you can think of a way to accomplish this, I'm open to suggestions.

mikesoltow:
@wilysword Because I don't have the CSS skills to pull anything like that off, I went the plagiarism route.  If you look at GitHub's nav bar, they use the following styles on their `nav` element to get that effect:

`box-shadow: inset 0 -1px 0 {insert color}`

I tried it out quickly on the MenuBar, setting the color to lightContrast, and it appears to work, but I didn't poke too hard around edge cases or anything like that 🤷

wilysword:
Alright, I think I've fixed all the issues that are applicable to this ticket.

wilysword:
I managed to get it recognizing the correct props for any value of `as` that I tried, including `ref` on `forwardRef` components. I did _not_ get it to recognize `ref` on class components, but since that's a pretty uncommon use case (I would think) and I've already spent two and a half days on this, I'm just going to leave it.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -47,26 +42,38 @@ const preventAction = (event: React.KeyboardEvent<HTMLElement>) => {
...
+  const original = propsCopy.onKeyUp
   propsCopy.onKeyUp = !original
     ? preventAction
     : (e: React.KeyboardEvent<HTMLElement>) => {
         original(e)
         preventAction(e)
       }
+  // Overriding these two can completely screw up the algorithms used.
+  delete propsCopy.tabIndex
+  delete propsCopy.role
```
mikesoltow, line 64: Since we are silently deleting these props, I wonder if we should log a warning to the console if they are passed in to indicate to the user that they are passing in invalid props that we are not respecting.  I would say that maybe `props` should be an `Omit` type so that these two values aren't included, but I don't know if that would screw this up.

wilysword: I don't feel strongly either way. It's mentioned in the docs, but I guess some weirdos don't read those. As far as Typescript, I originally had it like that (before I moved `AsProps` to a different file), but I took it out because it felt semi-pointless: (a) I was deleting the props anyway to account for those who don't use Typescript; (b) if they use a custom component instead of an intrinsic element (which is probably the more common use case) there's no guarantee they'd accept those props anyway, or that they'd have the same names.

mikesoltow: Fair enough.  I'm happy to keep the TypeScript stuff out...not least to keep the complexity down.  I think the console warnings would probably be helpful if you don't mind adding those in really quickly.

wilysword: Added the warnings via proptypes. Also fixed a bug that was introduced when we got rid of the big `setTimeout` call relating to submenu visibility.

luisniebla:
I wish we had allowed the title prop on MenuBar.Item.

mikesoltow:
@luisniebla We could potentially look into that, but I'd be curious to know how that would help, etc.

luisniebla:
It just would've made the transition from semantic to cactus-web like 5% easier. It's already done so it's not a big deal.

PR #286, opened 2020-07-28
fix(cactus-web): fix ConfirmModal text overflow

NicolasSimmonds:


modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -100,6 +100,7 @@ export const ConfirmModal = styled(ConfirmModalBase)`
     }
   }
   .children {
+    word-break: break-word;
```
mikesoltow, line 103: Unfortunately this value of `word-break` is deprecated and does not work in IE11.  This should be replaced with `overflow-wrap: break-word;'.

modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -100,6 +100,7 @@ export const ConfirmModal = styled(ConfirmModalBase)`
     }
   }
   .children {
+    overflow-wrap: break-word;
```
mikesoltow, line 103: Despite assurances from MDN, this still doesn't work in IE11.  But it looks like you can change this to the following:

```suggestion
    word-wrap: break-word;
    overflow-wrap: break-word;
```

PR #288, opened 2020-07-30
Getting started docs

NicolasSimmonds:
To review the docs

Launch the docs website: yarn docs start

On the sidebar there will be a Getting Started tab and there you'll find the docs.

There is also a link on the main page on the Getting Started section. This option is only available in the published version of the Website though

### DIFF
--- a/docs/Framework/README.md
+++ b/docs/Framework/README.md
@@ -1,5 +1,5 @@
 ---
-order: 1
+order: 2
 ---
 
 # Cactus Framework
--- a/docs/Getting-Started/README.md
+++ b/docs/Getting-Started/README.md
@@ -10,12 +10,12 @@ The best way to get started with Cactus, is to use our CLI `@repay/create-ui`. T
 
 The CLI will generate a boilerplate with a basic configuration depending on the arguments you pass or the answers you provide to the prompts. The options are
 
-| Argument     | Action                                                                              |
-| ------------ | ----------------------------------------------------------------------------------- |
-| `<app-name>` | The name for your app. You'll find the generated code under a folder with this name |
-| --javascript | Generate the app with javascript configuration                                      |
-| --typescript | Generate the app with typescript configuration                                      |
-| --git, -g    | Initialize a git repository                                                         |
+| Argument       | Action                                                                              |
+| -------------- | ----------------------------------------------------------------------------------- |
+| `<app-name>`   | The name for your app. You'll find the generated code under a folder with this name |
+| `--javascript` | Generate the app with javascript configuration                                      |
+| `--typescript` | Generate the app with typescript configuration                                      |
+| `--git, -g`    | Initialize a git repository                                                         |
 
 For example:
 
--- a/docs/Getting-Started/README.md
+++ b/docs/Getting-Started/README.md
@@ -1,16 +1,20 @@
-# Getting Started
+---
+order: 1
+---
+
+# Getting Started with a new application
 
 The best way to get started with Cactus, is to use our CLI `@repay/create-ui`. This tool provides a basic configuration and structure to start implementing a website using all the cactus resources.
 
-## The CLI
+## Using the CLI
  The CLI will generate a boilerplate with a basic configuration depending on the arguments you pass or the answers you provide to the prompts. The options are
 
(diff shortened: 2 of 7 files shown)


docs/Getting-Started/README.md
```diff
@@ -0,0 +1,29 @@
...
+| --git, -g    | Initialize a git repository                                                            |
+
+For example:
+```shell
+create-repay-ui myapp --typescript -g
+```
+Will generate an app under `myapp` folder, with typescript and a git repository.
+
+
+## How to start
```
mikesoltow, line 22:
```suggestion
## Installing the CLI
```

docs/Getting-Started/README.md
```diff
@@ -0,0 +1,29 @@
+# Getting Started
+
+The best way to get started with Cactus, is to use our CLI `@repay/create-ui`. This tool provides a basic configuration and structure to start implementing a website using all the cactus resources.
+
+## The CLI
```
mikesoltow, line 5:
```suggestion
## Using the CLI
```

docs/Getting-Started/README.md
```diff
@@ -0,0 +1,29 @@
+# Getting Started
```
mikesoltow, line 1: I wonder if we actually want to rename this to "Getting Started with a New Application."  As many users will get started with existing code and wouldn't use this CLI.  We can then point those users to the example apps on the docs homepage.

docs/README.md
```diff
@@ -2,6 +2,10 @@
 
 An application framework and design system built in React at [REPAY](https://github.com/repaygithub).
 
+
+## Getting Started
+To review the documentation for `@repay/create-ui` visit [this link](../Getting-Started/README.md)
```
mikesoltow, line 7: We could then change this to something like:
```suggestion
To get started with a brand new UI application, visit [this link](../Getting-Started/README.md).  For guidance on using Cactus with an existing application, browse the module documentation and the example applications linked below.
```

docs/Getting-Started/README.md
mikesoltow, line 1: Could we adjust the order so that this appears at the top of the sidebar menu (or at least as the second item, since `Design Principles` is always pegged to be at the top)?  To do that, it looks like you need to add an order parameter to the README.  See the README for the Framework, which is currently pegged at #1.  It would be nice if we could squeeze Getting Started in right before that.

docs/Framework/README.md
```diff
@@ -1,5 +1,5 @@
 ---
-order: 1
+order: 2
```
mikesoltow, line 2: You'll also need to edit the order of the Theme, Web, Icons, and I18n docs

docs/Getting-Started/README.md
```diff
@@ -1,16 +1,20 @@
-# Getting Started
+---
+order: 1
+---
+
+# Getting Started with a new application
```
mikesoltow, line 5: Could you capitalize `New Application`?  That will make it look more in sync with the other items on the side menu.  Thanks!

NicolasSimmonds: @mikesoltow Done

docs/Getting-Started/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Using the CLI
+
+The CLI will generate a boilerplate with a basic configuration depending on the arguments you pass or the answers you provide to the prompts. The options are
+
+| Argument     | Action                                                                              |
+| ------------ | ----------------------------------------------------------------------------------- |
+| `<app-name>` | The name for your app. You'll find the generated code under a folder with this name |
+| --javascript | Generate the app with javascript configuration                                      |
+| --typescript | Generate the app with typescript configuration                                      |
+| --git, -g    | Initialize a git repository                                                         |
```
wilysword, line 18: These options should also be surrounded by backticks.
```suggestion
| `--javascript` | Generate the app with javascript configuration                                      |
| `--typescript` | Generate the app with typescript configuration                                      |
| `--git`, `-g`    | Initialize a git repository                                                         |
```

NicolasSimmonds: Done, thanks!

PR #291, opened 2020-08-04
CACTUS- 281 ::  TypeScript Linting

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-281

I would definitely recommend just skimming through the actual changes, as they are extensive.

### Testing
1. Check out this branch and run `yarn install`.
2. Run `yarn cleanup && yarn build` at the root to clean & rebuild all modules.
3. Run `yarn test` at the root to:
  - Test types for all modules
  - Run all tests (including integration)
  - Run the linter
4. Make sure that you don't encounter any errors during the tests, type tests, or linting.

If you have any questions, just let me know!


wilysword:
So I made a scratch branch and merged this into my Menubar stuff to see how much work will need to be done there, and I have a question as a result. You added a _ton_ of explicit return types. (Since that didn't happen automatically for me, I assume you did that manually? Or maybe I'm missing some config for it to happen automatically.) However, the only rule I saw was that _exported_ functions need to have explicit return types, not that _all_ functions do. Do you have any insight on this?

Dhalton:
> So I made a scratch branch and merged this into my Menubar stuff to see how much work will need to be done there, and I have a question as a result. You added a _ton_ of explicit return types. (Since that didn't happen automatically for me, I assume you did that manually? Or maybe I'm missing some config for it to happen automatically.) However, the only rule I saw was that _exported_ functions need to have explicit return types, not that _all_ functions do. Do you have any insight on this?

I did have to do it manually -- fun times! When I was working on this, there was a rule enforcing that all functions have an explicit return type. However, after I merged master into this branch, which had Mike's recent upgrades, I noticed different rules were being enforced. Turns out, before I had merged in those updates, the linter was using the version of `@typescript-eslint/eslint-plugin` that is installed for the Gatsby, *not* the version installed in our eslint config. Now that the version is resolved to 3.7.1, what it should be, it looks like that's the case -- only exported functions need explicit types. Which means, of course, that a lot of those return types probably weren't necessary. 😢 Fortunate for you, though, that should mean less work on your end.

mikesoltow:
> Which means, of course, that a lot of those return types probably weren't necessary.

Oh 💩  are you serious?  Man, I feel terrible how this ticket turned out.  You are the man for doing this!

Dhalton:
> Oh 💩 are you serious? Man, I feel terrible how this ticket turned out. You are the man for doing this!

Haha not your fault! It happens. It's kind of on me for not noticing that it was using the wrong version sooner. Live and learn, I guess!

PR #298, opened 2020-08-13
CACTUS-267: Tooltip

wilysword:
Make `Tooltip` accept JSX; stop ignoring `position` prop. Thanks to the change I made previously with `FieldProps`, the change to the prop type in `AccessibleField` is propagated to all the components that use it. There was quite a bit of unused/unnecessary complexity in the types in that file.

BREAKING CHANGE: Removed unused props from `Tooltip`; the component's behavior is unchanged so it shouldn't break anything functionally, but the type signature is different so it could break Typescript builds. As a result, I had to remove the `children` from the `Tooltip` stories.

(Note that if I were designing the interface, I would have made the `children` be the content of the tooltip, and had an optional prop for changing the icon.)

### DIFF
--- a/modules/cactus-web/src/Tooltip/Tooltip.tsx
+++ b/modules/cactus-web/src/Tooltip/Tooltip.tsx
@@ -1,4 +1,4 @@
-import { PRect } from '@reach/rect'
+import { Position } from '@reach/popover'
 import { TooltipPopup as ReachTooltipPopup, useTooltip } from '@reach/tooltip'
 import VisuallyHidden from '@reach/visually-hidden'
 import { NotificationInfo } from '@repay/cactus-icons'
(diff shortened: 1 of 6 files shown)


modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -1,66 +1,38 @@
...
-}
-
-type Position = (
-  triggerRect: PRect | null | undefined,
-  tooltipRect: PRect | null | undefined
-) => DOMRect
 
-interface TooltipProps
-  extends MarginProps,
-    Omit<React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'ref'> {
```
mikesoltow, line 31: I think I'm okay with this change, especially with this component, but we might want to be careful more generally with removing potential props from developers.  Could be worth adding `Read.AriaAttributes` back in, although most likely the devs won't need that much control.

mikesoltow: Actually, just ignore me...now that I think about it, `AriaAttributes` wouldn't do anything if we don't pass them down, because our component isn't an actual DOM element.  Carry on...I'm going to get more ☕

wilysword: It wouldn't actually make a difference either way, I don't think, because Typescript doesn't seem to validate props with hyphens. (Or at least, it doesn't complain if you add one that doesn't exist; I'm not sure if it checks types for those that _do_ exist.)

PR #299, opened 2020-08-18
Integration test menu bar overflow

NicolasSimmonds:
Run: 
- yarn w mock-ebpp test with a specified browser
 - yarn w mock-ebpp test:ci

Check that everything passes as expected

### DIFF
--- a/examples/mock-ebpp/tests/menubar-integration.test.ts
+++ b/examples/mock-ebpp/tests/menubar-integration.test.ts
@@ -44,8 +44,9 @@ test('Interact with dropdown', async (t: TestController): Promise<void> => {
 test('Can click item when MenuBar is overflowed', async (t: TestController): Promise<void> => {
   const innerHeight = await getWindowHeight()
   await t.resizeWindow(1024, innerHeight)
-  await t.click(Selector('nav').child('div').nth(-1))
-  await t.click(Selector('nav').child('div').nth(-1))
+  const overflowedItem = Selector('nav').child('div').nth(-1)
+  await t.click(overflowedItem)
+  await t.click(overflowedItem)
   await t.click(Selector('button').withText('Explore our modules on GitHub'))
   await t.expect(queryByText('Cactus Framework').exists).ok()
 })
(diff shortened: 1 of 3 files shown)


examples/mock-ebpp/tests/menubar-integration.test.ts
```diff
@@ -38,3 +40,12 @@ test('Interact with dropdown', async (t: TestController): Promise<void> => {
   const url = await getUrl()
   await t.expect(url).eql('http://localhost:33567/account/45789')
 })
+
+test('Can click item when MenuBar is overflowed', async (t: TestController): Promise<void> => {
+  const innerHeight = await getWindowHeight()
+  await t.resizeWindow(1024, innerHeight)
+  await t.click(Selector('nav').child('div').nth(-1))
```
Dhalton, line 47: You could probably save the value from `Selector('nav').child('div').nth(-1)` into a variable and pass that to `click` instead of finding the element twice, but that's not really a huge deal imo.

NicolasSimmonds: I'll change this, also think it might be better!

examples/mock-ebpp/tests/menubar-integration.test.ts
```diff
@@ -38,3 +40,13 @@ test('Interact with dropdown', async (t: TestController): Promise<void> => {
   const url = await getUrl()
   await t.expect(url).eql('http://localhost:33567/account/45789')
 })
+
+test('Can click item when MenuBar is overflowed', async (t: TestController): Promise<void> => {
+  const innerHeight = await getWindowHeight()
+  await t.resizeWindow(1024, innerHeight)
+  const overflowedItem = Selector('nav').child('div').nth(-1)
```
mikesoltow, line 47: I'd rename this to `scrollIcon` or something like that.  I initial thought `overflowedItem` meant an item that was cut off, so I was confused why it was being clicked twice.

modules/cactus-web/src/DateInput/DateInput.story.tsx
```diff
@@ -25,6 +25,7 @@ storiesOf('DateInput', module)
       >
         {({ value, onChange }): ReactElement => (
           <DateInput
+            disabled
```
mikesoltow, line 28: This looks like something that shouldn't be here 😄

NicolasSimmonds: Wow, that is definitely not supposed to be there, sorry

PR #301, opened 2020-08-18
CACTUS-267 DataGrid, PrevNext, & FileInput

wilysword:
FileInput also has a `labels` prop but I couldn't change those because they're used as both JSX and an `aria-label`, so I had to leave them as strings.


modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -309,11 +309,7 @@ const FileInputBase = (props: FileInputProps): React.ReactElement => {
     rawFiles,
     name,
     accept,
-    labels = {
-      delete: 'Delete File',
-      loading: 'Loading',
-      loaded: 'Successful',
-    },
```
mikesoltow, line 316: Curious as to the reasoning here...what's the reason not to just sync up with `defaultProps`?

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -22,14 +16,14 @@ import Table from '../Table/Table'
 
 interface DataGridContextType {
   addDataColumn: (dataColumn: DataColumn) => void
-  addColumn: (key: string, columnFn: ColumnFn, title?: string) => void
+  addColumn: (key: string, columnFn: ColumnFn, title?: React.ReactChild) => void
```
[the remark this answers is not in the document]
wilysword: `ReactChild` has to be a single element, a string, or a number. `ReactNode` can also be an array of elements, a boolean, or null. The array could be useful, but isn't really needed: worst case, they can wrap whatever they've got in a `React.Fragment`. Boolean or null don't make sense. To be honest, I kind of feel like I should have been using `ReactChild` all along, but there were some places where doing so inexplicably caused a Typescript error in proptypes, and it didn't seem an important enough distinction for me to spend much time on it.

modules/cactus-web/src/FileInput/FileInput.tsx
wilysword: I saw no reason to define the same defaults in two places (actually, three places for `loading` & `loaded`, since they also have defaults where they're used). If they're not passed they'll be pulled from defaultProps, so as far as I can tell this code would never be needed.

PR #302, opened 2020-08-19
fix(cactus-web): fix DateInput colors

NicolasSimmonds:
run cactus web and check the background color of the calendar popup

Check the color of the month/year:  Darkest contrast when it's closed, CallToAction when it's open

### DIFF
--- a/modules/cactus-web/src/DateInput/DateInput.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.tsx
@@ -356,6 +356,10 @@ const MonthYearSelect = styled.button`
   justify-content: space-between;
   align-items: center;
 
+  &[aria-label='Click to use calendar picker'] {
+    color: ${(p) => p.theme.colors.callToAction};
+  }
+
   [data-is-open='month-year'] {
     transform: rotate3d(1, 0, 0, 180deg);
   }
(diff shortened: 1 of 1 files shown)


modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -356,6 +356,10 @@ const MonthYearSelect = styled.button`
   justify-content: space-between;
   align-items: center;
 
+  &[aria-label='Click to use calendar picker'] {
```
wilysword, line 359: Given that this label should be translatable (if not now, then in the future), I'd prefer a different CSS selector if you can find one.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1524,6 +1528,7 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
                       }}
                     >
                       <MonthYearSelect
+                        className={isOpen}
```
wilysword, line 1531: Based on the variable name I would think this is a boolean, are you sure this is what you meant for a CSS class?

NicolasSimmonds: Yeah I had this thought too. But this variable is a string (I didn't create it) and has to possible values: "calendar' and 'month-view'

PR #304, opened 2020-08-24
Match disabled styles

NicolasSimmonds:


modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -710,6 +710,7 @@ interface DateInputState {
   isOpen: false | 'month-year' | 'calendar'
   // YYYY-MM-dd
   focusDay?: string
+  disabled?: boolean
```
mikesoltow, line 713: I don't think you need to add this to the state type.

modules/cactus-web/src/DateInputField/DateInputField.tsx
```diff
@@ -49,6 +51,7 @@ function DateInputFieldBase(props: DateInputFieldProps): React.ReactElement {
       {({ fieldId, status, labelId, ariaDescribedBy }): React.ReactElement => (
         <DateInput
           {...rest}
+          disabled={disabled}
```
mikesoltow, line 54: If you make the change I suggested earlier, I would pull this value from the render function args (just in case the `useAccessibleField` hook ever evolves to have control over the `disabled` state of the child component).

modules/cactus-web/src/RadioButtonField/RadioButtonField.tsx
```diff
@@ -80,6 +80,7 @@ export const RadioButtonField = styled(RadioButtonFieldBase)`
   ${Label} {
     cursor: ${(p): string => (p.disabled ? 'not-allowed' : 'pointer')};
     padding-left: 8px;
+    color: ${(p) => p.disabled && p.theme.colors.lightGray};
```
mikesoltow, line 83: This should be `mediumGray`

modules/cactus-web/src/SelectField/SelectField.tsx
```diff
@@ -51,6 +53,7 @@ const SelectFieldBase: React.FC<SelectFieldProps> = (props): React.ReactElement
       {({ fieldId, labelId, name, ariaDescribedBy, status }): React.ReactElement => (
         <Select
           {...rest}
+          disabled={disabled}
```
mikesoltow, line 56: Same comment here about potentially pulling from the function args

modules/cactus-web/src/TextArea/TextArea.tsx
```diff
@@ -76,8 +77,12 @@ const Area = styled.textarea<TextAreaProps>`
...
   }
 
-  &::placeholder {
-    color: ${(p): string => p.theme.colors.mediumContrast};
+  &:disabled {
+    cursor: not-allowed;
+  }
+
+  &:disabled::placeholder {
+    color: ${(p): string => p.theme.colors.mediumGray};
```
mikesoltow, line 85: This color should apply to the regular text too, not just the placeholder

modules/cactus-web/src/TextAreaField/TextAreaField.tsx
```diff
@@ -67,6 +69,7 @@ const TextAreaFieldBase = (props: TextAreaFieldProps): React.ReactElement => {
     >
       {({ fieldId, status, ariaDescribedBy }): React.ReactElement => (
         <TextArea
+          disabled={disabled}
```
mikesoltow, line 72: Same comment about pulling from the function

modules/cactus-web/src/TextInput/TextInput.tsx
```diff
@@ -74,11 +74,15 @@ const Input = styled.input<InputProps>`
...
 
-  [disabled] {
+  &:disabled {
+    cursor: not-allowed;
     border-color: ${(p): string => p.theme.colors.lightGray};
     background-color: ${(p): string => p.theme.colors.lightGray};
   }
 
+  &:disabled::placeholder {
+    color: ${(p): string => p.theme.colors.mediumGray};
```
mikesoltow, line 84: Like the text area, I think this color needs to apply to regular text as well

modules/cactus-web/src/TextInputField/TextInputField.tsx
```diff
@@ -68,6 +70,7 @@ const TextInputFieldBase = (props: TextInputFieldProps): React.ReactElement => {
       {({ fieldId, status, ariaDescribedBy }): React.ReactElement => (
         <TextInput
           {...inputProps}
+          disabled={disabled}
```
mikesoltow, line 73: Same comment here about pulling from the function

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -108,7 +113,9 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
       <Label {...props.labelProps} id={labelId} htmlFor={fieldId}>
         {props.label}
       </Label>
-      {props.tooltip && <Tooltip label={props.tooltip} id={tooltipId} maxWidth={maxWidth} />}
+      {props.tooltip && (
+        <Tooltip label={props.tooltip} id={tooltipId} maxWidth={maxWidth} disabled={disabled} />
+      )}
       {typeof props.children === 'function'
         ? props.children(accessibility)
         : React.cloneElement(React.Children.only(props.children), {
```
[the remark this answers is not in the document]
Dhalton: Totally on board with this btw! Would be less confusing if the disabled prop were passed to `cloneElement` rather than a developer being required to pass it to both `AccessibleField` and their element. 👍

PR #305, opened 2020-08-24
CACTUS-306 :: Footer

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-306

I tried to do this a few different ways but ultimately I ended up using the same strategy I used on the DataGrid to take data from the sub-components and pass them back up to the parent (Footer) component, which just involves providing some functions to the child components via context and calling them to save data.

### Testing

Bring up the Footer in storybook and set some custom text, add/remove some links and make sure the links are being positioned correctly. Check that it looks good at different screen sizes, and make sure to test in IE11 as well.

Let me know if you have any questions/suggestions or anything like that


modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,307 @@
...
+                <React.Fragment key={`link-col-${colIndex}`}>
+                  <LinkCol numCols={dividedLinks ? dividedLinks.length : 1}>
+                    {links.map((link: LinkType, linkIndex: number) => (
+                      <Link key={`list-link-${linkIndex}`} to={link.to}>
+                        {link.content}
+                      </Link>
+                    ))}
+                  </LinkCol>
+                  {dividedLinks !== undefined && colIndex !== dividedLinks.length - 1 && (
+                    <Divider />
```
wilysword, line 238: Couldn't you just style the columns with border-left (and/or border-right) and the remove the border using first/last child?

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,307 @@
...
+      )}
+    </div>
+  )
+}
+
+interface MainSectionProps {
+  children: React.ReactNode
+}
+
+export const MainSection = (props: MainSectionProps): React.ReactNode => {
```
wilysword, line 254: I'm not sure this component is needed: as long as the `Footer.Link` components return null, you could just use the Footer's `props.children` directly as the main content.

Dhalton:
@wilysword 
> Instead of adding the REPAY logo to the repo, is there perhaps a REPAY logo in a CDN somewhere that we could use?

Not sure, but I can try to find out.

> I'm not sure I see the point of all the complexity/width-checking etc. with the links. The way I see it, you have a max number of columns for a given size (which the ScreenSizeContext tells you). Divide the links as evenly as possible between the columns. Don't give the columns a max-width so that if there are fewer than the max, they can expand to take up as much room as they need; `flex-grow` and/or `flex-shrink` may also be needed, I'm not too sure how exactly those work.

I'm trying this suggestion out right now. The function to divide links will work fine, but I'm just having trouble making the links look like the design with your suggestion. I think we actually _should_ provide either a max-width or just explicitly set the width on each column -- in the design, when there are fewer links than the max number of columns, the columns don't grow to fill the extra space; they stay the same size as they would if it had the max number of columns. I'd think we should set the width to the available width / max number of columns regardless of how many columns there are. That would fit the design better imo.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,307 @@
...
+interface LinkContainerProps {
+  isListView: boolean
+}
+
+interface LinkColProps {
+  numCols: number
+}
+
+interface FooterProps extends React.HTMLAttributes<HTMLDivElement> {
+  src?: string
```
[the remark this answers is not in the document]
Dhalton: I guess I'd be okay with that but I'm having trouble seeing how it'd work. I have a few questions. Maybe you can provide some clarification...but anyone else who sees this feel free to chime in with your opinions.

1. As far as I know, you can't pass a component to `src` on an `img` tag (maybe I'm wrong though?)...so that would mean they'd pass a component that renders an `img` in its JSX somewhere, right?
2. Assuming that's the case, should we put something in the docs recommending that the props passed to said component be mapped to the top-level element? (So we can ensure that our className is passed on correctly, enabling us to add the spacing around the logo)
3. Would we want to change this prop name at all to reflect the component option or should we leave it as `src`?

wilysword:
I'm okay with that, although I didn't actually mean "grow to fill the available space"; I meant "grow with the content. So if you had two links and one was very long, it could take _up to_ half the space before it started wrapping, while the other would stay small. I _think_ it's possible to do that with flex, but if not then I'd be fine with a max-width instead.

modules/cactus-web/src/Footer/Footer.tsx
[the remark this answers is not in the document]
wilysword: 1. Correct, although I think they could just as easily do an `svg` as an `img`.
2. If all we need to do is add a margin, we could just wrap it in a styled div regardless of whether it was an `img` or custom component.
3. I'd call it `logo`.

Dhalton:
Alright, got some changes pushed up. The code is looking quite a bit cleaner so thanks for the suggestions, Glen. Still waiting to hear back from Emily about whether or not there's a logo we can pull from a CDN somewhere though.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,208 @@
...
+  children: React.ReactNode
+  to: string
+}
+
+export const FooterLink = (props: LinkProps): React.ReactNode => {
+  const { children, to } = props
+  const { addLink } = useContext(FooterContext)
+  useEffect(() => {
+    addLink(children, to)
+  }, [children]) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
Dhalton: Hmm, that's a good point. I'll update it to use a key system

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,208 @@
...
+      divided[i].push(links[linkIndex++])
+    }
+  }
+  return divided
+}
+
+const FooterBase = (props: FooterProps) => {
+  const { logo: Logo, className, children } = props
+  const [state, setState] = useState<FooterState>({
+    links: [],
```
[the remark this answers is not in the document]
Dhalton: Right, yeah there used to be more in the state here but it was removed. I'll switch this to an array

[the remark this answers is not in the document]
Dhalton: Well, actually it might end up being a `Map`, but you know what I mean 😄

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,208 @@
...
+interface FooterState {
+  links: LinkType[]
+}
+
+interface LinkColProps {
+  maxCols: number
+}
+
+interface FooterProps extends React.HTMLAttributes<HTMLDivElement> {
+  logo?: string | React.ComponentType<any>
```
wilysword, line 28: Since you can't pass props, you should probably make this a `React.ReactElement<any>` instead, that way they have more flexibility in what kinds of components they can use.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,208 @@
...
+  ${(p) => p.theme.mediaQueries && p.theme.mediaQueries.small} {
+    margin-right: 16px;
+    margin-bottom: 0px;
+  }
+`
+
+const Img = styled('img')`
+  width: 138px;
+  height: 40px;
+`
```
wilysword, line 62: This seems awfully restrictive/specific. Not all logos will have the same aspect ratio; we could probably have a max height/width to prevent it from getting too crazy, but I see no need to specifically dictate the size of other people's logos.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,208 @@
...
+}
+
+export const Footer = styled(FooterBase)`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  width: 100%;
+  background-color: ${(p) => p.theme.colors.lightContrast};
+  ${(p) => boxShadow(p.theme, 1)};
+` as any
```
wilysword, line 192: Might want to take a change at the recent changes I made to the `Grid` component. They were doing something similar, and it led to quite a few type errors in the sample apps.
https://github.com/repaygithub/cactus/pull/303/commits/4eea4cbc6aa6caf02e3feaac030bfb27e1db10a6
https://github.com/repaygithub/cactus/pull/303/commits/7c780ca92115ae382cfb2116f44e9f8895b08f56

modules/cactus-web/src/Footer/Footer.tsx
Dhalton: Did you mean something like `(props: any) => React.ReactElement`? TS doesn't seem to like me rendering the logo component when its type is set to `React.ReactElement<any>`. I keep getting the error: `JSX element type 'Logo' does not have any construct or call signatures`.

wilysword: No, I mean they should pass the instantiated component instead of the type, i.e.
```
<Footer logo={<MyAwesomeLogo aspectRatio="wide" />}>
```
so the type here would be
```suggestion
  logo?: string | React.ReactElement
```

Dhalton: Ah, that's what I was missing. Gotcha.

Dhalton:
Alright, @wilysword @mikesoltow I think I've got everything you guys mentioned taken care of. Let me know what you think!

PR #308, opened 2020-08-26
New color variants

NicolasSimmonds:
Checkout this branch.
build the theme
run cactus-web

If you find more components on which these new color variants can be applied, pleas tell me


Dhalton:
It looks like this has the changes from your disabled styes PR. That's fine, but can we target your `match-disabled-styles` branch on this PR until that other one is merged? It'd make this a lot easier to review

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -149,11 +148,10 @@ const DropdownButton = styled(ReachMenuButton)`
   ${(p): string =>
     p.disabled
       ? `
-  color: ${p.theme.colors.mediumGray};
-  background-color: ${p.theme.colors.lightGray};
   cursor: not-allowed;
   `
       : ''}
+  ${(p) => p.theme.colorStyles.disable};
```
wilysword, line 154: This is wrong, I think; the button has the disabled style even when it's enabled.

NicolasSimmonds: Do you mean the DropdownButton? I'm not seeing any disable style being applied when its enabled

PR #310, opened 2020-08-27
CACTUS-290 Side-MenuBar

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-290

I strongly recommend you look at the diff with whitespace ignored.

- Rather than the finicky scroll buttons, I just went with a builtin scrollbar. Maybe not the prettiest, but it's easy to implement _and_ use. (To be honest, I rather wish we did that for the original MenuBar styles as well...)
- For `mock-ebpp`, I basically hard-coded it to use the `large` screen size, so that the integration tests will still pass.
- ~I've noted that the scrolling behavior is weird when expanding/collapsing submenus on the sidebar (sometimes it'll scroll to the top for no apparent reason); I'll be looking into this.~ EDIT: Should be fixed now.
- I saw that two styles weren't quite right on IE: the focus outline and the overlay on `tiny` screen sizes. I don't see much point in fixing either: the partial focus outline is still sufficient to visually indicate focus, and I doubt anyone would ever actually shrink a desktop browser to the `tiny` size.
- You may also notice that the underlines on expanded submenus don't go all the way to the left, unlike how it's depicted in the style guide. I did this because not only is it easier to implement that way, but I think it gives a better visual clue of the current nesting level than padding alone.

### DIFF
--- a/modules/cactus-web/src/MenuBar/MenuBar.tsx
+++ b/modules/cactus-web/src/MenuBar/MenuBar.tsx
@@ -602,6 +602,9 @@ const HamburgerButton = styled.button.attrs({ role: 'button' })`
       left: 0;
       right: 0;
       cursor: default;
+      border: 0;
+      width: auto;
+      height: auto;
       ${(p) => media(p.theme, 'small')} {
         display: none;
       }
--- a/modules/cactus-web/src/MenuBar/MenuBar.tsx
+++ b/modules/cactus-web/src/MenuBar/MenuBar.tsx
@@ -11,7 +11,7 @@ import styled from 'styled-components'
 
 import { isActionKey, keyPressAsClick } from '../helpers/a11y'
 import { AsProps, GenericComponent } from '../helpers/asProps'
-import { border, boxShadow, media, radius, textStyle } from '../helpers/theme'
+import { border, borderSize, boxShadow, media, radius, textStyle } from '../helpers/theme'
 import { ScreenSizeContext, SIZES } from '../ScreenSizeProvider/ScreenSizeProvider'
 import {
   focusMenu,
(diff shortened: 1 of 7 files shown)


wilysword:
1. I didn't see this, could you give some more information on how to reproduce?
2. As you say, there's really no general way to fix that without the layout provider; unless we want to make it so that the menubar _isn't_ `fixed` when it appears by itself?
3. I see this as the same problem as (2).

mikesoltow:
> 1. I didn't see this, could you give some more information on how to reproduce?

Open up storybook and reduce the screen size to tiny.  Click on the hamburger icon.  The menu bar will become grayed out due to the overlay.

> 2. As you say, there's really no general way to fix that without the layout provider; unless we want to make it so that the menubar _isn't_ `fixed` when it appears by itself?

Yeah, I'm fine leaving this as-is for now.

> 3. I see this as the same problem as (2).

Yeah, that's a good point...as even if you add a `z-index`, you'll just run into the problem from number 2 if the content positioning isn't adjusted.  I'm happy to leave this as well.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -602,6 +602,9 @@ const HamburgerButton = styled.button.attrs({ role: 'button' })`
       left: 0;
       right: 0;
       cursor: default;
+      border: 0;
+      width: auto;
+      height: auto;
```
luisniebla, line 607: Why don't you alphabetize this 😡

mikesoltow: TROLL :trollface:

wilysword: I'm not saying the way I have it now is ideal, but I don't think alphabetical is the best way to arrange styles either. I think it's best to group them by function: positioning, sizing, appearance, etc. Unless there's a whole lot of entries, though, it doesn't seem worth it to me to go through and change what's already there.

Dhalton (approved):
Really well done 🤯

PR #311, opened 2020-08-31
Refactor tooltip

NicolasSimmonds:
Testing:

- Make sure the default tooltip shows op on the top left, and check it adjusts well on collisions
- Take a look at the tooltip icon and make sure it matches the Styleguide

### DIFF
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
@@ -7,7 +7,7 @@ import { FieldWrapper } from '../FieldWrapper/FieldWrapper'
 import useId from '../helpers/useId'
 import Label, { LabelProps } from '../Label/Label'
 import StatusMessage, { Status } from '../StatusMessage/StatusMessage'
-import { Tooltip } from '../Tooltip/Tooltip'
+import { Tooltip, TooltipHandle } from '../Tooltip/Tooltip'
 
 interface AccessibleProps {
   name: string
@@ -98,6 +98,8 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
   } = accessibility
 
   const ref = React.useRef<HTMLDivElement | null>(null)
+  const tooltipRef = React.useRef<TooltipHandle>(null)
+
   const [maxWidth, setMaxWidth] = React.useState<string | undefined>(undefined)
   React.useLayoutEffect((): void => {
     if (ref.current instanceof HTMLElement) {
@@ -108,13 +110,23 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
     }
   }, [maxWidth, setMaxWidth])
 
+  const handleFieldFocus = () => {
+    if (tooltipRef.current !== null) tooltipRef.current.toggle()
+  }
+
   return (
     <FieldWrapper className={props.className} ref={ref}>
       <Label {...props.labelProps} id={labelId} htmlFor={fieldId}>
         {props.label}
       </Label>
       {props.tooltip && (
-        <Tooltip label={props.tooltip} id={tooltipId} maxWidth={maxWidth} disabled={disabled} />
+        <Tooltip
+          label={props.tooltip}
+          id={tooltipId}
+          maxWidth={maxWidth}
+          disabled={disabled}
+          ref={tooltipRef}
+        />
       )}
       {typeof props.children === 'function'
         ? props.children(accessibility)
@@ -124,6 +136,8 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
             'aria-describedby': ariaDescribedBy,
             status,
             disabled,
+            onFocus: handleFieldFocus,
+            onBlur: handleFieldFocus,
           })}
       {status !== undefined && (
         <div>
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
@@ -21,11 +21,7 @@ interface AccessibleProps {
   disabled?: boolean
 }
 
-type RenderFunc = (
-  props: AccessibleProps,
-  handleFieldBlur: () => void,
-  handleFieldFocus: () => void
-) => JSX.Element | JSX.Element[]
+type RenderFunc = (props: AccessibleProps) => JSX.Element | JSX.Element[]
 
 // These are the props commonly used by components that wrap AccessibleField.
 export interface FieldProps {
@@ -122,7 +118,12 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
   }
 
   return (
-    <FieldWrapper className={props.className} ref={ref}>
+    <FieldWrapper
+      className={props.className}
+      ref={ref}
+      onFocus={handleFieldFocus}
+      onBlur={handleFieldBlur}
+    >
       <Label {...props.labelProps} id={labelId} htmlFor={fieldId}>
         {props.label}
       </Label>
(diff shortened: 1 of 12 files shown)


modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -108,13 +110,23 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
     }
   }, [maxWidth, setMaxWidth])
 
+  const handleFieldFocus = () => {
```
Dhalton, line 113: Kind of nit-picky but can we rename this function to `toggleTooltip` or something to that effect? I only bring it up because this function is passed to both `onFocus` and `onBlur`, so the function name is slightly misleading.

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -108,13 +110,23 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
...
+        <Tooltip
+          label={props.tooltip}
+          id={tooltipId}
+          maxWidth={maxWidth}
+          disabled={disabled}
+          ref={tooltipRef}
+        />
       )}
       {typeof props.children === 'function'
         ? props.children(accessibility)
```
mikesoltow, line 132: I agree with Dhalton that all components that use the `AccessibleField` component should have this focus behavior.  You can accomplish this by passing down the focus/blur handlers into the `props.children` function.  You'd then have to actually wire those handlers up in the components that use the children-as-function form of `AccessibleField`.

mikesoltow:
@NicolasSimmonds I also forgot one other thing in my review.  I'd recommend using a prop on the `Tooltip` component, such as `forceVisible`, to control the visibility as opposed to making `AccessibleField` interact with a ref.  The prop could be toggled in the focus/blur handlers.  I think this would be a bit easier to follow and requires less knowledge from components that consume `Tooltip`.  See https://github.com/repaygithub/cactus/commit/726e32182136cc2fdba66e658c2eeaa86e695379 for an example.

modules/cactus-web/src/SelectField/SelectField.test.tsx
```diff
@@ -226,11 +226,12 @@ describe('component: SelectField', (): void => {
       )
       const select = getByLabelText('Requires a label') as HTMLElement
       fireEvent.blur(select)
-      expect(onBlur).toHaveBeenCalledWith('the-test-select-field')
+      expect(onBlur).toHaveBeenCalled
```
Dhalton, line 229: This should be a function call

modules/cactus-web/src/SelectField/SelectField.test.tsx
```diff
@@ -226,11 +226,12 @@ describe('component: SelectField', (): void => {
...
     })
 
     test('onFocus', async (): Promise<void> => {
       const onFocus = jest.fn()
+      console.log(onFocus())
```
Dhalton, line 234: Probably don't need this console log here

Dhalton (commented):
Looks good to me, there are just a couple of small things in the `SelectField` tests but other than that I'm ready to approve

Dhalton (approved):
LGTM, thanks for making those changes!

modules/cactus-web/src/DateInputField/DateInputField.tsx
```diff
@@ -57,6 +61,8 @@ function DateInputFieldBase(props: DateInputFieldProps): React.ReactElement {
           status={status}
           aria-labelledby={labelId}
           aria-describedby={ariaDescribedBy}
+          onFocus={handleFieldFocus}
+          onBlur={handleFieldBlur}
```
mikesoltow, line 65: Why don't you also handle `onFocus`/`onBlur` events passed into `DateInputField` (like you do with the other components)?

mikesoltow (commented):
Changes look great!  Just one question regarding the `DateInputField`, then I'm good with this.

modules/cactus-web/src/DateInputField/DateInputField.tsx
NicolasSimmonds: Ups, I guess I missed that

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -108,22 +114,37 @@ function AccessibleFieldBase(props: AccessibleFieldProps): React.ReactElement {
...
-        ? props.children(accessibility)
+        ? props.children(accessibility, handleFieldBlur, handleFieldFocus)
         : React.cloneElement(React.Children.only(props.children), {
             id: fieldId,
             name,
             'aria-describedby': ariaDescribedBy,
             status,
             disabled,
+            onFocus: handleFieldFocus,
+            onBlur: handleFieldBlur,
```
wilysword, line 147: Setting these will override any handlers the dev has already assigned to the child element, leaving their only option to use a render func and then frankenstein the two handlers together. I would suggest putting the `RenderFunc` type back how it was, removing these from the cloneElement call, and instead setting them on the `FieldWrapper`. The focus/blur events will bubble up from the child element so you can still set/unset the tooltip visibility, but without compromising the dev's control over the child.

```
<FieldWrapper className={props.className} ref={ref} onFocus={handleFieldFocus} onBlur={handleFieldBlur}>
```

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -107,22 +118,31 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
...
   pointer-events: none;
   position: absolute;
   padding: 16px;
-  border-radius: 8px 8px 8px 8px;
-  ${(p): string => boxShadow(p.theme, 2)};
+  ${(p): string => boxShadow(p.theme, 1)};
   font-size: 15px;
   ${(p): ColorStyle => p.theme.colorStyles.standard};
-  border: 2px solid ${(p): string => p.theme.colors.callToAction};
+  border-color: ${(p): string => p.theme.colors.callToAction};
```
wilysword, line 135: This is redundant with the `border` declaration on line 139.

wilysword (changes requested):
I like the changes to the tooltip, but I think the changes to `AccessibleField` are a little too invasive. Should be easy enough to fix, though.

wilysword (approved):

mikesoltow (commented):
Changes look good!  Just a couple of questions about the remaining event handlers.

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -46,54 +50,75 @@ const cactusPosition: Position = (triggerRect, tooltipRect) => {
...
+  } else if (props.forceVisible) {
+    return css`
+      color: ${(p): string => p.theme.colors.callToAction};
+    `
+  }
+  return css`
+    color: ${(p): string => p.theme.colors.darkestContrast};
+  `
+}
 const StyledInfo = styled(NotificationInfo)<StyledInfoProps>`
```
mikesoltow, line 97: One more thing.  I'm seeing console errors related to the fact that the `forceVisible` prop is being passed all the way down to the icon SVG, and React isn't recognizing it as a valid HTML prop.  We'll need to filter it out to get rid of that error.

```suggestion
styled(({ forceVisible, ...props }) => <NotificationInfo {...props} />)<
  StyledInfoProps
>`
```

mikesoltow (commented):
Thanks for the changes!  Unfortunately I found one more thing right before I was going to approve 😢

modules/cactus-web/src/Tooltip/Tooltip.tsx
wilysword: Alternative ways of doing the same thing: https://styled-components.com/docs/api#transient-props

NicolasSimmonds: I liked Glen's idea. Updated

mikesoltow: One downside to transient props, however, is that it's only available in Styled Components v5.  That would require a breaking Cactus change, as currently we support apps that use Styled Components v4.  Therefore my vote is to do it the non-transient way, as I don't think this refactor is worth a breaking change.

mikesoltow (approved):
Thanks for the changes!  Looks good to me!

PR #312, opened 2020-08-31
CACTUS-198 ::  Mobile Combobox Accessibility

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-198

So I finally figured out why VoiceOver wasn't able to focus on the input...for some reason it didn't like it being set to `role="search"`. When I changed it to `role="textbox"`, VoiceOver was able to focus on & type in the input as expected.

I added some docs outlining how you can connect an iPhone to a Mac & test Storybooks locally, but obviously that only works if you're using both an iOS device and a MacOS device. This leaves me with a couple of questions:

1. Should I add some steps to connect other devices to Mac, such as Android? Could be useful, as I have no idea what phones other members of the team have. Although I have no way to test it, so I wouldn't know if it actually works.
2. Should we include any docs for non-Mac users, like if someone needed to connect a mobile device (either iPhone or Android) to a machine on a different OS like Windows or Linux? I think this would be slightly less useful since most of us use Mac, and again I have no way to test it.

We could probably at least provide some links to some guides out there for the aforementioned devices, if you guys are on board with that.

Another thought I just had is that we should probably put an `aria-label` on there. Would you guys be ok with me adding it to the select props and making it default to "Search for an option" or something to that effect?

## Testing
Hopefully there are at least two of you that use iPhones, since we should be testing that VoiceOver on mobile works correctly...
1. Follow the steps I listed to connect your phone to your computer & bring up the storybooks. Open the "With Multiselect Combobox" story under the Select.
2. Turn on VoiceOver. I'd suggest enabling the shortcut by going to Settings > Accessibility > Accessibility Shortcut and selecting VoiceOver. Once that's enabled, you can triple-tap the side button to toggle VoiceOver.
3. Click on the select and make sure that you're able to filter the options in the list using the input at the bottom with VoiceOver turned on.

### DIFF
--- a/examples/mock-ebpp/tests/helpers/actions.ts
+++ b/examples/mock-ebpp/tests/helpers/actions.ts
@@ -9,7 +9,7 @@ const getDropdown = Selector((): Element | null => {
 })
 
 const getCombo = Selector((): Element | null => {
-  if (document.activeElement && document.activeElement.getAttribute('role') === 'search') {
+  if (document.activeElement && document.activeElement.getAttribute('role') === 'textbox') {
     return document.activeElement
   }
   return null
--- a/modules/cactus-web/README.md
+++ b/modules/cactus-web/README.md
@@ -152,3 +152,18 @@ test('should not trigger onChange event', () => {
   expect(onChange).not.toHaveBeenCalled()
 })
 ```
+
+### Accessing Local Storybooks on iOS
+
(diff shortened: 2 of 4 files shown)


mikesoltow (approved):
Changes LGTM.  Re-approve!

wilysword (approved):

PR #313, opened 2020-09-02
feat(cactus-web): tag Component

NicolasSimmonds:
This Component is not ready Yet. Just wanted to create a PR to see what you guys think, the  Tag component is working as expected in the Select component, however there is one thing to consider regarding the Tag component by its own.

- Since the props interface must be the same, what happens when the close icon is clicked? At this moment the responsable of doing this action is the Select Component, and so when you click this icon nothing happens. So It's gonna need some other logic that probably will mess with all the implementation of the Select component.

[CACTUS-339]

[CACTUS-339]: https://repayonline.atlassian.net/browse/CACTUS-339


mikesoltow:
@NicolasSimmonds My main comment is that I was only thinking of extracting the `ValueTag` component, as stated in the ticket.  I don't think we need all the logic of the `ValueSwitch`, etc., as that's specific to the `Select` component and its restrictions.

As for the close handler, I think we do need to add an `onClose` property or something (sorry, I thought there already was one for some reason).  We don't know what will be powering these components in terms of data, so we need to leave the management up to the parent, much like a controlled form input.  You could still have the story be interactive, however, by wrapping `Tag` in a container that manages some basic state (like an array of tag names or something).

PR #315, opened 2020-09-03
Tag component

NicolasSimmonds:
I added  MarginProps to the Tag interface props, felt like in some cases a dev would like to have them a little bit more separated than they are by default. Let me know if you are good with this.


modules/cactus-web/src/Tag/Tag.tsx
```diff
@@ -0,0 +1,79 @@
...
+  hidden?: boolean
+  onCloseIconClick?: () => void
+}
+
+const TagBase = React.forwardRef<HTMLSpanElement, TagProps>(
+  ({ id, className, closeOption, children, onCloseIconClick }, ref): React.ReactElement => {
+    return (
+      <span id={id} ref={ref} className={className}>
+        <span className="value-tag__label">{children}</span>
+        {closeOption && <NavigationClose data-role="close" onClick={onCloseIconClick} />}
```
wilysword, line 23: I feel like if `onCloseIconClick` is passed this should be displayed even if `closeOption` isn't present.

wilysword (approved):
Looks good; I have a couple of thoughts, but nothing important enough to withhold approval.

In the storybook, it might be useful to have a way to add tags back without having to leave the page and come back, but I wouldn't say it' important enough to spend a lot of time on it.

modules/cactus-web/src/Tag/Tag.tsx
NicolasSimmonds: I like this idea.

wilysword: That's what I would say.

Dhalton (approved):
Nice work, looks good to me

PR #319, opened 2020-09-04
CACTUS-338 :: Formik Example

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-338

Like I mentioned in our meeting, I wanted to configure the form validation to use the `error` prop on our form elements. As I did this I found a couple of components that don't expose that prop:
  - `RadioButtonField` - I think it would probably make the most sense to include `error`/`warning`/`success` props on the _group_ that Glen has been working on rather than individual radio button fields.
  - `FileInputField` - This component provides ways for developers to render error messages when a file fails to upload, but not for things like when a user doesn't upload a file at all. I think we should add that feature and just display a status message below the box.

Let me know what you guys think about those suggestions.

### Testing
This should really only affect the `mock-ebpp` example app, so I'd just run `yarn w mock-ebpp start` and play around with the UI Config form; make sure you can fill it out & submit, make sure errors are showing up, etc.

### DIFF
--- a/docs/Components/API Documentation.md
+++ b/docs/Components/API Documentation.md
@@ -39,6 +39,15 @@ Inverse props are a _work in progress_ and should not be used in production curr
 
 Most of the time, a theme will make use of a lighter base color, and components placed on that background will be easily visible, however, if the base color is dark, it will be necessary to use the inverse colors for a component so that it is easy to see. The inverse components flip the colors used in their standard counterpart which will provide more contrast, making them easier to see.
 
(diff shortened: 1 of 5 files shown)


wilysword (approved):

wilysword:
I was indeed planning on adding status messages to the RadioGroup component (or rather, I already did so. Adding something similar to the file input seems reasonable.

NicolasSimmonds (approved):

PR #324, opened 2020-09-09
CACTUS-348 :: DataGrid Column Width

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-348

To do this, I basically just took any extra props passed to `Column` or `DataColumn` and passed those to `Table.Cell`. Then, in `Table.Cell` I adjusted the CSS a bit so that we only set a `min-width` when `width` isn't passed. If it is passed, we just call the `width` function from Styled System.

The way I did this essentially gives developers the freedom to pass any extra props they want onto the table cells via the column props. Are you guys ok with that? It seemed like the right call to me.

### Testing
1. Open the DataGrid Basic Usage story.
2. Set width values using the two new knobs: "Active Column Width" and "Action Column Width".
3. Make sure that the values you set are setting the column widths appropriately.

### DIFF
--- a/modules/cactus-web/src/Table/Table.tsx
+++ b/modules/cactus-web/src/Table/Table.tsx
@@ -40,7 +40,7 @@ interface TableHeaderProps
   variant?: TableVariant
 }
 
-interface TableCellProps
+export interface TableCellProps
   extends React.DetailedHTMLProps<
     React.TdHTMLAttributes<HTMLTableDataCellElement> &
       React.ThHTMLAttributes<HTMLTableHeaderCellElement>,
--- a/modules/cactus-web/src/Table/Table.tsx
+++ b/modules/cactus-web/src/Table/Table.tsx
@@ -4,7 +4,7 @@ import React, { createContext, useContext } from 'react'
 import styled, { css, FlattenSimpleInterpolation } from 'styled-components'
 import { width, WidthProps } from 'styled-system'
 
-import { border, boxShadow, textStyle } from '../helpers/theme'
+import { border, boxShadow, media, textStyle } from '../helpers/theme'
 import variant from '../helpers/variant'
 import { ScreenSizeContext, Size, SIZES } from '../ScreenSizeProvider/ScreenSizeProvider'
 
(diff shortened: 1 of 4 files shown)


mikesoltow (commented):
This is looking good from what I can see so far.  However, I did notice one quirk: because styled components will pass down any valid HTML attributes to the styled element, the `width` prop is getting passed down to the `td` directly rather than going through Styled System.  This is probably not ideal, as the `width` attribute on `td` elements is deprecated, so we should be allowing Styled System to handle this.

I tried filtering it out manually by replacing StyledCell with:

```javascript
const StyledCell = styled(({ width, children, ...props }) => <td {...props}>{children}</td>)`
  ${variant({
  // ...etc.
```

But nothing was returning from the `${width}` function for some reason.  I'd have to play around with it more to see what is going on, which I'm happy to do later if you need me to.

Dhalton:
@mikesoltow I wonder if it has anything to do with `TableCellProps` not extending styled system's `WidthProps`? I'll do some digging. Thanks for letting me know

mikesoltow (approved):
Looks good to me!  I still don't see `width` actually show up in the styles when using DevTools, but it seems to be being applied anyway.  I couldn't get Styled System's responsive styles to work (i.e. width={[10, 100, 500, 1000]}`, but I'm not sure if we care...

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -231,24 +236,23 @@ const ContentBox = styled.div`
...
-
-      ${(p): string | undefined =>
-        p.theme.mediaQueries &&
-        `${p.theme.mediaQueries.extraLarge} {
-      min-width: 160px;
-    }`}
+
+      ${(p) =>
+        p.width
+          ? `${width}`
```
wilysword, line 247: This just converts the function to a string, which is why it's not working correctly. Try this on for size: https://github.com/repaygithub/cactus/commit/6c1e38ded1d9986075de3fbdff1bc32c7c0b210d.

Dhalton: I like that, it seems like it does the trick. But the width attribute is still being set on the `td` element in the DOM. Shouldn't we still be filtering it out since it's deprecated?

wilysword: I don't see any particular reason to. The CSS overrides whatever's set in the attribute, and it's not like we're setting it on purpose. If styled-components ever decides to remove it, it'll be removed, and it's not breaking anything in the meanwhile.

wilysword (approved):

mikesoltow (approved):
LGTM!

PR #325, opened 2020-09-10
CACTUS-352 :: DataGrid Scroll Bug

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-352

Pretty quick one here. The bug came up because the Channels team uses a CSS framework that applies `border-box` styles to all of their elements. For this reason, ~I left the `* { box-sizing: border-box }` style in the storybook CSS~ I added a field to the theme addon in storybook so that we can toggle the border-box styles easily at any time. That way, we can hopefully try to prevent similar issues from cropping up later on.

### Testing
1. Bring up the DataGrid storybook and open the "Cactus Theme" addon tab. At the bottom, check the box to set all elements to `box-sizing: border-box`
2. Make sure there's no vertical scrollbar on the table itself. The pagination component should not be cut off.
2. If you want to test that the fix is _really_ working, you can remove the `box-sizing: content-box` style from the `li`s in the pagination component and you should see the problem pop back up.

### DIFF
--- a/modules/cactus-web/cactus-addon/constants.js
+++ b/modules/cactus-web/cactus-addon/constants.js
@@ -3,3 +3,4 @@ export const PROP_NAME = 'cactus'
 export const THEME_CHANGE = `${NAME}/themeChange`
 export const BACKGROUND_CHANGE = `${NAME}/backgroundChange`
 export const DECORATOR_LISTENING = `${NAME}/decoratorListening`
+export const BORDER_BOX_CHANGE = `${NAME}/borderBoxChange`
--- a/modules/cactus-web/cactus-addon/index.tsx
+++ b/modules/cactus-web/cactus-addon/index.tsx
@@ -4,7 +4,14 @@ import * as React from 'react'
 import styled, { CSSObject } from 'styled-components'
 
 import { StyleProvider } from '../src/StyleProvider/StyleProvider'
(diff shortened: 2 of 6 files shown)


wilysword (commented):
I agree with the fix, but I'm not so sure about the change to the default storybook styles. That could cause some unintended changes (so far I've seen differences on DateInput, and FileInput when there's an error). Of course, given your description, I guess those differences would also show up in EBPP, but why on earth are they using something that changes the default styles so fundamentally? If we can't rely on browser defaults, how do we know what properties we need to explicitly set?

Dhalton:
> why on earth are they using something that changes the default styles so fundamentally? If we can't rely on browser defaults, how do we know what properties we need to explicitly set?

Your guess is as good as mine. I can remove the border-box default styles, but what if we created a storybook helper to switch between default styles and border-box everything? It's something we should account for given that we know Channels is doing that now. It'd be nice to have the ability to quickly toggle between the two so we can make sure our components will work for both scenarios.

wilysword:
If you can work out how to do that, I'd be okay with it (though I'd prefer if they just stopped using whatever changes all the styles).

wilysword (approved):

mikesoltow (approved):
Looks great!  I like the Storybook option.  Let's just hope I remember to use it in the future 😄

PR #326, opened 2020-09-11
Fix select focus

NicolasSimmonds:
There is no easy way to test this since the select is pretty much working fine in the mock-ebpp, it faild for me just once or so.
I worked this solution with John who was the reporter and told me it worked just fine for him.


wilysword (approved):

mikesoltow (approved):
Looks great!

PR #329, opened 2020-09-14
CACTUS-355 :: Tooltip Fix

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-355

This was happening because `reach/tooltip` doesn't get the client rect for the trigger unless the mouse hovers over the tooltip icon. This wouldn't be a problem if the icon were always in the same place, but when the page is big enough to scroll, then the icon moves. My solution was to manage the ref on the trigger myself and call `getBoundingClientRect()` to pass that into our `cactusPosition` function.

### Testing
1. Clean & rebuild `cactus-web` with `yarn web cleanup && yarn web build`
2. Run the mock-ebpp example app with `yarn w mock-ebpp start`
3. Open the UI Config form and make sure the tooltip comes up in the right spot when you trigger it by focusing on fields, even after you scroll on the page.

### DIFF
--- a/modules/cactus-web/src/TextInputField/TextInputField.story.tsx
+++ b/modules/cactus-web/src/TextInputField/TextInputField.story.tsx
@@ -39,7 +39,8 @@ textInputFieldStories
         name="input-1"
         {...eventLoggers}
       />
-    )
+    ),
+    { cactus: { overrides: { height: '110vh', width: '110vw' } } }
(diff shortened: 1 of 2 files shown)


mikesoltow (approved):
LGTM!

NicolasSimmonds (approved):
Awesome!

PR #332, opened 2020-09-15
CACTUS-345 :: Dropdown Widths

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-345

I bumped up the percentage the `MenuButton` uses so that the dropdown could be at most 2x as large as the `MenuButton`, and I used that value for the `SplitButton` & `Select` as well.

Should we consider setting the maximum width the `Select`'s dropdown can grow past its trigger to something larger than 300px? I think 300px is probably fine for the other two, but for the `Select`, I think it'd be much more common to see them with width greater than 150px (the way it's written, if 150px < Select width < 300px, then the dropdown will grow past the select, but only up to 300px max). Maybe we could set the hard cap somewhere in the 400 - 500 pixel range for the `Select`?

### Testing
1. Open the `MenuButton` storybook
2. Set the button's text to something small, 1-3 characters or so. Make sure the dropdown is growing past the button.
3. Open the `SplitButton` storybook
4. Set the main action text to something small, 1-3 characters or so. Make sure the dropdown is growing past the component's width.
5. Open the `Select` "Long option labels" storybook
6. Open the dropdown and make sure its width is longer than that of the Select

### DIFF
--- a/modules/cactus-web/src/MenuButton/MenuButton.tsx
+++ b/modules/cactus-web/src/MenuButton/MenuButton.tsx
@@ -140,7 +140,7 @@ function MenuButtonBase(props: MenuButtonProps): React.ReactElement {
 
           return {
             minWidth: targetRect.width,
-            maxWidth: Math.max(targetRect.width, Math.min(targetRect.width * 2, 300)),
+            maxWidth: Math.max(targetRect.width, Math.min(targetRect.width * 2, 400)),
             left: targetRect.left + scrollX,
             ...getTopPosition(targetRect, popoverRect),
           }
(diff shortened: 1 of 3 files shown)


mikesoltow (approved):
LGTM.  I'm open to increasing the max width from 300, but why limit it to just Select?

Dhalton:
> LGTM. I'm open to increasing the max width from 300, but why limit it to just Select?

I just would think it more common for a Select to need the extra width than a SplitButton/MenuButton to need it. I think Selects are typically longer and probably have longer option labels. I could be wrong though, maybe we could bump it up to 400px for all of them and see if that works for Channels?

mikesoltow:
> > LGTM. I'm open to increasing the max width from 300, but why limit it to just Select?
> 
> I just would think it more common for a Select to need the extra width than a SplitButton/MenuButton to need it. I think Selects are typically longer and probably have longer option labels. I could be wrong though, maybe we could bump it up to 400px for all of them and see if that works for Channels?

Works for me if you all are okay with that

NicolasSimmonds (approved):

wilysword (approved):

PR #334, opened 2020-09-17
CACTUS-364 :: Improve SplitButton Performance

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-364

The main culprit actually turned out to be Styled Components.  Channels was using v4, while we were using v5, and it turns out that they had significant performance improvements between the two major versions.  However, even after upgrading Styled Components to v5, the Channels app was still seeing some jankiness when you scrolled down a large table of SplitButtons.  To fix that, we still need to apply the updates to the Reach packages.

Therefore, this PR does two things:
* Upgrade all the Reach packages
* Alter the Styled Components peer dependency so that only v5.x.x is acceptable (this is a breaking change, unfortunately).

### Testing

You'll want to use the new Data Grid "Lots and Lots of Rows" story.  However, I'd recommend viewing that outside of the standard Storybook interface so that you get minimal interference from Storybook.  You can do that by navigating to http://localhost:9001/iframe.html?id=datagrid--lots-and-lots-of-rows after starting Storybook.

1. Start with the current `master` branch, but add in the code from the new "Lots and Lots of Rows" story so that you can use it.
2. Downgrade StyledComponents to v4.4.1 in `@repay/cactus-web` (and maybe in other packages too, to make sure yarn doesn't get confused and use the yarn version).
3. Navigate to the new story linked above.  Notice that it takes ~30s to render.
4. Revert the StyledComponents downgrade so that you're back to v5, but stay on `master`.  Reload the story and notice that the render time is down to ~10s.  However, note that when you scroll down the DataGrid, the screen goes white and everything locks up for a few seconds before rows are rendered.
5. Now you can finally check out this branch.  Reload the story.  Note that the render time is down to ~6s.  Also notice that you can scroll normally.
6. Click around all other components that use Reach to make sure we didn't break anything.  These include:

* Accordion
* DateInput/DateInputField
* Footer (via the rect helper)
* MenuButton
* Modal
* Select/SelectField
* SplitButton
* Tooltip

**Note:** The multi-select combo box was previously broken in IE...so don't blame this PR!  Ticket coming shortly.


Dhalton (approved):
Awesome! Big improvement. We're still going to be disabling "portal mode" for reach dropdowns, right? That's just going to be a separate ticket I assume

mikesoltow:
> Awesome! Big improvement. We're still going to be disabling "portal mode" for reach dropdowns, right? That's just going to be a separate ticket I assume

Actually, I'm not sure.  I previously thought it offered an improvement, but lately I haven't been able to notice a difference vs. what I have here.

Would you mind taking a look when you have a sec?  (No rush!)  To test, just go into `SpitButton.tsx` and pass in `portal={false}` to `ReachMenuPopover`.  It will mess up the styles, of course, but you can ignore that for now.  I'm curious if you notice any difference, either in the initial load or when scrolling.

Dhalton:
> > Awesome! Big improvement. We're still going to be disabling "portal mode" for reach dropdowns, right? That's just going to be a separate ticket I assume
> 
> Actually, I'm not sure. I previously thought it offered an improvement, but lately I haven't been able to notice a difference vs. what I have here.
> 
> Would you mind taking a look when you have a sec? (No rush!) To test, just go into `SpitButton.tsx` and pass in `portal={false}` to `ReachMenuPopover`. It will mess up the styles, of course, but you can ignore that for now. I'm curious if you notice any difference, either in the initial load or when scrolling.

I didn't see a huge difference with the amount that you have rendered in the storybook in this PR; it seemed like I was shaving off maybe a quarter of a second or so. When I increase the number of rows (I have it at ~550) right now, it seems like there's a more apparent improvement in the load time. I'm seeing ~7.3s to load as it stands now and ~6.3s with portals turned off, so I'd say it's definitely a step further in the right direction, but it's not a massive difference. The scrolling seems fine to me both with portals on and off. I haven't noticed a change there.

mikesoltow:
> I didn't see a huge difference with the amount that you have rendered in the storybook in this PR; it seemed like I was shaving off maybe a quarter of a second or so. When I increase the number of rows (I have it at ~550) right now, it seems like there's a more apparent improvement in the load time. I'm seeing ~7.3s to load as it stands now and ~6.3s with portals turned off, so I'd say it's definitely a step further in the right direction, but it's not a massive difference. The scrolling seems fine to me both with portals on and off. I haven't noticed a change there.

Nice!  I think we could definitely look at going that direction, but it should be a separate ticket due to the fact that we'll have to mess with styles (and maybe accessibility?).

NicolasSimmonds (approved):

wilysword (approved):

PR #335, opened 2020-09-17
Brandbar component

NicolasSimmonds:
For now, the Menu only appears on small screens or larger. On tiny screens it disappears.

The Component was also incorporated on the layout component replacing the temp brandbar


modules/cactus-web/src/BrandBar/BrandBar.mdx
```diff
@@ -0,0 +1,54 @@
...
+    <BrandBar
+      usernameText="Hershell Jewess"
+      logo="https://repay-merchant-resources.s3.amazonaws.com/staging/24bd1970-a677-4ca7-a4d2-e328ddd4691b/repay_logo_new.jpg"
+    >
+      <BrandBar.Item onSelect={someAction}>Settings</BrandBar.Item>
+      <BrandBar.Item onSelect={someAction}>Logout</BrandBar.Item>
+    </BrandBar>
+  )
+}
+```
```
wilysword, line 50: Seems like you may not need this section at all, since all the information it contains is in the "Try it out" section.

modules/cactus-web/src/BrandBar/BrandBar.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      )
+
+      userEvent.click(getByText('Test name'))
+      await animationRender()
+      userEvent.click(getByText('Settings'))
+      expect(actionOne).toHaveBeenCalled()
+    })
+  })
+
+  describe('mouse usage', (): void => {
```
wilysword, line 42: Should this be "keyboard usage"?

modules/cactus-web/src/index.ts
```diff
@@ -54,3 +54,4 @@ export { default as useId } from './helpers/useId'
 export { ConfirmModal } from './ConfirmModal/ConfirmModal'
 export { Tag } from './Tag/Tag'
 export { Footer } from './Footer/Footer'
+export { BrandBar } from './BrandBar/BrandBar'
```
wilysword, line 57: For a lot of our components this doesn't matter, but I think it'd be better if we got in the habit of writing our exports like this.
```suggestion
export { default as BrandBar } from './BrandBar/BrandBar'
```

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -0,0 +1,192 @@
...
+    justify-content: space-between;
+  }
+`
+const LogoWrapper = styled.div`
+  padding: 16px;
+`
+const Img = styled('img')`
+  max-width: 200px;
+  max-height: 80px;
+`
```
wilysword, line 117: I think this would be a little more general if they want to use a custom component for the logo.
```suggestion
const LogoWrapper = styled.div`
  padding: 16px;
  max-width: 200px;
  max-height: 80px;
  & > * {
    max-width: 100%;
    max-height: 100%;
  }
`
```

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -0,0 +1,192 @@
...
+            }
+          }}
+        >
+          <MenuList>{children}</MenuList>
+        </MenuPopover>
+      </Menu>
+    </div>
+  )
+}
+BrandBarBase.Item = MenuItem
```
wilysword, line 89: Although it's more verbose I'd like to call this something more descriptive, like maybe `BrandBar.UserMenuItem`. Unlike `Menu.Item` or `ActionBar.Item`, when I read `BrandBar.Item` there's no intuitive understanding of what that actually means.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -0,0 +1,192 @@
...
+import React from 'react'
+import styled, { createGlobalStyle, css } from 'styled-components'
+
+import { getTopPosition } from '../helpers/positionPopover'
+import { getScrollX } from '../helpers/scrollOffset'
+import { border, boxShadow, fontSize, media, textStyle } from '../helpers/theme'
+
+interface BrandBarProps extends React.HTMLAttributes<HTMLDivElement> {
+  logo?: string | React.ReactElement
+  usernameText: string
```
wilysword, line 21: Similar to my comment on `BrandBar.Item`, I think this would be better as `userMenuText` or `userMenuLabel`.

wilysword (commented):
Functionality-wise it looks good. There are a couple of minor styling issues, and I think we might be better off renaming the variables related to the user menu.

The way the drop-down handles positioning seems a little odd, at least when there's no username text. I went to the layout story, deleted the username text, and edited one of the menu items to have a longer label, and it kept the drop-down very narrow and wrapped the text. As in other recent PRs, I think the drop-down should be able to be wider than the button.

When focused/expanded the button border appears over the brand bar border, instead of overlapping it like the menubar does it. You might consider using an inset `box-shadow` instead of a border, similar to how MenuBar does it. You can also look at [how I changed the sidebar/actionbar styles](https://github.com/repaygithub/cactus/compare/cactus-333-actionbar?w=1#diff-53fa83d495d7ff0d621485f441470694R125) (look for all the references to the `insetBorder` helper function).

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -0,0 +1,192 @@
...
+    return css`
+      border: ${border(theme, 'lightContrast')};
+      ${getDropShape};
+    `
+  } else {
+    return css(getDropShape)
+  }
+}
+
+const BrandBarBase = (props: BrandBarProps): React.ReactElement => {
```
wilysword, line 51: This is more a request for discussion than a comment on this PR specifically (right before I leave and can't take part in the discussion).

We do this pattern a lot in cactus: make a "Base" component, then export a styled version of it. You may have noticed I don't really do that in my PRs: instead I style an intrinsic HTML element, and then export a component that uses that styled component as its root. I do this mostly to restrict access to the "extras" that styled-components gives, like the `as` prop (which would break half our components if anyone used it).

I don't believe this should be a universal rule: lower-level components like `Box`, `Link`, `Text`, etc. make sense to export the styled-component directly. For other components however, that have complex structures or hooks that would break when using the `as` prop, I think we should come to a consensus on which pattern we prefer and use it going forward (while possibly refactoring older components little by little).

Current paradigm (custom base component + export styled component):
- Pros: mostly easier to deal with Typescript, can use components in other styles (`${MyComponent} { ...more styles }`)
- Cons: have to remember to forward `className`, `as` prop can completely break components (less control over input in general)

My preferred paradigm (styled intrinsic + export custom component):
- Pros: we have much greater control over inputs, much more difficult for devs to break our components, (in my opinion) easier to read & reason about code
- Cons: Typescript is a little trickier for elements that have sub-elements (e.g. MenuBar.Item; I recently found a workaround for this, though), can't directly use components in styles (there's an easy workaround for this, too)

modules/cactus-web/src/BrandBar/BrandBar.mdx
NicolasSimmonds: It's just there so the documentation has all the sections as the other ones. But since it's the same I think we can remove it. What do you think @mikesoltow @Dhalton

Dhalton: Yeah, I'm fine with removing this. I've been removing these examples when convenient in favor of the editable code. Doesn't really make sense to have both.

modules/cactus-web/src/BrandBar/BrandBar.tsx
Dhalton: I think you make a good point here with the `as` prop leaving the door open for devs to break our components. I agree that we should come to an agreement on which pattern we want to use in most cases. I'm on board with the pattern you've been using, especially if you've figured out solutions for TypeScript & sub-components. I think the only thing I'd really miss would be the ability to directly reference our other components in the styles with `${MyComponent} {...}`, but there are other ways to accomplish that same thing. So I'm cool with this pattern and I'll start doing that moving forward if everyone else is on board too.

wilysword (commented):
Looks much better. I still have a concern about using an outset border instead of an inset border, though: if the logo is shorter than the user menu button, then the top of the outset border (when the button is focused) is cut off. The reason I used inset borders is because it's easier to size them.

modules/cactus-web/src/BrandBar/BrandBar.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+# BrandBar
+
+The `BrandBar` Components allows yout to render a Logo of your choice and a user menu with a dropdown for actions such as settings and Loggin in/out.
+
+### Try it out
+
+export const code = `<BrandBar
+    onProfilePage
+    usernameText'Hershell Jewess'
```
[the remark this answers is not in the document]
NicolasSimmonds: Yeah I tried using your inset border function and even create my own inset Boxshadow, but for some reason it always appeared to be on top of the main border, it wouldn't overlap.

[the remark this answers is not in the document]
wilysword: The main border also has to be changed to an inset border.

wilysword (commented):
Everything looks good, as far as I can tell. Unfortunately, good ol' IE11 can't allow happiness to exist in the world, so it's displaying things incorrectly. For some reason it's stretching the menu button to take up as much space as it can. I have no idea how to fix it: looking at the `Flex` component in isolation, `justify-content: space-between` _should_ work, but for some reason it's not.

NicolasSimmonds:
> Everything looks good, as far as I can tell. Unfortunately, good ol' IE11 can't allow happiness to exist in the world, so it's displaying things incorrectly. For some reason it's stretching the menu button to take up as much space as it can. I have no idea how to fix it: looking at the `Flex` component in isolation, `justify-content: space-between` _should_ work, but for some reason it's not.

Ough I thought everything was working alright in IE11, I'll take a look to see if I can find a solution

wilysword (approved):
No idea why, but that does seem to have fixed the problem.

Dhalton (approved):
Lookin' good!

PR #339, opened 2020-09-22
Upgrade node-forge

mikesoltow:
This just affects webpack-dev-server, so the only thing you should need to do to test is to make sure `yarn w mock-ebpp start` still works fine.


Dhalton (approved):
LGTM, webpack dev server still working fine

wilysword (approved):

PR #340, opened 2020-09-22
Inherit props documentation

NicolasSimmonds:
Build the docs page.

Go to ConfirmModal and make sure both Modal and ConfirmModal props exist. Also ensure no unwanted  props from node_modules are listed in other components


wilysword (changes requested):
Looks perfect, but you should look through the docs pages and remove any references to `includeProps`. I added it very recently, so there should only be one or two pages using it.

Dhalton (commented):
LGTM, I'm ready to approve after you get rid of those `includeProps` references

wilysword (approved):

Dhalton (approved):

PR #341, opened 2020-09-22
CACTUS-215 :: CheckBoxGroup

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-215

I tried to match what Glen did on the `RadioGroup` pretty closely, since this component is very similar. There is a key difference, though, which is that, unlike radio buttons, multiple checkboxes in a group can be checked at one time. There are a couple of changes to the API in comparison to the `RadioGroup` that I want to mention here:
- The `name` prop is not forwarded from the `CheckBoxGroup` to the individual checkboxes because they don't need to have the same name. The developer should provide a name on each checkbox.
- When the `onChange` handler is called, this component passes the name of the checkbox that was changed, along with a boolean value indicating whether or not its checked, instead of simply passing a value. This differs from the `onFocus` and `onBlur` handlers which receive the name on the fieldset instead of a checkbox name.
- Any `value`/`defaultValue` passed to `CheckBoxGroup` will have to be an object, mapping checkbox names to their values.

### Testing
Bring up each of the storybooks and play around with them. Pay attention to the console in the "Basic Usage" story to make sure the values being logged are as expected.


modules/cactus-web/src/index.ts
```diff
@@ -54,3 +54,4 @@ export { default as useId } from './helpers/useId'
 export { ConfirmModal } from './ConfirmModal/ConfirmModal'
 export { Tag } from './Tag/Tag'
 export { Footer } from './Footer/Footer'
+export { CheckBoxGroup } from './CheckBoxGroup/CheckBoxGroup'
```
wilysword, line 57:
```suggestion
export { default as CheckBoxGroup } from './CheckBoxGroup/CheckBoxGroup'
```

modules/cactus-web/src/CheckBoxGroup/CheckBoxGroup.tsx
```diff
@@ -0,0 +1,225 @@
...
+interface CheckBoxGroupProps
+  extends MarginProps,
+    WidthProps,
+    Omit<FieldProps, 'labelProps'>,
+    Omit<
+      React.FieldsetHTMLAttributes<HTMLFieldSetElement>,
+      'name' | 'onChange' | 'onFocus' | 'onBlur' | 'defaultValue'
+    > {
+  value?: { [K: string]: boolean }
+  defaultValue?: { [K: string]: boolean }
```
wilysword, line 27: To be honest, I don't think this adds any value to the API (pun not intended). With RadioGroup it's a singular value so it makes things simpler, but with these complex objects it doesn't really save much. I think the interface would be cleaner without them, but I'm not super-opposed so I'd like some other opinions.

modules/cactus-web/src/CheckBoxGroup/CheckBoxGroup.tsx
```diff
@@ -0,0 +1,225 @@
...
+}
+
+type CheckBoxGroupType = typeof CheckBoxGroup & { Item: typeof CheckBoxGroupItem }
+
+const DefaultCheckBoxGroup = CheckBoxGroup as any
+DefaultCheckBoxGroup.Item = CheckBoxGroupItem
+
+export default DefaultCheckBoxGroup as CheckBoxGroupType
+
+const FieldSet = styled(FieldWrapper)<{ disabled?: boolean }>`
```
wilysword, line 194: This looks identical to what's in `RadioGroup`, maybe just import it from there? (Or put it in a separate file and import _that_ in both places.)

wilysword (commented):
Looks good, just one discussion point before I approve.

modules/cactus-web/src/CheckBoxGroup/CheckBoxGroup.tsx
Dhalton: Yeah, it is identical so that's probably a better idea. Will do

modules/cactus-web/src/CheckBoxGroup/CheckBoxGroup.tsx
Dhalton: I think we should keep `value` but I'm fine with getting rid of `defaultValue`. I'll wait until we get some other opinions before I go and remove anything, though. cc @mikesoltow @NicolasSimmonds

wilysword: And maybe rename it to `values` or `checked` or `boxesChecked` or something, if we keep it.

NicolasSimmonds: Yeah keeping value and getting rid of defaultValue seems like a good option

mikesoltow: Just one question: would removing `defaultValue` prevent us from seeding the group in the uncontrolled instance?

wilysword: No, you can still set the `defaultChecked` on the individual checkboxes.

mikesoltow: Gotcha, then I'm fine with removing `defaultValue`.  I agree that having a `values` prop (feel free to name it something else) would make it easier when using it as a controlled component, so I'd vote for keeping that one.

Dhalton: Cool, I'll do this and get it pushed up in a bit.

wilysword (approved):

NicolasSimmonds (approved):

PR #343, opened 2020-09-23
CACTUS-333 ActionBar Component

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-333

There are two outstanding issues (that I know of):
1. In the Layout story, sometimes when un-checking the MenuBar knob you get a console warning about setting state of an unmounted component. I don't think that's because of the ActionBar, but I couldn't pinpoint the cause so I'm not positive.
2. The ActionBar doesn't style correctly when it's not inside Layout. I'd like to fix this in a followup ticket where we decouple the `Sidebar` from `Layout`, because it's more refactoring than I felt comfortable with in this ticket.


Dhalton (approved):

NicolasSimmonds (commented):
Are you planning to include the Live code example once the ActionBar.Panel is finished? Or why didn't you include it?

wilysword:
I might add one later, but the components with fixed positioning are rather annoying in the live demo (e.g. try looking at Footer).

NicolasSimmonds (approved):

modules/cactus-web/src/ActionBar/ActionBar.mdx
```diff
@@ -0,0 +1,121 @@
...
+const MySite = () => (
+  <Layout>
+    <ActionBar>
+      <ActionBar.Item
+        icon={<ActionsPrint/>}
+        onClick={printThisPage}
+        aria-label="Print this page"
+      />
+    </ActionBar>
+    <ActionBar.Content>
```
mikesoltow, line 31: I assume you mean Layout.Content here?

mikesoltow (changes requested):
I'm just skimming this out of curiosity (cause it's awesome), not actually reviewing it.  But I did notice one typo in the docs.  Then I'm happy to merge!

modules/cactus-web/src/ActionBar/ActionBar.mdx
wilysword: Find & replace strikes again.

mikesoltow (approved):

PR #346, opened 2020-09-29
Decouple Sidebar from Layout

wilysword:
Though this is not directly related to the ActionBar.Panel (CACTUS-346), it's something I wanted to do in my previous ticket to make future work (including the Panel) easier; I'm putting it up in a separate PR so as not to clutter up the ActionBar tickets, but this is something like CACTUS-333.5 or CACTUS-346.-5.

In order to decouple Sidebar, I changed Layout from directly referencing the Sidebar `styled-components` class, to referencing a generic CSS class that could be assigned to any component. This will also make it easier to cascade styles down the tree based on the current layout position.

While I was at it, I did a little refactoring of the `Footer`:
- Changed exported component to the new paradigm (not directly exposing styled component).
- Changed element from `div` to `footer`.
- Replaced media queries with references to the new CSS classes, which are in turn controlled by the `ScreenSizeProvider`.

### DIFF
--- a/modules/cactus-web/src/ActionBar/ActionBar.tsx
+++ b/modules/cactus-web/src/ActionBar/ActionBar.tsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-import { Sidebar } from '../Layout/Layout'
+import { Sidebar } from '../Layout/Sidebar'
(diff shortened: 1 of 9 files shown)


Dhalton (approved):
LGTM 👍

mikesoltow (approved):
LGTM

PR #347, opened 2020-09-29
fix(cactus-web): ignore width on card view of table and Datagrid

NicolasSimmonds:
Check that the width prop is ignored in card view in both DataGrid and Table

[CACTUS-374]

[CACTUS-374]: https://repayonline.atlassian.net/browse/CACTUS-374

### DIFF
--- a/modules/cactus-web/src/Table/Table.tsx
+++ b/modules/cactus-web/src/Table/Table.tsx
@@ -258,7 +258,7 @@ const StyledCell = styled.td(
     card: css`
       && {
         display: flex;
-        max-width: 100%;
+        width: 288px;
         flex-flow: row wrap;
         justify-content: space-between;
         padding: 8px 16px;
--- a/modules/cactus-web/src/Table/Table.tsx
+++ b/modules/cactus-web/src/Table/Table.tsx
@@ -256,7 +256,7 @@ const StyledCell = styled.td(
     card: css`
       && {
         display: flex;
-        width: 272px;
+        width: 240px;
         flex-flow: row wrap;
         justify-content: space-between;
         padding: 8px 16px;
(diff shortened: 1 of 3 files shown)


Dhalton (approved):
LGTM 👍

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -258,7 +258,7 @@ const StyledCell = styled.td(
     card: css`
       && {
         display: flex;
-        max-width: 100%;
+        width: 288px;
```
mikesoltow, line 261: What would you think about bumping this down to `272`?  I notice that it overflows the screen on iPhone 5.  It overflows at 272 as well, but not by as much, at least.

mikesoltow (changes requested):
Ooops, I actually found an IE11 bug.

Screen Shot 2020-09-30 at 4 45 28 PM

I think you can get rid of this by removing the `max-width: 100%` in the `HeaderBox` and `ContentBox` (for card mode only!), although I'm not sure if that will have any adverse effects.

modules/cactus-web/src/Table/Table.tsx
NicolasSimmonds: Yaeah its fine by me!

Dhalton:
I'm still seeing the text overflowing on the cards in IE11

NicolasSimmonds:
> I'm still seeing the text overflowing on the cards in IE11

Really? Thats weird. I don't see the text overflowing at all

mikesoltow:
@NicolasSimmonds I can reproduce it in IE11 if I reduce the width of the browser window to tiny/small size.  For some reason, changing the width to `240px` fixes it.  And I'd recommend that anyway because that makes the width of the entire card, including padding, equal to 272, which is what the Style Guide intended from what I can see.

Dhalton (approved):
Re-approving for the latest changes. Looks good in IE 👍

mikesoltow (approved):
Agreed...latest changes look good!

PR #349, opened 2020-10-01
Splitbutton variants

NicolasSimmonds:
In  storybook, go to the basic usage split button, and test that the three variants match the Styleguide.

[CACTUS-354]

[CACTUS-354]: https://repayonline.atlassian.net/browse/CACTUS-354

### DIFF
--- a/modules/cactus-web/src/SplitButton/SplitButton.tsx
+++ b/modules/cactus-web/src/SplitButton/SplitButton.tsx
@@ -330,7 +330,7 @@ SplitButton.propTypes = {
   onSelectMainAction: PropTypes.func.isRequired,
   mainActionIcon: PropTypes.elementType,
   disabled: PropTypes.bool,
-  VARIANT: PropTypes.oneOf(['standard', 'danger', 'success']),
+  variant: PropTypes.oneOf(['standard', 'danger', 'success']),
 }
 
 SplitButton.defaultProps = {
--- a/modules/cactus-web/src/SplitButton/SplitButton.tsx
+++ b/modules/cactus-web/src/SplitButton/SplitButton.tsx
@@ -38,7 +38,7 @@ interface SplitButtonActionProps extends Omit<MenuItemImplProps, 'onSelect'> {
   onSelect: () => any
   icon?: React.FunctionComponent<IconProps>
 }
-interface SplitButtonListProps {
+interface VariantInterface {
   variant?: SplitButtonVariant
 }
 
@@ -60,7 +60,7 @@ const getVariantDark = variant({
   `,
 })
 
-const MainActionButton = styled.button`
+const MainActionButton = styled.button<VariantInterface>`
   box-sizing: border-box;
   border: ${(p) => border(p.theme, '')};
   ${(p) => mainShapeMap[p.theme.shape]}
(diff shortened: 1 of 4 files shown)


modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -260,17 +257,74 @@ type SplitButtonType = StyledComponent<typeof SplitButtonBase, DefaultTheme, Spl
...
+      })}
+
+      &:hover,
+      &:focus {
+        ${variant({
+          standard: css`
+            border-color: ${(p): string => p.theme.colors.callToAction};
+          `,
+          danger: css`
+            border-color: ${(p): string => p.theme.colors.error};
```
mikesoltow, line 281:
```suggestion
            border-color: ${(p): string => p.theme.colors.errorDark};
```

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -260,17 +257,74 @@ type SplitButtonType = StyledComponent<typeof SplitButtonBase, DefaultTheme, Spl
...
+          `,
+          success: css`
+            border-color: ${(p): string => p.theme.colors.success};
```
mikesoltow, line 284:
```suggestion
            border-color: ${(p): string => p.theme.colors.successDark};
```

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -281,6 +335,10 @@ SplitButton.propTypes = {
   disabled: PropTypes.bool,
 }
 
+SplitButton.defaultProps = {
+  variant: 'standard',
```
mikesoltow, line 339: We should probably add this to `propTypes` too

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -260,17 +257,74 @@ type SplitButtonType = StyledComponent<typeof SplitButtonBase, DefaultTheme, Spl
...
+      `,
+      success: css`
+        ${(p): ColorStyle => p.theme.colorStyles.success};
+      `,
+    })}
+
+    &:hover,
+    &:focus {
+      ${(p) => !p.disabled && getVariantDark};
+    }
```
mikesoltow, line 327: Why did you move these styles out of the `DropdownButton` styled component?  By putting them here, you're putting them in conflict with the styles defined up there...mostly notably the `disabled` styles.  As a result, the dropdown button now has a color even when the split button is disabled.

mikesoltow (changes requested):
Just a few comments

modules/cactus-web/src/SplitButton/SplitButton.tsx
NicolasSimmonds: I did this so that I didn't have to create an interface for the dropdown button and the Menubutton and pass the variant into those components. But I guess it was a bad call. Will change this

mikesoltow: Ah gotcha.  If you still want to avoid that, you could...you'd just have to make a clear separation of which style attributes go in which group of styles (i.e. `background-color` should be controlled here).

It's probably more readable to change it, but I'm just saying you don'g _have_ to 😄

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -279,6 +330,11 @@ SplitButton.propTypes = {
   onSelectMainAction: PropTypes.func.isRequired,
   mainActionIcon: PropTypes.elementType,
   disabled: PropTypes.bool,
+  VARIANT: PropTypes.oneOf(['standard', 'danger', 'success']),
```
mikesoltow, line 333: Don't think this matters too much, but we might as well change this to lower-case before we merge.

mikesoltow (approved):
Changes look good!

wilysword (approved):

PR #350, opened 2020-10-01
CACTUS-378 ActionBar.Item key fix

wilysword:
Note that I didn't actually make id required, since it could still be used in the absence of ActionProvider.

### DIFF
--- a/modules/cactus-web/src/ActionBar/ActionBar.mdx
+++ b/modules/cactus-web/src/ActionBar/ActionBar.mdx
@@ -23,6 +23,7 @@ const MySite = () => (
   <Layout>
     <ActionBar>
       <ActionBar.Item
+        id="ab-print"
         icon={<ActionsPrint/>}
         onClick={printThisPage}
         aria-label="Print this page"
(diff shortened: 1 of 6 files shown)


modules/cactus-web/src/ActionBar/ActionBar.mdx
```diff
@@ -58,7 +59,7 @@ const MyPage = () => (
 )
 ```
 
-- You must provide either a _key_ or an _id_ prop to uniquely identify the action on the ActionBar (`id` takes precedence over `key`); if neither is present, an error will be thrown.
+- You must provide either an _id_ prop to uniquely identify the action on the ActionBar; if it is not present, an error will be thrown. (We would use _key_ as well, but React doesn't forward it with the props.)
```
mikesoltow, line 62: You can delete the `either` here

mikesoltow (approved):
LGTM...just a tiny thingy in the docs

Dhalton (approved):

NicolasSimmonds (approved):

PR #351, opened 2020-10-01
CACTUS-366 :: Fix IE11 ComboBox bug and mobile Select bugs

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-366

While testing the IE11 fix, I also found two quirks with the mobile "dropdown," so I just knocked those out here as well.  I've separated out the mobile fixes into their own commits if it helps to look at them that way.

### Testing

1. In your normal browser, go through every story under `Select` and make sure everything works normally.
2. In IE11, repeat step 1.
3. You can test the mobile fix on your phone (see https://stackoverflow.com/questions/11005540/localhost-running-on-mac-can-i-view-it-on-my-android-phone for Android instructions...see Dhalton for iOS instructions).  Alternatively, you could use an emulator.  Or you could take my word for it 🤷


Dhalton (approved):
Nice, looks to be working for me on Chrome, IE11, and Safari on iOS. 🚀

NicolasSimmonds (approved):

PR #353, opened 2020-10-07
CACTUS-375 :: Configurable Card Padding

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-375

Testing should be pretty simple; just bring up the storybook and use the new knobs to make sure all of the acceptance criteria are being met.

### DIFF
--- a/modules/cactus-web/src/Card/Card.tsx
+++ b/modules/cactus-web/src/Card/Card.tsx
@@ -55,13 +55,6 @@ export const Card = styled.div<CardProps>`
   padding: ${(p): number => p.theme.space[4]}px;
   ${padding}
   ${(p): ReturnType<typeof css> => getBoxShadow(p.theme, p.useBoxShadow)}
-
-  & > & {
-    padding: ${(p): number => p.theme.space[5]}px;
-  }
-  & > & > & {
-    padding: ${(p): number => p.theme.space[6]}px;
-  }
 `
 
 Card.defaultProps = {
--- a/modules/cactus-web/src/Card/Card.tsx
+++ b/modules/cactus-web/src/Card/Card.tsx
@@ -1,11 +1,11 @@
 import { BorderSize, CactusTheme, ColorStyle, Shape } from '@repay/cactus-theme'
 import PropTypes from 'prop-types'
 import styled, { css } from 'styled-components'
-import { margin, MarginProps, width, WidthProps } from 'styled-system'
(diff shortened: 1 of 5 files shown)


mikesoltow (approved):
This all looks great so I'm going to approve

...with a question (that you'll hate me for only thinking of now 🤦 ):

Should we just abandon the small/medium/large levels and use the standard styled-system levels for consistency?  We could still default `padding` to `4` and tell the users in the docs that we recommend using 4, 5, or 7 (minimum of 5 when nesting).  But that way they have flexibility and can use the same system that we use with other components?

Dhalton:
> Should we just abandon the small/medium/large levels and use the standard styled-system levels for consistency? We could still default `padding` to `4` and tell the users in the docs that we recommend using 4, 5, or 7 (minimum of 5 when nesting). But that way they have flexibility and can use the same system that we use with other components?

Yeah that's probably a good idea. I'll do this today. I think that's actually easier to implement, anyway.

wilysword (approved):

modules/cactus-web/src/Card/Card.tsx
```diff
@@ -53,6 +53,7 @@ export const Card = styled.div<CardProps>`
   ${(p): ColorStyle => p.theme.colorStyles.standard};
   ${(p): ReturnType<typeof css> => getShape(p.theme.shape)}
   padding: ${(p): number => p.theme.space[4]}px;
+  ${padding}
   ${(p): ReturnType<typeof css> => getBoxShadow(p.theme, p.useBoxShadow)}
 
   & > & {
```
mikesoltow, line 59: I know this wasn't in the ACs, but I think that's just because I forgot these were there: I think we should probably delete these hard-coded nested styles and make them configurable by the developer.  Otherwise, these styles will overwrite anything the developer passes in for padding, which is not the intent.

mikesoltow (changes requested):
Just one requested change that was really just an oversight with the ACs.  Feel free to push back if you disagree, however.

modules/cactus-web/src/Card/Card.tsx
Dhalton: Oh, yeah I'm pretty sure I had gotten rid of those the first time but must've forgot to get rid of them again after I reset the commits 🤦 I'll remove them. Thanks for letting me know!

mikesoltow (approved):
Re-approved!

PR #354, opened 2020-10-08
CACTUS-383 :: Fix mobile menu overlay issue

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-383

### DIFF
--- a/modules/cactus-web/src/Layout/Layout.story.tsx
+++ b/modules/cactus-web/src/Layout/Layout.story.tsx
@@ -7,6 +7,7 @@ import ActionBar from '../ActionBar/ActionBar'
 import BrandBar from '../BrandBar/BrandBar'
 import Footer from '../Footer/Footer'
 import MenuBar from '../MenuBar/MenuBar'
+import TextInputField from '../TextInputField/TextInputField'
 import Layout from './Layout'
 
 function action(msg: string) {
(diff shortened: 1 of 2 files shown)


wilysword (approved):
This fix makes absolutely no sense to me. The menu panel already has a z-index of 100, but that doesn't do anything; and somehow adding a z-index to an ancestor helps? It works, but I'm not sure I even _want_ to know why.

mikesoltow:
> This fix makes absolutely no sense to me. The menu panel already has a z-index of 100, but that doesn't do anything; and somehow adding a z-index to an ancestor helps? It works, but I'm not sure I even _want_ to know why.

I _think_ (because my understanding is still poor) that it's because this ancestor has the same stacking context as the other elements in the page.  Therefore the child's z-index only applied within this ancestor's stacking context, and had no effect relative to other elements.

mikesoltow:
> I _think_ (because my understanding is still poor) that it's because this ancestor has the same stacking context as the other elements in the page. Therefore the child's z-index only applied within this ancestor's stacking context, and had no effect relative to other elements.

I re-read this, and it made no sense.  I think what I meant to say is that this ancestor created a new stacking context for some reason, isolating z-index values for all children so that they only have an effect within this ancestor.  I'm still not exactly sure why that is though.

Dhalton (approved):
LGTM, works well

PR #355, opened 2020-10-08
fix(cactus-web): fix height of table with long values

NicolasSimmonds:
There is a knew story for the Table in the storybook in which you can see the result. 
If you want to see a different length, or behavior for other cell, the text should be changed in the code because of the way this is calculated; it does not work as a knob.

[CACTUS-370]

[CACTUS-370]: https://repayonline.atlassian.net/browse/CACTUS-370

### DIFF
--- a/modules/cactus-web/src/Table/Table.tsx
+++ b/modules/cactus-web/src/Table/Table.tsx
@@ -20,7 +20,6 @@ interface TableContextProps {
   headers?: React.ReactNode[]
   cellIndex: number
   variant: TableVariant
-  maxHeights: number[]
 }
 
 interface TableProps
@@ -68,7 +67,6 @@ const DEFAULT_CONTEXT: TableContextProps = {
   inHeader: false,
   cellIndex: 0,
   variant: 'table',
-  maxHeights: [],
 }
 
 const TableContext = createContext<TableContextProps>(DEFAULT_CONTEXT)
@@ -84,6 +82,32 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
   ({ children, cardBreakpoint, ...props }, ref): React.ReactElement => {
     const size = useContext(ScreenSizeContext)
     const context: TableContextProps = { ...DEFAULT_CONTEXT, headers: [] }
+    const tableRef = React.useRef<HTMLTableElement>(null)
+    const mergedRef = useMergedRefs(ref, tableRef)
+    useLayoutEffect(() => {
+      const heightInfo = []
+      if (context.variant === 'card') {
+        const cells = tableRef.current?.querySelectorAll(`${StyledCell}`)
+        if (cells?.length > 0) {
+          for (const cell of cells as HTMLTableCellElement[]) {
+            const cellIndex = cell.cellIndex
+            const info = heightInfo[cellIndex] || (heightInfo[cellIndex] = { max: 0, cells: [] })
+            const height = cell.clientHeight
+            info.max = Math.max(info.max, height)
+            info.cells.push({ height, cell })
+          }
+          // We queue the updates then make them all at once to avoid layout thrashing.
+          for (const info of heightInfo) {
+            for (const cellInfo of info.cells) {
+              if (cellInfo.height !== info.max) {
+                cellInfo.cell.style.minHeight = `${info.max}px`
+              }
+            }
+          }
+        }
+      }
+    })
+
     if (props.variant) {
       context.variant = props.variant
     } else if (cardBreakpoint && size <= SIZES[cardBreakpoint]) {
@@ -93,12 +117,12 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
     return (
       <TableContext.Provider value={context}>
         {props.as ? (
-          <StyledTable {...props} ref={ref}>
+          <StyledTable {...props} ref={mergedRef}>
             {children}
           </StyledTable>
         ) : (
           <Wrapper fullWidth={props.fullWidth}>
-            <StyledTable {...props} ref={ref}>
+            <StyledTable {...props} ref={mergedRef}>
               {children}
             </StyledTable>
           </Wrapper>
@@ -107,43 +131,6 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
     )
   }
 )
-interface CardCellProps {
-  ref: React.MutableRefObject<HTMLTableDataCellElement | null> | any
-  children: React.ReactNode
-  colSpan: number
-  ctx: TableContextProps
-}
-
-const CardCell: React.FC<CardCellProps> = (props: CardCellProps): React.ReactElement => {
-  const { ref, colSpan, children, ctx, ...rest } = props
-  const context = useContext<TableContextProps>(TableContext)
-  const innerRef = React.useRef<HTMLTableDataCellElement>(null)
-  const combinedRef = useMergedRefs(ref, innerRef)
-
-  useLayoutEffect(() => {
-    if (innerRef && innerRef.current) {
-      const cellIndex = innerRef.current.cellIndex
-
-      const maxHeight = context.maxHeights[cellIndex]
-      if (innerRef.current.clientHeight > maxHeight || maxHeight === undefined) {
-        context.maxHeights[cellIndex] = innerRef.current.clientHeight
-      }
-    }
-  }, [context.maxHeights])
-
-  if (innerRef.current)
-    innerRef.current.style.minHeight = `${context.maxHeights[innerRef.current?.cellIndex]}px`
-
-  const headerContent = context.headers && context.headers[context.cellIndex]
-  context.cellIndex += colSpan
-
-  return (
-    <StyledCell ref={combinedRef} {...rest}>
-      {headerContent && <HeaderBox>{headerContent}</HeaderBox>}
-      <ContentBox>{children}</ContentBox>
-    </StyledCell>
-  )
-}
 
 export const TableCell = React.forwardRef<HTMLTableDataCellElement, TableCellProps>(
   ({ children, ...props }, ref): React.ReactElement => {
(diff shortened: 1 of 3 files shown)


Dhalton (commented):
Looks good for the most part, but I did notice some weirdness as I was testing. If you open the new story you added for the long values, and then go back to the basic usage story and change the screen size so that it renders in card view, you'll see the "row" has the same height as it did for the other story, even though there's no need for it.

It looks like the context still has the same value for `largestCell`, causing it to be rendered in the same way. I wonder if cleaning out that context when the component unmounts would fix the issue. That's what I'd try.

NicolasSimmonds:
> Looks good for the most part, but I did notice some weirdness as I was testing. If you open the new story you added for the long values, and then go back to the basic usage story and change the screen size so that it renders in card view, you'll see the "row" has the same height as it did for the other story, even though there's no need for it.
> 
> It looks like the context still has the same value for `largestCell`, causing it to be rendered in the same way. I wonder if cleaning out that context when the component unmounts would fix the issue. That's what I'd try.

Yeah I did tried this for other issue but form some reason the styles wont't apply to the first row. I can try and find another solution to it though

wilysword:
Shouldn't you commit the changes first so we can review the most up-to-date version, rather than having to review it twice?

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -259,7 +300,11 @@ const StyledCell = styled.td(
         width: 240px;
         flex-flow: row wrap;
         justify-content: space-between;
+        align-items: flex-start;
         padding: 8px 16px;
+        > div[aria-hidden='true'] + div {
+          align-self: flex-end;
+        }
```
wilysword, line 307: I'd move this to the `ContentBox` styles as
```
${HeaderBox}:not(:empty) + & {
  align-self: flex-end;
}
```

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -69,6 +75,10 @@ const DEFAULT_CONTEXT: TableContextProps = {
   inHeader: false,
   cellIndex: 0,
   variant: 'table',
+  largestCell: {
+    height: 0,
+    cell: 0,
+  },
 }
```
wilysword, line 82: I think the problem Dhalton mentioned is because when you make a copy of the context it's a shallow copy, meaning this `largestCell` property is actually a global value, the way you're using it.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -108,23 +118,54 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
+      useLayoutEffect(() => {
+        if (combinedRef && combinedRef.current) {
+          if (combinedRef.current.clientHeight > context.largestCell.height) {
+            context.largestCell.height = combinedRef.current.clientHeight
+            context.largestCell.cell = combinedRef.current.cellIndex
+          }
+        }
+      }, [combinedRef, context.largestCell.cell, context.largestCell.height, ref, children])
+      if (combinedRef.current && combinedRef.current.cellIndex === context.largestCell.cell) {
+        combinedRef.current.style.height = `${context.largestCell.height - 16}px`
```
wilysword, line 161: Assuming this -16 is a padding issue, we should probably make the cells `box-sizing: border-box;`.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -69,6 +75,10 @@ const DEFAULT_CONTEXT: TableContextProps = {
   inHeader: false,
   cellIndex: 0,
   variant: 'table',
+  largestCell: {
+    height: 0,
+    cell: 0,
```
wilysword, line 80: I think this wouldn't work if you had multiple cells with large values; you'd need to get the largest cell _per column_.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -108,23 +118,54 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
       context.cellIndex += colSpan
     } else if (context.variant === 'card') {
+      const innerRef = React.useRef(null)
+      const combinedRef = useCombinedRefs(ref, innerRef)
+
+      useLayoutEffect(() => {
+        if (combinedRef && combinedRef.current) {
+          if (combinedRef.current.clientHeight > context.largestCell.height) {
+            context.largestCell.height = combinedRef.current.clientHeight
+            context.largestCell.cell = combinedRef.current.cellIndex
```
wilysword, line 156: Should probably account for `colSpan`, like the calculations on `context.cellIndex`.

wilysword (changes requested):
The algorithm seems like a good start, but it's still pretty flawed:

- It's using what is effectively a global variable
- It appears to only work for a single column at a time

I also have a couple of suggestions to make your algorithm more flexible:
- Instead of `style.height`, set `style.minHeight`
- On the tallest cell(s), set `style.minHeight = ''`; that way if the value in that cell shrinks, you can recalculate the min height for all the other cells in the column

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -108,23 +118,54 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
+          }
+        }
+      }, [combinedRef, context.largestCell.cell, context.largestCell.height, ref, children])
```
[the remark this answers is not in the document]
mikesoltow: I'm bummed that the linter didn't catch this 😢   After fiddling, the linter only catches it in regular components, not in a forwardRef callback.  It's weird, though, cause they had an issue that fixed this a while back.  I wonder if using TypeScript messes with their fix somehow.  https://github.com/facebook/react/issues/17220

modules/cactus-web/src/Table/Table.tsx
NicolasSimmonds: I don't think I understand what you meant here

wilysword: Look up at lines 143-147. `colSpan` is an attribute that makes a cell take up more than one column, and it's not the same for every cell in a column. If you have a cell at index 1 with `colSpan=2`, then the cell index of the next cell in the row is 3, not 2, because cell 1 takes up two columns. Not sure I'm very good at explaining that, so let me know if I'm still not being clear.

NicolasSimmonds: So to see if I understood correctly, I should take the colSpan value into account when it's 2 or more and make the cell at index x and subsequent cells untill colSpan value, the same height?

modules/cactus-web/src/Table/Table.tsx
NicolasSimmonds: The problem is that if I don't make this a global variable, the value of the largest cell(s) is lost after each TableCell is rendered. So I have no way to persist those values. Any Ideas on this?

modules/cactus-web/src/Table/Table.tsx
wilysword: Hmm, I hadn't thought about the actual algorithm, and now that I am I'm coming up blank. We may want to come back to this eventually, but I think for now it might be best to just ignore `colSpan` after all. Your understanding of the issue seems correct, though.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -108,23 +118,54 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
+    refs.forEach((ref: typeof targetRef) => {
+      if (!ref) return
+      else if (ref.current) {
+        ref.current = targetRef.current
+      }
+    })
+  }, [refs])
+
+  return targetRef
+}
```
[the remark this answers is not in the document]
NicolasSimmonds: Ive been trying to use the function ate helpers/react.ts, but I don't think I fully understand it. When I use it and save the combined refs in a variable, the variable is a function, not an actual ref so I can't access the current property or anything

modules/cactus-web/src/Table/Table.tsx
NicolasSimmonds: The problem Dhalton mentioned is already resolved. But I'm not sure how to modify the context in a non-global way. The way cellIndex and headers is being modified is the same as largestCell, as far as I can tell.

modules/cactus-web/src/Table/Table.tsx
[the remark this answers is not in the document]
wilysword: You've got the code
```
const innerRef = React.useRef(null)
const combinedRef = useCombinedRefs(ref, innerRef)
```
and then you use the `combinedRef` as if it were a normal ref, and just ignore `innerRef`. What you want is
```
const innerRef = React.useRef<HTMLTableDataCellElement>(null)
const combinedRef = useMergedRefs(ref, innerRef)
...
<StyledCell {...props} ref={combinedRef}>
```
so you still pass the combined ref to the element, but change all other references to `combinedRef` to use `innerRef` instead.

modules/cactus-web/src/Table/Table.tsx
wilysword: It's not the same, because `largestCell` is a nested object. Consider the following:
```js
const x = { one: 'uno', two: { value: 'dos' } }
const y = { ...x }
y.one = 'ein'
y.two.value = 'zwei'

x.one === 'uno'
x.two.value === 'zwei'
```
`y` is a _shallow_ copy of `x`, so you can change values at the top level without affecting the original; but the nested object `two` is the same reference in both so changes to one affect the other. That's what's happening with the context. To stop doing that you either need to stop nesting those values, or do a _deep_ copy of the default context.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -16,20 +17,26 @@ type BorderCorner = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right'
 
 type TableVariant = 'table' | 'card'
 
+interface CellHeightProps {
```
Dhalton, line 20: I don't think you need this interface anymore. I don't see it being used anywhere

wilysword (commented):
I couldn't get it to update properly like this, so after playing around for a bit I came up with https://github.com/repaygithub/cactus/commit/50b83f9d176b040134efc3c560dd3b4bb0d69848. It doesn't require any of the context changes, and it seems to work pretty well; it can grow if the cell contents grow, but it can't shrink if they shrink.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -397,7 +436,7 @@ const card = css`
   tr {
     display: flex;
     flex-flow: column nowrap;
-    min-width: 272px;
+    min-width: fit-content;
```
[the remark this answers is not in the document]
NicolasSimmonds: The reason for this is that when I changed the contentbox to box-sizing: border-box, the width was smaller than the actual cell. I guess I can change that

wilysword (commented):
Code looks good, and everything seems to be working. Unfortunately, as usual IE can't allow us to have nice things:
ie-table

Adding `box-sizing: border-box;` to `StyledCell` made it look better, so it's probably due to a discrepancy between the `tr` width and the cell width.

Dhalton (approved):
Nice, code looks good and it's working much better now. Thanks for sticking this out 👍

wilysword (approved):

PR #356, opened 2020-10-09
CACTUS-347 :: Storyshots

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-347

### A Few Notes
- I had to downgrade Jest to v24.0.1 to get the storyshots to work...it was either that or upgrade to storybook v6.
- `Spinner` isn't being tested by the storyshots because it's always moving so I don't think there's a way we can get the snapshots to match consistently on that one.
- There's a `test:visual` yarn script in the `cactus-web` directory, but I wouldn't recommend running this command outside of Docker; the storyshots should be run in a Docker container so they're always run on the same OS and therefore won't have any variations with text sizing and such.
- They are intended to be run from the root using the `test:build` and `test:docker` commands.
- I haven't the faintest clue why, but the docs website tests were failing because of a type that Gatsby defines. I added `--skipLibCheck` to get around this for now, but we should probably figure out why that is at some point.

### Testing
1. Pull down the branch and run `test:build`. Wait for Docker to finish building.
2. Run `test:docker` and make sure all of the snapshots pass.


modules/cactus-web/package.json
```diff
@@ -73,14 +82,19 @@
...
     "jest-styled-components": "^7.0.2",
     "prop-types": "^15.7.2",
+    "puppeteer": "^5.3.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-test-renderer": "^16.13.1",
     "rollup-plugin-commonjs": "^10.1.0",
+    "start-server-and-test": "^1.11.5",
     "styled-components": "^5.1.1",
+    "ts-jest": "^26.4.1",
```
mikesoltow, line 97: What does installing this do?  It looks like you didn't set it as a preset in the Jest config.

mikesoltow (approved):
I'll approve, but I do have one question on `ts-jest`.  Other than that, this looks great!  Thank you very much for the work you put in.  We'll have to keep an eye on the runtime for the tests to see if that becomes a problem, but from a testing confidence point of view I think this is going to be great.

Also, for the `--skipLibCheck` issue, it looks like it's some resolution conflict with the `@types/reach__router` package.  Both Gatsby and Storybook depend on it, but use different versions.  I'm not really sure why yarn's having trouble with it (which is somewhat worrying...).  That being said, I bet when I do the dependency upgrades for this repo and upgrade Storybook that this will be fixed.

modules/cactus-web/package.json
Dhalton: Oh, we don't actually need this. I added it at some point when I was trying to get things working & must've forgotten to remove it. Whoops! Nice catch.

wilysword (approved):
This is great! I'm fine with merging as-is, but for future consideration, I do have some questions:
1. Some of the stories are visually identical, only differing in how they behave; there's no point in having storyshots for all versions of those, so is there a way to omit individual stories, that you know of? (I saw how you omitted Spinner, but I'm not sure if that would work for this).
2. A lot of components have parts that are only visible under certain conditions (drop-downs, tooltip, etc); in fact the MenuBar on small screen sizes, 90% of the styles are only visible when it's expanded. Do you know if there's a way to test those sorts of modifications without writing special stories for that purpose?

Like I said, I think this is fine for a first pass, but there are still pretty significant gaps in the coverage.

Dhalton:
> 1. Some of the stories are visually identical, only differing in how they behave; there's no point in having storyshots for all versions of those, so is there a way to omit individual stories, that you know of? (I saw how you omitted Spinner, but I'm not sure if that would work for this).

I was just doing some looking around and didn't find anything quite like what we'd need to exclude individual stories in that way. I believe `storyKindRegex` is compared to the string set when `storiesOf` is called, so I don't think we could use that to omit sub-stories for each component. You make a good point, though. It is pointless to have multiple identical storyshots, but we may have to come up with some kind of workaround to avoid it.

> 2. A lot of components have parts that are only visible under certain conditions (drop-downs, tooltip, etc); in fact the MenuBar on small screen sizes, 90% of the styles are only visible when it's expanded. Do you know if there's a way to test those sorts of modifications without writing special stories for that purpose?

It _seems_ like this [puppeteerTest](https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer?source=post_page---------------------------#puppeteertest) configuration might be something that could help with that. Basically it looks like you can execute puppeteer test code in the story to interact with the elements, although admittedly, based on the docs I linked, I'm not 100% sure how that's supposed to work with stories, at least with the way ours are set up.

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+WORKDIR /code
+COPY package.json yarn.lock /code/
+COPY modules/cactus-web/package.json /code/modules/cactus-web/
+RUN yarn --frozen-lockfile
+COPY modules/cactus-web/.jest/ /code/modules/cactus-web/.jest/
+COPY modules/cactus-web/.storybook/ /code/modules/cactus-web/.storybook/
+COPY modules/cactus-web/cactus-addon/ /code/modules/cactus-web/cactus-addon
+COPY modules/cactus-web/.babelrc /code/modules/cactus-web
+COPY modules/cactus-web/src/ /code/modules/cactus-web/src
+COPY modules/cactus-web/tests/ /code/modules/cactus-web/tests
```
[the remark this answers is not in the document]
Dhalton: Yeah we probably could just copy the whole directory. Initially, as I was setting this up, I thought we'd only need a few things from the `cactus-web` directory, but the list just kept growing and I never thought to condense it down to the one copy command. I'd say we should condense it down whenever we move on to the next steps in terms of fleshing these tests out.

PR #357, opened 2020-10-09
CACTUS-377 Action/Brand Bar Integration

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-377

Note that the styling doesn't really work unless it's inside of Layout, at which point my original question of what to do when there's no ActionBar almost becomes moot: there's basically no reason _not_ to use ActionBar if they're using Layout.

Had some somewhat unrelated changes as well:
- Fixed the `as` prop on `ActionBar.PanelProp` (my original implementation was lazy and only worked with intrinsic elements).
- Changed `Sidebar.Button` from _always_ having `role=button` to having that be the default.
- Removed the `role=none` from the main menu when it's on the sidebar (so it will use the default role for `<nav>` elements).
- Changed the `ScreenSizeProvider` to always use the same `ScreenSize` instances, both to prevent unnecessary re-renders due to context changes and to enable comparison with `SIZES` using `===`.


modules/cactus-web/src/BrandBar/BrandBar.mdx
```diff
@@ -15,11 +15,11 @@ The `BrandBar` Components allows yout to render a Logo of your choice and a user
 
 export const code = `<BrandBar
     isProfilePage
-    userMenuText'Hershell Jewess'
+    userMenuText="Hershell Jewess"
```
luisniebla, line 18: ❤️

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -3,127 +3,141 @@ import {
...
+        <ActionMenuButton as={ReachMenuButton as any} $isProfilePage={isProfilePage}>
+          <DescriptiveProfile />
+          <VisuallyHidden>{userMenuText}</VisuallyHidden>
+        </ActionMenuButton>
+        <ActionBar.PanelPopup as={ActionMenuPopup} portal={false}>
+          <PopupHeader aria-hidden>
+            <DescriptiveProfile mr="8px" />
+            {userMenuText}
+          </PopupHeader>
+          <ReachMenuList>{children}</ReachMenuList>
```
Dhalton, line 93: I think we may need to do some styling on `ReachMenuList` to add `outline: none;`. In Safari, when you click the user icon to open the menu, this list gets focus and it Safari is adding its own outline here, so it just looks a little off. Let me know if you need a screenshot for more clarification.

Dhalton (commented):
Looks great and works well overall. There was one thing I noticed when I was testing, but that's pretty much it as far as I can tell! 👍

Dhalton (approved):
LGTM thanks for fixing that outline in Safari

mikesoltow (approved):
LGTM 🎉

PR #359, opened 2020-10-15
CACTUS-287 :: Theme PropTypes

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-287

I ended up having to `@ts-ignore` this for a couple of types in the theme that make it pretty difficult (maybe impossible?) to get the PropTypes to agree with TypeScript. The "problem" types I'm referring to are the ones associated with `fontSizes`, `mobileFontSizes`, and `iconSizes`. These attributes should all be set to arrays of numbers, but the arrays end up with some extra keys that are mapped to the array values, and the TypeScript types reflect this. However, I have not been able to find a way to define an array with those extra keys via PropTypes. I have custom validators set up to check the length and the types of the values set in the arrays, but I wasn't able to convince TS that that is acceptable. However, if anyone has any suggestions, I'm all ears.

### Testing
Pull down this branch, run storybook, and open the console. Make sure there aren't any PropTypes errors. If you want to try to trigger some by modifying the theme, you can do so in `cactus-addon/index.tsx`. The values I'd most recommend testing would be the ones mentioned above: `fontSizes`, `mobileFontSizes`, and `iconSizes`.

### DIFF
--- a/modules/cactus-web/src/StyleProvider/StyleProvider.tsx
+++ b/modules/cactus-web/src/StyleProvider/StyleProvider.tsx
@@ -107,9 +107,110 @@ export const StyleProvider: React.FC<StyleProviderProps> = (props): React.ReactE
   )
 }
 
-// @ts-ignore
+// PropTypes declarations of some recurring types so they can be re-used.
+const textStyleObj = PropTypes.shape({
+  fontSize: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


wilysword (approved):

mikesoltow (approved):
LGTM!  I didn't spend a ton of time on the TypeScript stuff, but I don't think it's worth it.

PR #362, opened 2020-10-21
feat(cactus-web): add useBoxShadows prop to Accordion

NicolasSimmonds:
In storybook, go to the accordion, and in the _**Basic Usage**_ and _**With outline**_ you'll see a knob for the useBoxShadows prop.

Play with it and with the theme's boxshadows and make sure everything works as expected

[CACTUS-376]

[CACTUS-376]: https://repayonline.atlassian.net/browse/CACTUS-376


Dhalton:
I feel like, if `useBoxShadows` is `false` on the outline variant, we should keep an outline even when the accordion is open. It looks strange when it has an outline and then it disappears when you open the accordion. That's just my opinion of course, so maybe we should talk about it as a team.

mikesoltow (approved):
LGTM.  I'm curious about the churn in the image snapshots, but it looks like it's just pulling in the recent changes Glen made.  My guess is that this was another "race condition" with a branch that started before the visual tests were merged.  Hopefully this is less of an issue going forward?

NicolasSimmonds:
> LGTM. I'm curious about the churn in the image snapshots, but it looks like it's just pulling in the recent changes Glen made. My guess is that this was another "race condition" with a branch that started before the visual tests were merged. Hopefully this is less of an issue going forward?

Yeah I also think it has something to do with that. Hopefully after this PR is merged, it no longer will be an issue

Dhalton (approved):
LGTM, thanks for adding those borders. Yeah hopefully we don't have so much trouble with the image snapshots moving forward. :/

PR #364, opened 2020-10-23
CACTUS-394 :: Manually add Select polyfills to Storybook webpack

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-394

This ticket was created because the multi-select wasn't working properly for the deployed Storybook in IE11 (see reproduction steps in the ticket).  However, it turns out that was the _only_ place it wasn't working.  In IE11, a polyfill is required to get the Select component to work, and we manually include that in `src/helpers/polyfills.ts`.  For some reason, that polyfill wasn't being included in the Storybook Webpack build, even though it was being included for regular Wepback builds.  **In other words, this bug never actually appeared in our end users' prod apps.**

I still have no idea why it wasn't working, but I decided to cut my losses and just manually add the polyfill file to the `entry` field of the Storybook Webpack config.  Not worth spending too much time on when it's working for our real users.

### Testing

I published the updated Storybook using this config, so you can check out the story at https://repaygithub.github.io/cactus/stories/cactus-web/?path=/story/select--with-multiselect.

Not sure if there's anything else you can test.  You might want to verify that your local Storybook works normally too.


NicolasSimmonds (approved):

Dhalton (approved):
LGTM!

PR #365, opened 2020-10-23
Fix BrowserStack Tests

mikesoltow:
Tests hang for some reason when run in parallel, so this changes makes them run sequentially.  It's a major bummer, but we'll try to look into why this is happening later.

### DIFF
--- a/examples/mock-ebpp/run-ete-tests.js
+++ b/examples/mock-ebpp/run-ete-tests.js
@@ -19,7 +19,7 @@ program.requiredOption(
 )
 program.parse()
 
-const runTests = async () => {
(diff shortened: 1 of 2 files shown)


Dhalton (approved):
Weird that this just started happening out of nowhere. Code looks good to me though

mikesoltow:
> Weird that this just started happening out of nowhere. Code looks good to me though

Yeah, we're going to have to look into that.  My hunch is that it's a resource issue on the GitHub side...so maybe something changed with our plan?  or GitHub Actions in general?  I'd blame BrowserStack if it was happening locally...

NicolasSimmonds (approved):

PR #367, opened 2020-10-23
CACTUS-392 :: Dependency updates

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-392

~I had to update all of the visual snapshots, but I could confirm that this was only caused by the Storybook update (they passed on the previous commit and failed as soon as I upgraded Storybook).  For some reason, Storybook v6 changed the viewport size.~

UPDATE: It looks like the viewport change was caused by Storybook adding a `1 rem` padding on the `body` element in v6.  I was able to get rid of that, and it fixed most of the snapshots so that they didn't have to be updated.

However, there are still a few snapshots that created diffs.  It looks like they represent changes related to _different_ stories.  I think they weren't caught in those stories because we currently have a 2% failure threshold for visual snapshots.  I'm currently experimenting with removing that, and might push up a PR shortly.

### Testing

First, pull down the branch, run `yarn install`, and run `yarn cleanup && yarn build`.

If you'd like to be thorough, you can run all of these tests on IE11 as well as your local browser.

#### cactus-i18n changes

1. Run `yarn w standard start`.
2. Pull up the app at `https://localhost:3435`
3. Make sure toggling between languages works as you'd expect.

#### Storybook Update

1. Run `yarn web start`
2. Click around and make sure everything seems to be working normally.

#### Website Updates

1. Run `yarn docs start`
2. Click around and make sure nothing looks wrong.

#### React Upgrades

1. Run `yarn w mock-ebpp start`, as that's probably the most fully-featured app
2. Click around as much as you can.  Pay special attention to events, including focus events, just to make sure nothing blows up.


NicolasSimmonds (approved):
Everything worked good for me, didn't see any odd behavior.

wilysword (approved):

PR #369, opened 2020-10-24
Footer improvements

NicolasSimmonds:
You can review the changes to the footer on its own StoryBook component, and also in the Layout.

Play with different screen-sizes to make sure it works as expected in all of them.

[CACTUS-385]

[CACTUS-385]: https://repayonline.atlassian.net/browse/CACTUS-385


wilysword (approved):

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -55,12 +55,13 @@ const LogoWrapper = styled('div')`
 `
 
 const ContentWrapper = styled('div')`
-  max-width 100%;
+  max-width: 100%;
+  font-size: 15px;
```
mikesoltow, line 59: Can we make this theme-based and pull in the textStyles object?

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -83,6 +84,7 @@ const LinkCol = styled('div')<LinkColProps>`
 
   ${Link} {
     max-width: 100%;
+    font-size: 15px;
```
mikesoltow, line 87: Same here

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -199,6 +201,7 @@ const StyledFooter = styled.footer.attrs({ role: 'contentinfo' as string })`
   justify-content: center;
   width: 100%;
   background-color: ${(p) => p.theme.colors.lightContrast};
+  margin-top: auto;
```
mikesoltow, line 204: If you change to using the `flow` setting, then I don't think you need this.

modules/cactus-web/src/Layout/Layout.tsx
```diff
@@ -113,6 +113,8 @@ const LayoutWrapper = styled.div<LayoutProps>(
   right: 0;
   top: 0;
   bottom: ${p.fixedBottom}px;
+  display: flex;
+  flex-direction: column;
```
mikesoltow, line 117: If you change to using the `flow` setting, then I don't think you need this.

modules/cactus-web/src/Layout/Layout.tsx
```diff
@@ -139,7 +141,8 @@ const LayoutWrapper = styled.div<LayoutProps>(
 
   ${
     !p.floatLeft
-      ? 'display: block;'
+      ? `display: flex;
+        flex-direction: column;`
```
mikesoltow, line 145: If you change to using the `flow` setting, then I don't think you need this.

mikesoltow (changes requested):
It looks like this doesn't actually solve the following AC: "On tiny screens, the user must still scroll down to see the Footer if there is enough content to fill the viewport height."

If you look at the Layout story (and the associated visual snapshot) you'll see that content is running off the page, but the footer still displays over some of the text.  To get it to work, you'll need to pass in `flow` to the `useLayout` hook on line 146 in `Footer.tsx`, as that will move the Footer below all content in every instance.

If you do that, you'll also need to change styles in the LogoAndContentSection and LogoWrapper that rely on `.cactus-layout-fixedBottom` to rely on mediaQueries (using `${(p) => media(p.theme, 'small')}`) instead.

wilysword:
@mikesoltow I'd really prefer if we didn't use media queries in these layout components.

modules/cactus-web/src/Layout/Layout.tsx
mikesoltow: Failure...ignore this

modules/cactus-web/src/Layout/Layout.tsx
mikesoltow: I need to re-learn how to read.  Please ignore this.

modules/cactus-web/src/Footer/Footer.tsx
mikesoltow: So much shame.  Ignore this.

mikesoltow (commented):
@NicolasSimmonds Changes look good, but you'll want to merge in `master` and update the visual snapshots.  The line height made a slight change to the footer, I think.

mikesoltow (approved):
🎉

PR #372, opened 2020-10-27
Change visual snapshot comparison method to SSIM

mikesoltow:
This should hopefully reduce false positives.  See
https://github.com/americanexpress/jest-image-snapshot#recommendations-when-using-ssim-comparison

To test, simply run `yarn test:build` followed by `yarn test:docker` on your machine, and make sure they pass.

### DIFF
--- a/modules/cactus-web/tests/storyshots.test.ts
+++ b/modules/cactus-web/tests/storyshots.test.ts
@@ -7,11 +7,17 @@ const supportedDevices = ['iPhone 5', 'iPad']
 
 interface MatchOptions {
   customSnapshotsDir?: string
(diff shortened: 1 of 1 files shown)


Dhalton:
I'm not sure what's going on, but I haven't been able to get these to pass on my machine. I keep getting the following error
`FATAL ERROR: MarkCompactCollector: young object promotion failed Allocation failed - JavaScript heap out of memory` 😳

Dhalton (approved):
Turns out I'm just an idiot and only had 2GB of RAM allocated to Docker. All passing for me now!

NicolasSimmonds (approved):
They passed for me too!

PR #373, opened 2020-10-27
Fix overflow for long words in table card view

mikesoltow:
See https://repay.slack.com/archives/CABGTU4S3/p1603824711026100

### Testing

1. Open up the Table -> With Long Values story
2. Change one of the values provided to be a long word instead of text with a bunch of spaces.
3. Ensure that the text wraps and does not overflow the card
4. Open up the Table -> Layout story
5. Switch to the Card variant and play around with the Cell and Header text to be sure that everything is behaving as you'd expect for both short and long values.


NicolasSimmonds (approved):

NicolasSimmonds (approved):

Dhalton (approved):
LGTM!

PR #374, opened 2020-10-27
feat(cactus-web): add dividers option to table and Datagrid components

NicolasSimmonds:
Test:

There is a knob in the Table: layout and DataGrid: basic-usage for you to activate the Dividers on the components.

[CACTUS-395]

[CACTUS-395]: https://repayonline.atlassian.net/browse/CACTUS-395


modules/cactus-web/src/Table/Table.tsx
```diff
@@ -354,6 +363,7 @@ const table = css`
   tr:nth-of-type(even) {
     td,
     th {
+      border-right: ${(p): string => (p.dividers ? border(p.theme, 'white') : '')};
```
wilysword, line 366: This overrides the border on the last cell in the row, which makes it uneven because the table border is still supposed to be uniformly `lightContrast`. Aside from that, this looks good (although personally I find the alternating colors to be a little trippy).

wilysword (changes requested):

modules/cactus-web/src/Table/Table.tsx
NicolasSimmonds: Yeah I also think it looks weird, and also the hover state for the rows.

wilysword (approved):

mikesoltow (approved):
LGTM!

PR #375, opened 2020-10-27
CACTUS-398 :: Button Polymorphism

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-398

Pretty straightforward...the only thing I want to mention is that the prop types for `as` are being `@ts-ignored` because apparently I'm not able to generate an array of keys from `JSX.IntrinsicElements`, so I don't have a way to explicitly state which strings are allowed without listing them all out by hand. Kind of annoying 😠

### Testing
1. Start the storybooks and open the new Button "as Link" story.
2. Inspect the html and make sure an `<a />` tag is rendered in place of `<button />`.
3. Make sure there aren't any style changes compared to a regular button
4. Click the button & make sure the link works. It should take you to Google.


modules/cactus-web/src/Button/Button.tsx
```diff
@@ -19,6 +19,8 @@ interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement>, Mar
   loadingText?: string
 }
 
+type ButtonPropsWithAs = ButtonProps & AsProps<GenericComponent>
```
wilysword, line 22: When you take away the generic-ness like this, you're basically making props the `any` type. This code should fail on every line, but it passes the type check:
```typescript
const Link: React.FC<{ to: string }> = ({ to, children }) => (<a href={to}>{children}</a>)

const X = () => {
  const r1 = React.useRef<HTMLButtonElement>()
  const r2 = React.useRef<HTMLButtonElement>()
  const r3 = React.useRef<HTMLAnchorElement>()
  return (
    <>
      <Button type="button" ref={r1} value="there" href="fom">Hey</Button>
      <Button as="a" ref={r2}>Hey</Button>
      <Button as="a" ref={r3} href="fom" type="button" random="">Hey</Button>
      <Button as={Link}>Hey</Button>
    </>
  )
}
```
The ref stuff is optional so I'll leave that up to you, but you should make the `Button` function generic so that it will actually type check the props. (See ActionBar and MenuBar for examples though, like I said, you don't have to do the forwardRef + workaround if you don't want to.)

modules/cactus-web/src/Button/Button.tsx
```diff
@@ -161,35 +163,36 @@ const variantOrDisabled = (props: ButtonProps) => {
...
   ...rest
 }): React.ReactElement => {
-  const props = omitMargins(rest)
   let spanProps = null
   if (loading === true) {
     spanProps = { style: { visibility: 'hidden' } as React.CSSProperties, 'aria-hidden': true }
   }
+  if (rest.as) {
+    delete rest['inverse']
+  }
```
wilysword, line 179: For `inverse` and `variant` you could convert them to [transient props](https://styled-components.com/docs/api#transient-props), and for the margins you could use `shouldForwardProp`:
```
const StyledButton = styled.button.withConfig({
  shouldForwardProp: (p) => !margin.propNames.includes(p)
})<{ $inverse?: boolean, $variant?: Button>`
  // etc.
`

**Review by wilysword — CHANGES_REQUESTED** [review]
You should change how you're utilizing the `AsProps` type. The second comment isn't a request, it's just an alternate way of doing things, so you can take it or leave it.

**Review by wilysword — APPROVED** [review]
Looks good enough to approve, but I do have two more suggestions.

You can remove the `React.ButtonHTMLAttributes<HTMLButtonElement>` from the `ButtonProps` interface, that's added by `AsProps<'button'>`.

The docs page for Button has a section on polymorphism that is now out-of-date. I suggest removing the part that talks about a TSX solution, and changing the example to either use `as="a" href="..."` _or_ make it clear that the `Link` it's talking about is from a routing library, not the cactus Link (to avoid styling conflicts).

File: modules/cactus-web/src/Button/Button.tsx [hunk 3]
```diff
@@ -231,15 +241,23 @@ export const Button = styled(ButtonBase)<ButtonProps>`
...
   disabled: PropTypes.bool,
   inverse: PropTypes.bool,
   loading: PropTypes.bool,
   loadingText: PropTypes.string,
+  // @ts-ignore
+  as: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
 }
 
-Button.defaultProps = {
+ButtonFR.defaultProps = {
```
mikesoltow, line 260: So I noticed a funny thing here: there's a `type` default prop being set to `button`.  I can see why that was put in there, but now that this is polymorphic, we'll either need to:
* Remove `type` from `defaultProps` and document that the user is responsible for providing this if using the normal button behavior, or
* Add `type` to the `ButtonProps` interface and remove it from `defaultProps`.  Do a check for `!type && (!as || as === 'button')`.  If the check passes, then we pass `type="button"` to `StyledButton`.

I think I prefer the latter, as that doesn't force users to add `type="button"` all over the place, but it is a little gross.  Open to ideas to handle that more smoothly.

modules/cactus-web/src/Button/Button.tsx
```diff
@@ -231,15 +241,23 @@ export const Button = styled(ButtonBase)<ButtonProps>`
   ${variantOrDisabled}
 `
 
-Button.propTypes = {
+type ButtonType = typeof ButtonFunc
+const ButtonFR = React.forwardRef(ButtonFunc) as any
```
mikesoltow, line 245: Because you have the `// @ts-ignore` on line 256, I think you can remove the `as any` here.  I'd rather have the type safety on this component than on the `as` prop type.

mikesoltow (changes requested):
Functionality looks good, but I just realized that we're going to need a way to handle the `type` attribute when the user doesn't pass anything into `as`.

modules/cactus-web/src/Button/Button.tsx
wilysword: It's not a meaningful attribute, but my browser didn't complain about it, so I'm not sure this is a problem worth fixing.

mikesoltow: It does matter in that it defaults to `type="submit"`.  I don't know how likely it is, but it's possible that some users could have non-typed buttons inside forms that they don't want to trigger the submit event.

wilysword: I think you may have misunderstood me; I mean that having `type="button"` on other elements doesn't seem like a problem to me. Although it _would_ probably be good to add it to `ButtonProps`, just so if someone wrote a custom component they'd know they need to handle it.

mikesoltow: Ahhhh...indeed I did. Yeah, that's fine with me.  If it does cause problems, we can revisit it.

mikesoltow (approved):
Change looks good.  Thanks!

PR #376, opened 2020-10-30
CACTUS-393 More Footer/Layout Changes

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-393

The first commit fixes a layout/styled-components bug I noticed while working on the footer; I included it here because I originally thought I'd need it to implement the footer changes (though that turned out not to be the case). The second commit is the fixes specific to the ticket.

One thing I noticed when testing in IE: when I changed the viewport it would look like the styles were wrong, but then I opened up the console and unchecked then re-checked a style and the page would reflow correctly. Not sure if that's a BrowserStack issue, and IE issue, or something else.


modules/cactus-web/src/Layout/Layout.tsx
```diff
@@ -105,51 +98,53 @@ const DefaultLayout = Layout as any
...
-  (p) => `
+const wrapperStyle = (p: LayoutProps) => css<LayoutProps>`
   overflow: auto;
   position: absolute;
   left: ${p.fixedLeft}px;
   right: 0;
   top: 0;
   bottom: ${p.fixedBottom}px;
-  display: flex;
-  flex-direction: column;
```
mikesoltow, line 117: By removing the flex styling here, there's a regression from Nicolas's change to ensure that the footer is rendered at the bottom of the page if there's not enough content.

Screen Shot 2020-11-02 at 1 29 47 PM

mikesoltow (changes requested):
Looks pretty good for the most part.  Just a couple of minor things, plus a humble request to help my brain that can't read code.

modules/cactus-web/src/Layout/Layout.tsx
mikesoltow: Random thought: it might be worth having a story that generates a screenshot similar to this so that we have a test case, but I'll leave that up to you.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -128,85 +137,100 @@ type FooterType = React.FC<FooterProps> & { Link: React.FC<LinkProps> }
...
+    [key, setLinks]
+  )
+  return null
 }
 
-const StyledFooter = styled.footer.attrs({ role: 'contentinfo' as string })`
+const StyledFooter = styled.footer.attrs({ role: 'contentinfo' as string })<{ isGrid: boolean }>`
+  ${(p) => textStyle(p.theme, 'small')};
+  word-break: break-all;
+  overflow-wrap: anywhere;
```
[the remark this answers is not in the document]
wilysword: `break-word` doesn't work for me in Firefox (for some reason, which is odd because I'm fairly certain I've seen it work in other places). I will change from `word-break` to `word-wrap`, though.

modules/cactus-web/src/Layout/Layout.tsx
wilysword: You're right, I must not have tested it again after removing this. However, when I have long content and `display: flex;`, I get this:
footer-overlap

Does that happen to you as well on IE? If so, do you have any idea how to fix it (without breaking anything else)?

mikesoltow: Ugggh, of _course_ it does that.  I'll tinker with this for a bit and see if I can come up with anything.

mikesoltow: I can't seem to tell if that's an issue with viewport resizing or something legit.  Every time I refresh after resizing the screen, I get an "unspecified error" from some Storybook file.

In any case, one thing I _did_ have to do was add `-ms-grid-row-align: center;` to `gridCell2` in `Footer.tsx`, as the alignment was off.

mikesoltow: If it is legit, I just wonder if IE isn't dynamically updating the footer position for some reason (is it keeping the footer in the same place vertically as it was in the last viewport?).  I have to run now, but I can look at this more tomorrow if you'd like.

wilysword: I think I got it working using grid instead of flex.

mikesoltow (changes requested):
Changes looked great...until I looked at Safari 😭   The short content story looked like the following screenshot:

Screen Shot 2020-11-03 at 4 57 26 PM

It appears that Safari needs the parent div (i.e. `wrapperStyle`) to have a defined height set.  Why, I'm not sure exactly, but this SO answer is what led me down this road: https://stackoverflow.com/a/44773109.  You can add `height: 100vh` to `wrapperStyle` unless you're on a tiny screen, in which case you have to subtract `60px` for the bottom nav bar.  I didn't _thoroughly_ test everything after adding that, but I think it looked good.

Sorry that browsers suck 😞

mikesoltow (approved):
Changes look good. Thanks!

Dhalton (approved):
Sweet! Looks great to me 💯

PR #380, opened 2020-11-03
fix(cactus-web): fix tooltip z index problem with select

NicolasSimmonds:
I don't know if this is the best way to solvew the issue, but it does the job (I think).

TO test:
Go to the select field on storybook and check if the problme has been fixed.

[CACTUS-388]

[CACTUS-388]: https://repayonline.atlassian.net/browse/CACTUS-388

### DIFF
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.tsx
@@ -19,7 +19,6 @@ interface AccessibleProps {
   status?: Status
   statusMessage?: React.ReactNode
   disabled?: boolean
-  isOpen?: boolean
 }
 
 type RenderFunc = (props: AccessibleProps) => JSX.Element | JSX.Element[]
@@ -42,7 +41,6 @@ interface AccessibleFieldProps extends FieldProps, MarginProps, WidthProps {
   className?: string
   children: JSX.Element | RenderFunc
   disabled?: boolean
-  isOpen?: boolean
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
@@ -55,7 +53,6 @@ export function useAccessibleField({
   warning,
   success,
   disabled,
-  isOpen,
 }: Partial<AccessibleFieldProps>): AccessibleProps {
   const fieldId = useId(id, name)
   const labelId = `${fieldId}-label`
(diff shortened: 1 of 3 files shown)


wilysword (approved):

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -52,6 +55,7 @@ export function useAccessibleField({
   warning,
   success,
   disabled,
+  isOpen,
```
mikesoltow, line 58: It looks like this is just a pure pass-through of `isOpen`.  `useAccessibleField` isn't actually doing anything to it.  I'd get rid of this (both in the hook and in `AccessibleProps`) and then get `isOpen` from `props` instead of `accessibility` in `AccessibleFieldBase`

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -52,6 +52,7 @@ export interface SelectProps
   canCreateOption?: boolean
   matchNotFoundText?: string
   comboBoxSearchLabel?: string
+  getSelectOpen?: (prop: boolean) => void
```
mikesoltow, line 55: I would rename this function, as it's more of an event handler than a getter.  Nothing simple springs to mind, but a couple of options:
* onDropdownToggle
* onMenuToggle
* onToggle
* onDropdownChange
* onMenuChange

modules/cactus-web/src/SelectField/SelectField.tsx
```diff
@@ -61,6 +62,7 @@ const SelectFieldBase: React.FC<SelectFieldProps> = (props): React.ReactElement
           id={fieldId}
           aria-labelledby={labelId}
           aria-describedby={ariaDescribedBy}
+          getSelectOpen={(e) => setIsOpen(e)}
```
mikesoltow, line 65: I wouldn't use `e` here, as that implies an event.  Maybe something like `newOpenState`.

mikesoltow (commented):
Looks pretty good to me.  Just a couple of minor suggestions.

mikesoltow (commented):
Thanks for the changes!  Just one more tiny request: can you destructure `onDropdownToggle` on line 1455 of `Select.tsx` so that it's not included in `rest`?  `...rest` is passed into the `SelectTrigger` component, which eventually makes its way down to the raw `button` element, and that generates a console error since it's an unrecognized event.

mikesoltow (approved):
Thanks for those changes!

PR #382, opened 2020-11-03
CACTUS-410 :: Add Lerna to Cactus

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-410

Not really much to test here.  You can play around with `yarn release` if you'd like...just make sure to exit before you actually publish anything 😄

### DIFF
--- a/examples/mock-ebpp/package.json
+++ b/examples/mock-ebpp/package.json
@@ -6,7 +6,7 @@
   "repository": "https://github.com/repaygithub/cactus",
   "author": "Victoria Vyverman",
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


wilysword (approved):
Looks good to the ignorant. Do we still need `scripts/release.js` for anything, or could you delete it?

mikesoltow:
@wilysword  Probably delete it.  I was thinking of keeping it around for a bit while I make sure this works, but since it will live on in Git history, I can delete it now if you'd rather be done with it.

NicolasSimmonds (approved):

PR #385, opened 2020-11-04
CACTUS-397 :: DataGrid Customization

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-397

Currently, I don't have this set up to allow developers to control flex properties on `TopSection` or `BottomSection`, because I'd be worried that that could open the door for issues with device responsiveness. Do you guys think we should let them change things like `justify-content` for instance?

Note: I moved `<ScreenSizeProvider>` into the storybook cactus addon so that it's always rendered in storybooks. For this reason, I removed usage of `ScreenSizeProvider` in several other stories.

### Testing
1. Open the DataGrid basic usage story
2. Play with the different features like pagination, changing the page size, and sorting to make sure they still work.
3. Change the screen size and the `cardBreakpoint` and make sure everything still looks good and functions properly at different screens sizes.
4. Feel free to modify the story to move some components around and see if you think it should satisfy developer needs; using margin props should help with organizing component placement in the top/bottom flex components.


wilysword (approved):
DataGrid is soo complicated, but I can't really think of a good way to simplify it...

modules/cactus-web/src/DataGrid/Table.tsx
```diff
@@ -0,0 +1,169 @@
...
+      const { sortAscending: currentSortAscending } = sortOptions[0] || {}
+      const newOptions = [{ id, sortAscending: exists ? !currentSortAscending : false }]
+      onSort(newOptions)
+    }
+  }
+
+  return (
+    <>
+      {children}
+      <Table fullWidth={fullWidth} cardBreakpoint={cardBreakpoint}>
```
mikesoltow, line 41: I think we need to add `dividers` back in.

modules/cactus-web/src/DataGrid/TopSection.tsx
```diff
@@ -0,0 +1,160 @@
...
+      $cardBreakpoint={cardBreakpoint}
+    >
+      {isCardView && sortableColumns.size > 0 && (
+        <div className={`sort-buttons ${hasChildren && 'has-siblings'}`}>
+          <MenuButton variant="unfilled" label={sortLabels.sortBy || 'Sort by'} mr={4}>
+            {[...sortableColumns.keys()].map(
+              (key): ReactElement => {
+                const col = sortableColumns.get(key) as DataColumnObject
+                return (
+                  <MenuButton.Item onSelect={() => handleSortColChange(key)}>
```
mikesoltow, line 79: This may have been a pre-existing error, but it looks like these need keys.

mikesoltow (changes requested):
Everything looks pretty good, but just a couple of things.

Also, I have a couple of thoughts on styling, but these are more questions than requests:
* Should we be adding some margins to the children of top/bottom section?  They can always override them, but it would be nice to handle the responsive margins for them (i.e. your example in the docs is a little complex to ask of them in terms of having conditional `mb` and `ml` props...not that it's hard, but I think it's something we can handle for them).
* On that note, I noticed that the `DataGrid.PageSizeSelect` component in the story is handling margins manually.  Should we be setting `justify-content: space-between;` to achieve that instead of having them set margins manually?

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -197,194 +78,30 @@ const DataGridBase = (props: DataGridProps): ReactElement => {
...
-                    direction === 'prev'
-                      ? paginationOptions.currentPage - 1
-                      : paginationOptions.currentPage + 1,
-                })
-              }}
-              disableNext={
-                prevNextProps
-                  ? prevNextProps.disableNext || data.length < paginationOptions.pageSize
-                  : data.length < paginationOptions.pageSize
-              }
```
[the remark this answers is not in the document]
Dhalton: Yeah I'm not sure what I was thinking removing this. I'll add it back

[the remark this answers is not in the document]
Dhalton: After looking at this again, I _do_ know what I was thinking when I removed this...right now I have this set up so that `data` is only passed to `DataGrid.Table`, so as it stands now, this component doesn't know anything about the length of the data. So would you rather me change the API such that `data` is passed to the top-level `DataGrid` element and passed down to sub-components via context? That's the only thing I can think of off the top of my head right now

Dhalton:
> * Should we be adding some margins to the children of top/bottom section?  They can always override them, but it would be nice to handle the responsive margins for them (i.e. your example in the docs is a little complex to ask of them in terms of having conditional `mb` and `ml` props...not that it's hard, but I think it's something we can handle for them).

I like the idea but I just wonder how we could try to predict where they want the margins. Take the pagination portion in the storybook for instance: if we were to automatically put a `ml="auto"` on the pagination element, how do we know they _want_ to render it on the right hand side? Sure, they could override it, but I feel like it may be confusing to developers why an element is rendering on the right-hand side when they'd probably expect it to render on the left. I guess my stance is that, if developers want more control over where things are rendered, we should give them that control and stop trying to do it for them.

> * On that note, I noticed that the `DataGrid.PageSizeSelect` component in the story is handling margins manually.  Should we be setting `justify-content: space-between;` to achieve that instead of having them set margins manually?

I can totally get behind this. I probably should've just done that from the get-go. Should we let them configure `justify-content` or leave it set statically (for table view at least)?

mikesoltow:
> I like the idea but I just wonder how we could try to predict where they want the margins. Take the pagination portion in the storybook for instance: if we were to automatically put a `ml="auto"` on the pagination element, how do we know they _want_ to render it on the right hand side? Sure, they could override it, but I feel like it may be confusing to developers why an element is rendering on the right-hand side when they'd probably expect it to render on the left. I guess my stance is that, if developers want more control over where things are rendered, we should give them that control and stop trying to do it for them. That being said, I could definitely see how we could manage the bottom/top margins for them to ensure they're spaced out in card view, at least. Feel fee to push back if you feel strongly about this -- maybe I'm just not seeing something that you are.

Point taken.  There's definitely a balance to strike here (as with the accordion drag-and-drop discussion we had).  My concern is more for the spacing between adjacent components as opposed to the alignment.  For example, if you want 5px of spacing, you have to know how we rearrange the components in order to have the margin properly switch between `ml` and `mb`.  Because of that, I think that's something we should handle for them.  And we could do that by either configuring it for them or adding a prop to `TopSection` called `spacing` or something.

For alignment, maybe we allow them to configure `justifyContent` (and default it to `space-between`.  And that would answer your second question.

What do you think?  Again, I'm don't feel _strongly_ about this, as I don't want to add too much complexity.  But I also don't want to abstract away responsiveness in some cases but not others.

Dhalton:
> My concern is more for the spacing between adjacent components as opposed to the alignment. For example, if you want 5px of spacing, you have to know how we rearrange the components in order to have the margin properly switch between `ml` and `mb`. Because of that, I think that's something we should handle for them. And we could do that by either configuring it for them or adding a prop to `TopSection` called `spacing` or something.

I think I'm starting to understand a little bit better what use-case you want to solve for here. The distinction between how components are laid out and how adjacent component spacing should work is something I hadn't considered. I like the idea of passing a `spacing` prop to the top/bottom sections, so I'll do that as well as make `justify-content` configurable.

modules/cactus-web/src/DataGrid/DataGrid.tsx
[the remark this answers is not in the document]
mikesoltow: Nah, don't bother.  Now that I think about it, that should be up to the devs to manage.  Much like other options with this component, the behavior could be dependent on the API or something.

modules/cactus-web/src/DataGrid/BottomSection.tsx
```diff
@@ -1,25 +1,50 @@
...
+      $cardBreakpoint={cardBreakpoint}
+      $justifyContent={justifyContent}
+    >
+      {cloneAll(
+        children,
+        {
+          style: isTinyScreen ? { marginTop: margin } : { marginLeft: margin },
+        },
+        (element: React.ReactElement, props: any): React.ReactElement => {
+          if (element.key === '.0') {
```
mikesoltow, line 37: This doesn't look like it will work if the dev passes in a key to the first element for any reason.  One idea might be to use the index instead.  The `childArray.reduce` callback on line 13 of `react.ts` should take a `currentIndex` as a third argument.  You could then have `makeClone` take it as a third argument.

modules/cactus-web/src/DataGrid/TopSection.tsx
```diff
@@ -93,12 +103,23 @@ const TopSection = (props: TopSectionProps): ReactElement | null => {
...
-      {children}
+      {cloneAll(
+        children,
+        {
+          style: isTinyScreen ? { marginTop: margin } : { marginLeft: margin },
+        },
+        (element: React.ReactElement, props: any): React.ReactElement => {
+          if (!isTinyScreen && element.key === '.0') {
+            return element
+          }
```
mikesoltow, line 114: In addition to the comment I made in `BottomSection`, it would probably be a good idea to test for card view.  We should keep the margin in card view mode for `TopSection` since the sort buttons will appear on the same line if we're not on a tiny screen.

mikesoltow (commented):
I like these changes a lot!  Just a couple minor things, but then I think we're good.

modules/cactus-web/src/DataGrid/BottomSection.tsx
Dhalton: Can't believe I didn't think about that 🤦 your idea to pass the index through `makeClone` should work well though, thanks!

modules/cactus-web/src/DataGrid/TopSection.tsx
Dhalton: Nice catch!!

mikesoltow (approved):
Changes look great.  Thanks!

wilysword (approved):

PR #386, opened 2020-11-05
Merging Toggle Changes into Cactus Web v4

mikesoltow:
The main thing to test is the Toggle story.  Make sure that the toggle itself looks good and that the ACs of https://repayonline.atlassian.net/browse/CACTUS-399 are met.

### DIFF
--- a/modules/cactus-web/src/Toggle/Toggle.tsx
+++ b/modules/cactus-web/src/Toggle/Toggle.tsx
@@ -115,7 +115,7 @@ const Switch = styled.div`
     }
   }
 
-  input:checked:not([disabled]) ~ & {
+  input:checked:not(:disabled) ~ & {
     background-color: ${(p): string => p.theme.colors.success};
   }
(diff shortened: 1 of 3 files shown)


modules/cactus-web/src/Toggle/Toggle.tsx
```diff
@@ -111,12 +110,14 @@ const Switch = styled.div`
...
 
   input:checked ~ & {
-    background-color: ${(p): string => p.theme.colors.success};
-
     ::after {
       left: 26px;
     }
   }
+
+  input:checked:not([disabled]) ~ & {
```
wilysword, line 118: I think it'd be a bit more consistent to use the pseudo-class:
```suggestion
  input:checked:not(:disabled) ~ & {
```

wilysword (approved):
Looks good to me.

NicolasSimmonds (approved):

PR #387, opened 2020-11-05
CACTUS-406 BrandBar Contents

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-406

I mostly did just what it says in the ticket, but I did make it so that moving the `BrandBar.Item`s to the ActionBar on mobile screens is optional, instead of required. That way the devs can choose if they want their item to move to the ActionBar, or stay at the top of the page.

Also `userMenuText` on a component called `UserMenu` seemed redundant, so I renamed it to `label`. I'm not completely happy with that choice, so if anyone can think of a better name I'm open to suggestions.

Besides the changes for the ticket, I also racked up more minor fixes than I probably should have:
- Fixed a styling issue with empty ActionBar
- Fixed an error in the ActionBar docs
- Removed default prop on ToggleField that was causing PropTypes warning
- Added `useScreenSize` hook


Dhalton (approved):
Nice, seems good to me. I'll be glad to have a `useScreenSize` hook now 😄

mikesoltow (approved):
Looks great!  i can't really come up with much for the `label` prop name either.  Maybe `title`?  But maybe that's not better 🤷

wilysword:
I like `title` about as much (or as little) as `label`, but `label` seems more consistent with/similar to other components so I guess I'll leave it.

PR #388, opened 2020-11-06
Csf stories migration

NicolasSimmonds:
There are a lot of changes in this PR.

Most of them are on each component story, so the best way to review those would be in the storybook itself and see that nothing broke. Do this for both cactus-web and the single story in cactus-icons.

I also changed the configuration files in both modules too, ".storybook/**". I did this because of the way the docs say it needs to be configured. I also updated some things so it would be easier to migrate to other releases in the future.

Lastly, check the make-component function. I changed that to reflect the migration to CSF. A good way to test this is generating a new component.

[CACTUS-402]

[CACTUS-402]: https://repayonline.atlassian.net/browse/CACTUS-402

### DIFF
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.story.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.story.tsx
@@ -1,26 +1,28 @@
 import { boolean, text } from '@storybook/addon-knobs'
-import { storiesOf } from '@storybook/react'
+import { Meta } from '@storybook/react/types-6-0'
 import React from 'react'
 
 import AccessibleField from './AccessibleField'
 
-storiesOf('AccessibleField', module).add(
-  'Basic Usage',
-  (): React.ReactElement => {
-    const disabled = boolean('disabled', false)
-    return (
-      <AccessibleField
-        disabled={disabled}
-        name={text('name', 'field_name')}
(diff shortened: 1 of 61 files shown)


Dhalton (approved):
Nice! LGTM 👍

modules/cactus-web/src/AccessibleField/AccessibleField.story.tsx
```diff
@@ -1,26 +1,28 @@
...
-      >
-        <input style={{ minWidth: '300px' }} />
-      </AccessibleField>
-    )
-  }
-)
+export default {
+  title: 'AccessibleField',
+  component: AccessibleField,
+} as Meta
```
wilysword, line 10: What's the reason for this `as Meta`? `test:types` passes for me event when I remove it; was it added automatically?

wilysword (approved):
Just one question, for the sake of curiosity.

modules/cactus-web/src/AccessibleField/AccessibleField.story.tsx
NicolasSimmonds: I was having some problems with the linting, and after a quick search for the docs this was the way to do it with TS. 
The reason as I understood was that the default export was meta information for storybook, and TS likes to make that explicit.

PR #389, opened 2020-11-06
CACTUS-405 :: IE relatedTarget Support

Dhalton:
https://repayonline.atlassian.net/browse/CACTUS-405

Should be a quick & easy one. I tested other usages of `relatedTarget` that I found, but I didn't notice any problems. If you happen to notice anything I missed, let me know!

### Testing
Test the following in both IE and your preferred browser
1. Open the Select "with comboBox" story and focus the combobox input. Click outside the select and make sure the list closes.
2. Open the Layout story and click the gear icon in the action bar. Click outside the panel to make sure it closes when focus leaves the panel.

### DIFF
--- a/modules/cactus-web/src/Select/Select.tsx
+++ b/modules/cactus-web/src/Select/Select.tsx
@@ -1158,8 +1158,9 @@ class SelectBase extends React.Component<SelectProps, SelectState> {
       return
     }
 
-    // In IE, there's no relatedTarget on blur...it's already moved on to activeElement
-    if (isIE()) {
+    // In IE, there's no relatedTarget on blur with React versions lower than v17...it's already moved on to activeElement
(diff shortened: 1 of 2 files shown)


mikesoltow (approved):
LGTM

wilysword (approved):

Dhalton:
@wilysword @mikesoltow You guys may want to re-review this before it gets merged; because the logic in `usePopup` changed a bit, I don't think there's a need to explicitly check the React version anymore. `focused` defaults to `event.relatedTarget` anyway. I did keep the change to the wording of the comment, though.

wilysword:
I'm fine with it, given that I'm the one who wrote it that way to begin with. (And in the feature branch that bit of code was moved to a different file anyway...yay merge conflict waiting to happen!)

mikesoltow:
I'm fine with this too!

PR #390, opened 2020-11-09
CACTUS-411 Apply Global Text Styles to button, input, & textarea

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-411

Most of the elements already had local overrides for text size/style, so I think most of the storyshot changes are font, like a change from Arial to Helvetica (not that I can tell the difference...).


mikesoltow (approved):
LGTM

wilysword:
I had to remove one of the styles from `button` to get the tests to pass, no idea why. I plan on trying to re-add them after the `portal=false` ticket to see if there's some kind of weird interaction there, though I'm not too hopeful. Note that as far as I know, every button that has any text overrides the text styles locally, so this shouldn't have much impact visually.

mikesoltow (approved):
Thanks for looking into this!  I'm more than fine with the workaround, as applying text styles to button doesn't affect the Select font-size anyway, as the `span` within the button has a font size set.  All it does it jiggle it a bit because of the line height, but I don't care too much.

Dhalton (approved):

PR #393, opened 2020-11-09
CACTUS-414 :: Tooltip Hover

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-414

### Testing
Open the Tooltip story and hover the icon to bring up the tooltip text. Move your mouse inside the tooltip popup and verify that it stays open as long as your mouse stays inside the content box.

### DIFF
--- a/modules/cactus-web/src/Tooltip/Tooltip.test.tsx
+++ b/modules/cactus-web/src/Tooltip/Tooltip.test.tsx
@@ -54,4 +54,25 @@ describe('component: Tooltip', (): void => {
     expect(portal).not.toBeNull()
     expect(portal.firstChild).toMatchSnapshot()
   })
+
+  test('should continue to render tooltip when content is hovered', async (): Promise<void> => {
+    jest.useFakeTimers()
+    render(
+      <StyleProvider>
+        <Tooltip label="This should be displayed" />
+      </StyleProvider>
+    )
+    act((): void => {
+      fireEvent.mouseEnter(document.querySelector('span') as Element)
+      setTimeout(jest.fn(), 2000)
+      jest.runAllTimers()
+    })
+    act((): void => {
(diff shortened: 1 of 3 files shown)


NicolasSimmonds (approved):
LGTM

modules/cactus-web/src/Tooltip/Tooltip.test.tsx
```diff
@@ -54,4 +54,25 @@ describe('component: Tooltip', (): void => {
...
+      fireEvent.mouseEnter(document.querySelector('span') as Element)
+      setTimeout(jest.fn(), 2000)
+      jest.runAllTimers()
+    })
+    act((): void => {
+      fireEvent.mouseEnter(document.querySelector('div[role="tooltip"]') as Element)
+      setTimeout(jest.fn(), 2000)
+      jest.runAllTimers()
+    })
+    const tooltip = document.querySelector('div[role="tooltip"]') as Element
```
mikesoltow, line 75: I think there's a `queryByRole` that you might want to use instead 🤷

mikesoltow (approved):
Nice 👍 

One cleanup suggestion on the test, but it's NBD.

modules/cactus-web/src/Tooltip/Tooltip.test.tsx
Dhalton: I would, but there are two elements with the `tooltip` role. There's a div, which is what we want, and there's a hidden span for accessibility. AFAIK there isn't a way to specify the element with `queryByRole`, so I think, in this case, using `querySelector` is a little easier. If you know of a way I could make the query more granular so that it only picks up the one we want, I'll definitely swap this out for `queryByRole`

mikesoltow: 🤦 
Duh

I think there is a way to pass a selector, but it's no cleaner than what you have at that point.  Let it ride.

PR #395, opened 2020-11-10
Revert "Go back to running BrowserStack tests in parallel"

mikesoltow:
😢

### DIFF
--- a/examples/mock-ebpp/run-ete-tests.js
+++ b/examples/mock-ebpp/run-ete-tests.js
@@ -19,7 +19,7 @@ program.requiredOption(
 )
 program.parse()
 
-const runTests = async () => {
+const runIndividualTest = async (browser) => {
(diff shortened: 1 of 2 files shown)


NicolasSimmonds (approved):
😢

Dhalton:
Did this fail because Glen's PR is trying to run integration tests maybe?

mikesoltow:
That would be my guess.  It's still not foolproof, but it cuts in half the resources our integration tests will take up.

Dhalton:
Yeah. Glen's has been stuck on integration tests for almost 2 hours so maybe we should cancel that run...

mikesoltow:
> Yeah. Glen's has been stuck on integration tests for almost 2 hours so maybe we should cancel that run...

Done

Dhalton (approved):

PR #397, opened 2020-11-11
CACTUS-412 MenuBar scroll buttons are both visible or both hidden

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-412

I looked into using IconButton like the ticket said, but it didn't work out:
- The icons are 18px, which is not an available `IconSize`. Looking at the code I think you could actually pass `iconSize="18px"` and it would work, but you'd have to do `@ts-ignore` _and_ you'd still get a PropTypes warning.
- (I had more reasons, but when I considered them more deeply while writing them out here, I realized they weren't as good as I thought they were.)

### DIFF
--- a/modules/cactus-web/src/MenuBar/MenuBar.tsx
+++ b/modules/cactus-web/src/MenuBar/MenuBar.tsx
@@ -153,13 +153,13 @@ const FloatingMenu: React.FC<React.HTMLAttributes<HTMLElement>> = ({
   const [menuRef, scroll] = useScrollButtons('vertical', !hidden)
   return (
     <MenuWrapper aria-hidden={hidden} tabIndex={tabIndex} onKeyDown={onKeyDown}>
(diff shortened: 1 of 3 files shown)


Dhalton (approved):
LGTM!

mikesoltow (approved):
LGTM

Completely unrelatedly, I can't remember if I asked you about the fact that, if you click on a menu item that has a dropdown, there's a quick flash of an outline before the dropdown appears.  I wonder if we need a ticket to look at that?

wilysword:
Are you talking about the outline on the button that you clicked on? That's because when you click the button it gets focus, and then focus shifts to the submenu so that you can use the arrow keys to navigate.

mikesoltow:
> Are you talking about the outline on the button that you clicked on? That's because when you click the button it gets focus, and then focus shifts to the submenu so that you can use the arrow keys to navigate.

Yeah.  Makes sense...it just looks a bit weird, so it might be worth looking into.

PR #398, opened 2020-11-16
CACTUS-407 :: Controlled Accordions

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-407

I refactored some of the tests, which I put into a separate commit.  If you want to filter those changes out of the diff, just look at the second commit.

Originally the ACs stated that a warning should be logged to the console if a non-existent ID was passed into `openId`.  However, that seems to be more complex than initially thought, as the Provider is not initially aware of all the descendant Accordions on first render (the Accordions have to register themselves).  I'm sure there's some logic we could add to have it wait until subsequent renders to log the warning, but that seemed like more complexity than it was worth.  Feel free to push back though!

### Testing

1. First verify that all the existing Accordion stories work as expected. Note that the "With Outline" story has `maxOpen` set to `2`, so if you try to open more than 2, you should see Accordions start to close.  Verify that that happens in the correct order.
2. An `onChange` prop was added to the `Accordion.Provider` for both the controlled and uncontrolled variants.  The stories "With Open Initialization," "With Outline," and "With Controlled Outline" will all log to the console when that event is called.  Ensure that that event fires when you'd expect.  For example, it should _not_ fire on first render when one or more Accordions default to open.  It _should_ fire if an Accordion is closed because `maxOpen` was reached, etc.
3. Make sure that you can still use the down arrows to navigate between Accordions in both the "With Outline" and "With Controlled Outline" stories.
4. Feel free to mess around in the story code to play around with the props as you see fit.  For example, you can trigger warnings by passing `maxOpen` to the provider or `defaultOpen` to an Accordion in the "With Controlled Outline" story.


Dhalton (approved):
Looks great to me! 🚀

NicolasSimmonds (approved):

PR #399, opened 2020-11-16
CACTUS-409 Portal Removal

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-409

It should work almost exactly the same. The major difference is in the "many options" story for the Select component: before it would overlap the drop-down button if it couldn't fit on the screen; currently I have it just stay below. There's a max-height of 400px on the drop-down so unless there's a rather odd set of display conditions, I don't think there would be many situations where you couldn't scroll either up or down to get the whole thing in view. All the same, I'm open to suggestions on how/if we should handle that situation differently.

EDIT: I added the Action Label knob to the collision stories so you can check horizontal collisions as well as vertical ones.

PS Naming the branch as I did amused me more than it probably should have.


modules/cactus-web/src/helpers/positionPopover.ts
```diff
@@ -17,3 +18,107 @@ export function getTopPosition(targetRect: PRect, popoverRect: PRect): { top: st
...
+
+  const expectedBottom = bottom + ddHeight
+  const expectedTop = top - ddHeight
+  if (expectedBottom > maxBottom && expectedTop > 0) {
+    dd.style.top = ''
+    dd.style.bottom = `${maxBottom + MARGIN - top}px`
+  } else {
+    dd.style.bottom = ''
+    dd.style.top = `${bottom + MARGIN}px`
+  }
```
mikesoltow, line 111: What if we deleted the `max-height` defined in `Select.tsx` and changed this section to:

```suggestion
  if (expectedBottom > maxBottom && expectedTop > 0) {
    const styleBottom = maxBottom + MARGIN - top
    dd.style.top = ''
    dd.style.bottom = `${styleBottom}px`
    dd.style.maxHeight = `${view.clientHeight - styleBottom - 10}px`
  } else {
    const styleTop = bottom + MARGIN
    dd.style.bottom = ''
    dd.style.top = `${styleTop}px`
    dd.style.maxHeight = `${view.clientHeight - styleTop - 10}px`
  }
```

mikesoltow (commented):
Looking pretty good for the most part, although I still need to test on non-Chrome browsers.

Before I did that, however, I wanted to leave a suggestion for the long list of options story.  I was able to create several cases where I couldn't scroll all the way down, so I think it's definitely possible that we'd run into issues.  What if we dynamically changed maxHeight to the remaining space in the viewport below the anchor?  That way it never(? -- Not sure if fixed footers, etc. would cause issues) flow past the bottom of the screen.  I left a rough idea of potential code in the comments.

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -860,108 +825,92 @@ class List extends React.Component<ListProps, ListState> {
...
     const { options, activeDescendant: listManagedActiveDescendant } = this.state
 
     const activeDescendant = comboBox ? selectManagedActiveDescendant : listManagedActiveDescendant
+    const setListRef = (n: HTMLUListElement | null): void => {
+      this.listRef = n
+    }
 
     return (
-      <Portal>
-        <Rect observe={isOpen}>
```
mikesoltow, line 874: Looks like there is a small regression here.  If you have a dropdown open on a multi-select and start checking items, then the anchor element changes size, but the dropdown menu no longer changes size with it.  Looks like that was triggered by this `Rect` component that observed the anchor element (when the list was open).  If we don't want to use the `Rect` component, then we may want to make use of `@reach/observe-rect`, which `Rect` uses under the hood (see https://github.com/reach/reach-ui/blob/develop/packages/rect/src/index.tsx#L137).

mikesoltow (changes requested):
Actually I did find one regression we should probably look at.

modules/cactus-web/src/Select/Select.tsx
wilysword: To be honest, I'm not sure we should be allowing the select to change size like that in the first place. This should be doable, though.

mikesoltow: I assume you're talking about the entire Select and not just the dropdown?  That could be something we look into in a tech debt ticket if you'd like.

wilysword: In general I think it's best to avoid changing an element's size, but I'm also not sure how we could keep it the same size while still sizing correctly for dynamic content (e.g. translations). This particular element is probably an exception to that general rule.

wilysword: While I talked myself out of my initial concern with the anchor element resizing, I'm still not sure I want to change this back to how it was: it's one thing for an element to resize when your focus is somewhere else, but this would change the width of the container the user is currently clicking in. Try, for instance, opening the drop-down and unchecking elements by clicking near the checkbox on the left side: as the anchor shrinks, you have to keep moving the mouse over and if you were going quickly, would quite likely accidentally close the drop-down. In that respect, I think the new behavior is better than the old behavior.

mikesoltow: Fair enough.  After playing around with this some more, I'm fine the way you have it.

mikesoltow (commented):
Thanks for the updates regarding the dropdown height for long lists.

I did notice another bug: if you have a mult-select element that takes up the full viewport but then uncheck an item or two, the drop-down will not be correctly aligned with the Select element when first opened.

Screen Shot 2020-11-23 at 6 04 45 PM

It looks like in this case it's hitting the `  if (expectedRight > maxRight) {` of the `positionDropDown` function in `positionPopover.ts`.  I need to run, unfortunately, and can't dig into it anymore tonight, but I'm wondering if the `ddWidth` isn't what we expect the first time it opens.  In the screenshot of my console below, I logged the values of several variables.  Under "uh oh" are the values for the first toggle, when the list is not properly aligned.  I then close the list and re-open it, and this time it hits the `else` block and logs "all good".  You can see that `ddWidth` is smaller the second time around, which allows it to hit the `else` block.

Screen Shot 2020-11-23 at 6 05 02 PM

wilysword:
Yeah, that's because the width is calculated based on the _current_ min/max values, which would have been set when the anchor was wider. Once the min is lowered the auto-width is also lowered, but that doesn't happen till the calculations are already done, which is why I have the recursive call at the bottom. I guess it doesn't work as well for shrinking the drop-down as it does for growing it, though, so I'll have to see if I can figure out a way to do that.

NicolasSimmonds (approved):
Latest changes seem good to me!

Dhalton (approved):

mikesoltow (approved):
Thanks for the changes!  LGTM once the conflict is fixed.

PR #401, opened 2020-11-18
Cactus 420 make intermediate default shape in the theme

NicolasSimmonds:
Test:
-In storybook, when it loads check the default theme is Intermediate.
-Also check, that the onject returned by the theme generator function has the intermediate value for the shape.


Dhalton (approved):
LGTM! 👍

mikesoltow (commented):
This looks good, but it looks like not all of the snapshots updated.  That makes me wonder if our visual snapshot failure threshold is too high.  I tried removing it, and it updated a bunch more snapshots.  Most of the changes were related to your PR, but it looks like some of them were font related.  I know that you may have run tests outside of Docker in the past, Nicolas, so maybe that's how the current versions got in there.

See https://github.com/repaygithub/cactus/commit/9bacc13820629d3ec664a34e5bf1cfc6d5c9ebfc for a commit removing that threshold.

What do you think about making that change?

mikesoltow (approved):
Never mind, the build on my commit failed.  I think we'll need to keep some threshold around for now.  Once @Dhalton 's current ticket is done, we can better debug differences between CI and local dev and potentially lower the threshold.  But for now, I'm happy to just leave this as-is.

NicolasSimmonds (commented):
Everything looks good on my side, I also took a look into other features and they all work for me 👍

PR #402, opened 2020-11-19
typo confrimodal to confirmmodal

SkarDude:


modules/cactus-web/src/ConfirmModal/ConfirmModal.mdx
```diff
@@ -53,7 +53,7 @@ export const code = `() => {
 ```jsx
 import React from 'react'
 import { ConfirmModal, Text} from '@repay/cactus-web'
-const ConfriModalExample = () => {
+const ConfirmModalExample = () => {
 const [open, setOpen] = React.useState(false)
```
luisniebla, line 57: I copy/pasted this section of code and someone on my team pointed out `open` is a keyword that would get flagged by sonarqube. I know it's just example code but something to keep in mind 😄

luisniebla (approved):

NicolasSimmonds (approved):

modules/cactus-web/src/ConfirmModal/ConfirmModal.mdx
mikesoltow: Good call!  I've added a commit that will fix that.

NicolasSimmonds (approved):

mikesoltow (approved):
@SkarDude Thanks for fixing this!

PR #403, opened 2020-11-19
CACTUS-417: set accordion transition time to 200ms

NicolasSimmonds:
Test:
Play with the accordion in storybook and check that, regardless of the size, the transition time is the same
[CACTUS-417]

[CACTUS-417]: https://repayonline.atlassian.net/browse/CACTUS-417


Dhalton (approved):

mikesoltow (approved):
LGTM

PR #405, opened 2020-11-20
inherit for flexWrap options instead of initial

SkarDude:


mikesoltow (commented):
`initial` is a valid value for the component, even though it looks like we throw a propTypes warning for it.  My two cents would be to add `initial` to the propTypes and add `inherit` to the storybook options (but leave `initial` in there).

mikesoltow:
Also, @SkarDude if you ever end up doing a PR for code in `src`, we like to keep to a standard commit message.  We have a `yarn commit` command that will walk you through crafting a message so you don't have to worry about formatting.  See https://github.com/repaygithub/cactus/blob/master/CONTRIBUTING.md#git-commit-messages.

SkarDude:
@mikesoltow Ok cool, I can close this and I'll do a standard commit message on future commits. I tried initial and got a warning and inherit worked so I just assumed typo in your docs.

PR #408, opened 2020-11-23
CACTUS-425 Layout Bug

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-425

Obviously a ton of changed files. The vast majority of the screenshot changes should be because I changed the `default` storyshot config to be a desktop-like screen size (the smallest possible "large" breakpoint). To get that to work, I also had to change the MenuBar story to only randomize non-visible stuff.

Besides the storyshot changes, I fixed two bugs (the one that lead to this ticket, and a similar bug with the Footer) and pulled the common Layout story code into separate components, in case we want to add more default combinations in the future.


mikesoltow (commented):
This looks great!  While you're here, however, I did notice one more issue: the BrandBar is huge on tiny screens in IE11.  It seems that it can be fixed by moving `max-height: 80px;` up from `LogoWrapper` into `StyledBrandBar` inside `BrandBar.tsx`.   (Sorry, I know this was pre-existing...since it's a one-liner it just seemed quicker to do it here so we can get the release out).

wilysword:
Do we care about tiny screens in IE11? No actual mobile device is going to be using that browser, and not many desktop users would have their windows that small.

mikesoltow:
> Do we care about tiny screens in IE11? No actual mobile device is going to be using that browser, and not many desktop users would have their windows that small.

Practically, I would say that you're correct.  For this case, it just seems like such a low-hanging fruit that we might as well dot that i, but if you feel strongly I could approve anyway.

wilysword:
I wouldn't necessarily mind fixing it, but wouldn't the specific fix you mentioned interfere with adding other children to the BrandBar (like in the "Custom Items" story)? Not just IE, but several browsers seem to have odd display rules for img elements.

mikesoltow:
> I wouldn't necessarily mind fixing it, but wouldn't the specific fix you mentioned interfere with adding other children to the BrandBar (like in the "Custom Items" story)? Not just IE, but several browsers seem to have odd display rules for img elements.

Yeah, good point.  Okay, I'm happy to leave this in then.  If we need to revisit it, or if another solution occurs to us, we can fix it  later.

mikesoltow (approved):

Dhalton (approved):

PR #409, opened 2020-11-24
CACTUS-424 Remove DateInput Portal

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-424

Basically the same as the other portal removals, except I mostly kept the same positioning function it was using before; I couldn't use `positionDropDown` because the date picker's width is independent of the anchor's, which changes the calculations. Mostly I just removed the scroll references (because of the change from absolute to fixed positioning) and hard-coded the margin, but I did also fix a bug in how `collisions.right` was calculated.


Dhalton (approved):

mikesoltow (approved):
LGTM!

PR #411, opened 2020-11-25
Cactus 421 left justified tootlip

NicolasSimmonds:
Test: in the storybok, go to all the *Field components, the CheckBoxGroup and RadioButtonGroup. There is a alignTooltip knob. Toggle between left and right and check it works as expected.

 
[CACTUS-421]

[CACTUS-421]: https://repayonline.atlassian.net/browse/CACTUS-421


modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -169,19 +177,19 @@ export const AccessibleField = styled(AccessibleFieldBase)`
...
   ${width}
+  display: flex;
+  flex-direction: column;
 
   ${Label} {
     display: block;
     box-sizing: border-box;
     padding-left: 16px;
-    padding-right: 28px;
+    padding-right: ${(p) => (p.alignTooltip === 'right' ? '28px' : '5px')};
```
wilysword, line 187: I don't see any reason to have a different rule depending on the tooltip alignment; I'd suggest changing this to a flat 8px for both alignments (second opinions?).

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -169,19 +177,19 @@ export const AccessibleField = styled(AccessibleFieldBase)`
...
     color: ${(p) => p.disabled && p.theme.colors.mediumGray};
   }
 
   ${Tooltip} {
-    position: absolute;
-    right: 8px;
```
wilysword, line 183: I think we still need this spacing, try adding `padding-right: 8px`.

modules/cactus-web/src/CheckBoxGroup/CheckBoxGroup.tsx
```diff
@@ -125,20 +132,27 @@ export const CheckBoxGroup = React.forwardRef<HTMLFieldSetElement, CheckBoxGroup
...
         onBlur={handleBlur}
       >
-        <Label id={labelId} as="legend">
-          {label}
-        </Label>
+        <LabelWrapper
+          alignTooltip={alignTooltip}
+          justifyContent={alignTooltip === 'right' ? 'space-between' : 'flex-start'}
+        >
+          <Label id={labelId} as="legend">
```
wilysword, line 139: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend), the `legend` element must be the first child of the `fieldset`, so you can't have the intermediate flex div. You could probably get the same effect by putting the Tooltip inside the Label and changing the Label's display to flex.

modules/cactus-web/src/CheckBoxGroup/CheckBoxGroup.tsx
```diff
@@ -151,6 +165,25 @@ export const CheckBoxGroup = React.forwardRef<HTMLFieldSetElement, CheckBoxGroup
...
+    bottom: 0;
+  }
+  ${Label} {
+    border: 0;
+    width: auto;
+    display: block;
+    box-sizing: border-box;
+    padding-left: 16px;
+    padding-right: ${(p) => (p.alignTooltip === 'right' ? '28px' : '5px')};
+    padding-right: ${(p) => (p.alignTooltip === 'right' ? '28px' : '5px')};
```
wilysword, line 183: duplicate line

modules/cactus-web/src/RadioGroup/RadioGroup.tsx
```diff
@@ -136,20 +143,26 @@ export const RadioGroup = React.forwardRef<HTMLFieldSetElement, RadioGroupProps>
...
         onBlur={handleBlur}
       >
-        <Label as="legend" id={labelId}>
-          {label}
-        </Label>
+        <LabelWrapper
+          alignTooltip={alignTooltip}
+          justifyContent={alignTooltip === 'right' ? 'space-between' : 'flex-start'}
+        >
+          <Label id={labelId} as="legend">
```
wilysword, line 150: Same thing as with CheckBoxGroup.

wilysword (changes requested):
Functionality seems fine, but there's a nitpick with fieldset/legend that _could_ affect accessibility. I also have a suggestion on spacing, but I'm open to discussion on that one.

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
NicolasSimmonds: I agree with you

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
NicolasSimmonds: So you are saying leaving the `position: absolute` and `right: 8px`. and also adding the` padding-right: 8px`?

Because with the position absolute the problem is that is hard to align the tooltip exactly at the right of the label

wilysword: No, the things you removed here are fine; I meant that you need to add the `padding-right` _in place of_ the `right`. Accomplish the same thing visually, but in a different way.

Dhalton:
LGTM but it looks like there are a lot of conflicts so I'll wait to approve.

wilysword:
I think you had the wrong version of `cactus-theme` when you updated the snapshots/storyshots, they're showing round corners instead of intermediate.

wilysword (approved):

Dhalton (approved):

PR #413, opened 2020-11-27
docs(cactus-web): add clarification to actionbar and layout docs

NicolasSimmonds:
Check the docs page for Actionbar and Layout, and see if the docs are good now

[CACTUS-423]

[CACTUS-423]: https://repayonline.atlassian.net/browse/CACTUS-423


modules/cactus-web/src/ActionBar/ActionBar.mdx
```diff
@@ -94,6 +92,26 @@ const MyPage = () => (
...
+```jsx
+import React from 'react'
+import { ActionBar, Layout } from '@repay/cactus-web'
+import MyPage from './Mypage'
+
+const App = () =>(
+  <Layout>
+    // Other layout components
+    <ActionBar>
+      //Action bar options
```
wilysword, line 104:
```suggestion
      // Global action bar options
```

modules/cactus-web/src/ActionBar/ActionBar.mdx
```diff
@@ -94,6 +92,26 @@ const MyPage = () => (
...
+    </ActionBar>
+    <Layout.Content>
+      <MyPage />
+    <Layout.Content>
+  </Layout>
+)
```
wilysword, line 110: This is just what I had in mind for the example, but I think it would work better if you moved it up to be part of the code on lines 76-85, with comments separating the two "files": something like
```
// MyPage.tsx
import React from 'react'
Rest of the example, etc.
  </>
)

// App.tsx
the rest of your new example
```
That makes it clearer that the two examples are related.

modules/cactus-web/src/Layout/Layout.mdx
```diff
@@ -37,14 +41,20 @@ const MySite = () => (
...
+      </Router>
     </Layout.Content>
     <Footer logo={LOGO}>The Foot clan returns.</Footer>
   </Layout>
 )
 ```
 
+The `Router` inside `Layout.Content` is used to navigate between different pages and components.
+Note that `Home` and `Customer` are made-up components, just to show you the functionality.
+Also, we use `@reach/router` in this example, but you are free to use any routing library that you want.
```
wilysword, line 56: I don't think you need to specifically mention that the components are made-up. You should probably mention that depending on the library, the MenuBar may also need to be connected to the Router so that links will work; in fact, you should probably update the `MenuBar.Item`s in this example so they're linked to the sample content pages.

Also @mikesoltow I know bytemover isn't using `@reach/router`, is there a routing library we recommend?

wilysword (changes requested):

modules/cactus-web/src/ActionBar/ActionBar.mdx
```diff
@@ -78,13 +76,30 @@ const MyPage = () => (
...
     />
-    <PageContent/>
+    <PageContent />
   </>
 )
+import React from 'react'
+import { ActionBar, Layout } from '@repay/cactus-web'
+import MyPage from './Mypage'
+
+// App.tsx
```
wilysword, line 90: Move this up four lines, above the imports.

modules/cactus-web/src/Layout/Layout.mdx
```diff
@@ -25,26 +26,45 @@ All the children except `Layout.Content` are optional, though ActionBar is highl
...
+        <MenuBar.Item>
+          <Link to="/home">Teenage Mutant</Link>
+        </MenuBar.Item>
+        <MenuBar.Item>
+          <Link to="/other/path">Ninja Turtles</Link>
+        </MenuBar.Item>
+      </MenuBar>
+      <ActionBar />
+      <Layout.Content>
+        <Router>
```
wilysword, line 48: I think the `<Router>` element needs to go outside the `Layout` so the MenuBar links will work, like
```
<Router>
  <Layout>
    ...etc
    <Layout.Content>
      <Switch>
        ...etc
      </Switch>
    </Layout.Content>
  </Layout>
</Router>
```

modules/cactus-web/src/Layout/Layout.mdx
```diff
@@ -25,26 +26,45 @@ All the children except `Layout.Content` are optional, though ActionBar is highl
...
-    <Footer logo={LOGO}>The Foot clan returns.</Footer>
-  </Layout>
+      <BrandBar logo={LOGO} />
+      <MenuBar>
+        <MenuBar.Item>
+          <Link to="/home">Teenage Mutant</Link>
+        </MenuBar.Item>
+        <MenuBar.Item>
+          <Link to="/other/path">Ninja Turtles</Link>
+        </MenuBar.Item>
```
wilysword, line 44:
```suggestion
        <MenuBar.Item as={Link} to="/home">Teenage Mutant</MenuBar.Item>
        <MenuBar.Item as={Link} to="/other/path">Ninja Turtles</MenuBar.Item>
```

wilysword (commented):
Just a couple more clarifications.

modules/cactus-web/src/Layout/Layout.mdx
NicolasSimmonds: Well I tested this on the generated app from our cli and it worked this way. But if you prefer it outside `Layout` I'm happy to do it

Dhalton (approved):
LGTM -- looks like you do have merge conflict though.

wilysword (approved):

PR #414, opened 2020-11-30
CACTUS-418: blacklist stories for visual testing

NicolasSimmonds:
[CACTUS-418]

[CACTUS-418]: https://repayonline.atlassian.net/browse/CACTUS-418


modules/cactus-web/src/Button/Button.story.tsx
```diff
@@ -95,3 +96,4 @@ export const AsLink = (): ReactElement => (
 )
 
 AsLink.storyName = 'As Link'
+AsLink.parameters = { storyshots: false }
```
wilysword, line 99: I actually think we should keep this one. It does look just the same as the others, but because `<a>` has different default browser styles than `<button>`, I think the storyshot has value in case we accidentally remove something that makes them look the same.

modules/cactus-web/src/SelectField/SelectField.story.tsx
```diff
@@ -38,6 +38,7 @@ export const CustomStyles = (): React.ReactElement => (
     margin={text('margin', '3')}
   />
 )
+CustomStyles.parameters = { storyshots: false }
```
wilysword, line 41: Should this be on `ControllerForm` instead? That and `BasicUsage` look identical to me, but `CustomStyles` is slightly different. (Though it's not _that_ different, so maybe we should just disable both.)

wilysword (changes requested):
In Select.story.tsx, I think we could also remove the `Long option labels` storyshot.

I like that you changed the modals to default open, makes those storyshots much more meaningful.

wilysword (approved):

Dhalton (approved):
LGTM 👍

PR #415, opened 2020-12-01
CACTUS-419 Storyshot "before" callback

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-419

Surprisingly straightforward; good thing we upgraded our story format. There are probably more stories that could benefit, I just picked a few of the easiest/most obvious.


mikesoltow (approved):
Nice!  I love that the new screenshots even pick up on the touchscreen behavior of the Select.

NicolasSimmonds (approved):

PR #419, opened 2020-12-03
CACTUS-268: workaround testcafe click bug in Firefox

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-268

Mostly it was the click bug, though it seems Firefox also has a difference in how it formats the file input data, so I had to correct for that.

As far as I could tell, Safari was already passing (at least in browserstack), so I'm not sure what the issue was there.


mikesoltow (commented):
LGTM...but were we going to add `standard` to CI too?  Or did you decide against that for some reason?

mikesoltow (approved):

Dhalton (approved):

PR #420, opened 2020-12-03
CACTUS-430 :: Updated Tooltip Triggers

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-430

Unfortunately, I wasn't able to come up with a solution to close the tooltip without changing the focus when `autoTooltip` is enabled. I did do some investigating, and I think it's possible, but I believe the scope for the changes required would go beyond what I'd feel comfortable adding during work for this ticket. If we want to add that feature, I think we should create a separate ticket for it.

### Testing
1. Pull down this branch and bring up storybook in your web browser
2. Go to the tooltip story and click the icon
3. Verify that the content window stays open until you click outside the icon or the widget

I tested this using my phone and it worked for me, but if you're able to test it on your mobile device (or an emulator) that'd be great too.

### DIFF
--- a/modules/cactus-web/src/Tooltip/Tooltip.tsx
+++ b/modules/cactus-web/src/Tooltip/Tooltip.tsx
@@ -129,8 +129,7 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
     const { target } = event
     if (
       !(target instanceof Node) ||
-      (!(triggerRef.current && triggerRef.current.contains(target)) &&
-        !(widgetRef.current && widgetRef.current.contains(target)))
+      (!triggerRef.current?.contains(target) && !widgetRef.current?.contains(target))
     ) {
       setStayOpen(false)
     }
@@ -160,14 +159,10 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
             }
             style={{ maxWidth }}
             onMouseEnter={() => {
-              if (!stayOpen) {
-                setHovering(true)
-              }
+              setHovering(true)
             }}
             onMouseLeave={() => {
-              if (!stayOpen) {
-                setHovering(false)
-              }
+              setHovering(false)
             }}
           />
           <VisuallyHidden role="tooltip" id={id}>
--- a/modules/cactus-web/src/Tooltip/Tooltip.tsx
+++ b/modules/cactus-web/src/Tooltip/Tooltip.tsx
@@ -118,22 +118,22 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
   const [stayOpen, setStayOpen] = useState<boolean>(false)
 
   useEffect(() => {
+    const handleBodyClick = (event: MouseEvent): void => {
+      const { target } = event
+      if (
+        !(target instanceof Node) ||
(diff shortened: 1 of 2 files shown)


modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -112,22 +112,44 @@ const StyledInfo = styled(({ forceVisible, ...props }) => <NotificationInfo {...
...
+      document.body.removeEventListener('click', handleBodyClick)
+    }
+  }, [])
+
+  const handleBodyClick = (event: MouseEvent): void => {
+    const { target } = event
+    if (
+      !(target instanceof Node) ||
+      (!(triggerRef.current && triggerRef.current.contains(target)) &&
+        !(widgetRef.current && widgetRef.current.contains(target)))
```
mikesoltow, line 133: Could probably clean this up a bit with `?.` if you want, but not a big deal.

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -138,10 +160,14 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
             }
             style={{ maxWidth }}
             onMouseEnter={() => {
-              setHovering(true)
+              if (!stayOpen) {
+                setHovering(true)
+              }
```
mikesoltow, line 165: I wonder if the `if (!stayOpen)` is really needed here?  As long as it doesn't cause any weirdness (which it didn't seem to in a quick test I did), it's probably good to keep accurate track of the hovering state.  Not sure if there's a weird edge case where `hovering` would ever be `true` when it shouldn't because `mouseLeave` occurred while `stayOpen` was `true`.

mikesoltow (commented):
Looks pretty good to me.  Just a couple of questions.

modules/cactus-web/src/Tooltip/Tooltip.tsx
Dhalton: Yeah you're right, we can do without it. For some reason I thought they'd interfere with each other but after doing some testing I see that they don't. I'll push up and update in a bit!

mikesoltow (approved):
LGTM!

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -112,22 +112,43 @@ const StyledInfo = styled(({ forceVisible, ...props }) => <NotificationInfo {...
...
+
+  const handleBodyClick = (event: MouseEvent): void => {
+    const { target } = event
+    if (
+      !(target instanceof Node) ||
+      (!triggerRef.current?.contains(target) && !widgetRef.current?.contains(target))
+    ) {
+      setStayOpen(false)
+    }
+  }
```
wilysword, line 136: I would put this function definition inside the `useEffect`. Even if it makes you add things to the dependency list, the refs & state setter are all stable so the effect should still only run once.

wilysword (approved):
Looks good, just one suggestion for making the closure on that effect less confusing.

PR #421, opened 2020-12-04
CACTUS-429: allow configurable text for empty optins in select

NicolasSimmonds:
In the Select story, basic usage there is a knob to disable the option. If it's false the text "No options available" should be displayed.

In WithCombobox there is the same knob, but for it to show the no options available text, the **canCreateOption** knob should also be disabled.

[CACTUS-429]

[CACTUS-429]: https://repayonline.atlassian.net/browse/CACTUS-429

### DIFF
--- a/modules/cactus-web/src/Select/Select.tsx
+++ b/modules/cactus-web/src/Select/Select.tsx
@@ -251,7 +251,7 @@ const SelectTrigger = styled.button`
   min-width: 194px;
   width: 100%;
   height: 32px;
-  padding: 0 24px 0 16px;
+  padding: 0 28px 0 16px;
   background-color: transparent;
   ${(p): ReturnType<typeof css> => getShape(p.theme.shape)}
   ${(p): ReturnType<typeof css> => getBorder(p.theme.border)}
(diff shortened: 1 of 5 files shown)


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -56,6 +56,7 @@ export interface SelectProps
   matchNotFoundText?: string
   comboBoxSearchLabel?: string
   onDropdownToggle?: (prop: boolean) => void
+  noOptionsText?: string
```
mikesoltow, line 59: Can you add this to `propTypes` as well?  We should probably also add a `defaultProp`

mikesoltow (changes requested):
Looks good.  Just one little requested change.

Also, while I know this isn't in the scope of your ticket, I noticed that the no options text can bump right up to the down-arrow icon in a way that looks awkward.  Would you mind adding a little more padding to `SelectTrigger`?  I'd say your best bet is changing line 254 from `padding: 0 24px 0 16px` to `padding: 0 28px 0 16px', unless you can think of a better solution.

mikesoltow (approved):
Thanks, Nicolas!

wilysword (approved):

PR #426, opened 2020-12-07
CACTUS-428 :: IconButton Focus

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-428

### Testing
1. Bring up the IconButton story and focus the button to make sure the styles are being applied.
2. Wouldn't be a bad idea to test in IE as well since I had to do some special styles for IE.


wilysword (commented):
I was going to review further, but after your stand-up I decided to wait.

PR #437, opened 2020-12-10
New release script

mikesoltow:
This PR adds a new release script that (mostly) automates publishing a new package with Lerna, updating changelogs in the new format I've been using, and creating a release in GitHub.  It's not _fully_ automated, as the caller is given the option to override PR titles, classify what type of change each PR represents, and remove individual PRs from the changelog. However, it's still much faster than the manual process I've been using recently.

Don't worry...I didn't actually write most of this code.  I took almost all of it from Intuit's [Auto](https://intuit.github.io/auto/).  It's a pretty cool tool, but it didn't quite work well with the workflow we were using.  Plus I wanted more control over the text we use in changelogs, etc.

If you'd like to test this, you can run the script using `yarn release:new`.  Just make sure to bail out when it prompts you for your npm username.  You might also have to delete the changelog commit it creates if you get that far.

Before running the script, you'll need to set a `GITHUB_AUTH` env variable.  It's value should be a token you generate at https://github.com/settings/tokens.

You'll probably have to ignore a lot of PRs initially, as I don't think it works perfectly when you're not on the master branch.  But you can still narrow down the changelog to what you want.  If it's still a pain, we could make more tweaks later.

Also, this doesn't work for pre-releases yet, so I'll want to add in logic for that later.  Auto's code has logic for it, but I've removed it for now just to make sure this side of things works right.

Finally, I'm sure we'll want to package this up in UI tools or something so we can use it across all our repos, but we can do that later.


Dhalton:
Code looks good as far as I can tell...am I missing something as far as testing this goes, though? I keep getting the following error when I run `yarn release:new`:
```
(node:36071) UnhandledPromiseRejectionWarning: HttpError: Bad credentials
    at /Users/dhuber/Documents/Repos/cactus/node_modules/@octokit/request/dist-src/fetch-wrapper.js:49:31
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at Job.doExecute (/Users/dhuber/Documents/Repos/cactus/node_modules/bottleneck/light.js:405:18)
```

Do I need to put my github creds somewhere or something?

mikesoltow:
@Dhalton Good catch!  You do need a `GITHUB_AUTH` env variable set that has a GitHub token as its value.  You can generate a token at https://github.com/settings/tokens.  I'll update the PR description, but I'm going to hold off on adding anything to the README until I/we gain more confidence in this script and give it our official blessing, if that's all right.

wilysword (approved):
I don't much understand this side of things, but as far as I could tell this does what you said it does.

Dhalton (approved):

PR #448, opened 2020-12-14
CACTUS-443: add type=button to avoid submitting forms accidentally

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-443

Not as many of these as I thought there would be. No idea why the multiselect storyshot changed, though.


NicolasSimmonds (approved):

mikesoltow (approved):
LGTM!

PR #450, opened 2020-12-15
CACTUS-441: fix bottom border in brandbar's user menu in safari and edge

NicolasSimmonds:
The problem with safari was that it was adding a default margin so I had to delete it.
In edge the button had a border-radius, so I had to delete it too.

Test:
Check the Bottom border in the user menu of the Brandbar looks good in every browser

[CACTUS-441]

[CACTUS-441]: https://repayonline.atlassian.net/browse/CACTUS-441


wilysword (approved):

Dhalton (approved):
LGTM

PR #455, opened 2020-12-16
CACTUS-448 :: Toggle Alignment

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-448

### Testing

I added an "In Centered Table" story for the Toggle component, so you should check that out.  In addition, you should make sure all the other Toggle stories still work well.


modules/cactus-web/src/Toggle/Toggle.story.tsx
```diff
@@ -24,3 +25,25 @@ const ToggleManager = (props: any) => {
...
+      </Table.Row>
+      <Table.Row>
+        <Table.Cell align="center">
+          <Toggle />
+        </Table.Cell>
+      </Table.Row>
+      <Table.Row>
+        <Table.Cell align="center">
+          <Toggle />
+        </Table.Cell>
```
wilysword, line 45: You've got three cells, why not test all three alignments?

wilysword (approved):

modules/cactus-web/src/Toggle/Toggle.story.tsx
wilysword: We also probably don't need a storyshot for this one, though I guess I'm a little late making that point.

mikesoltow: :+1: good idea...adding that now

NicolasSimmonds (approved):

PR #463, opened 2020-12-17
CACTUS-292 :: BlackDuck Setup

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-292

This is the first of three PRs for BlackDuck. The other two will be in `ui-tools` and `ui-private`.

There isn't really a way to test this locally, as the script must be run through Jenkins to have access to the environment variables it needs. If you want, you can trigger the Jenkins job [here](https://jenkins.internal.repay.ninja/blue/organizations/jenkins/ui-blackduck/activity). You can also check out the project in BlackDuck [here](https://repay.blackducksoftware.com/api/projects/032b147c-1126-4910-bb80-e119676db52c/versions/9063aa7a-168a-47a7-b355-9e564e2f1c68/components?sort=projectName%20ASC&offset=0&limit=100&filter=bomInclusion%3Afalse&filter=bomMatchReviewStatus%3AREVIEWED). Note that you'll need to be on the VPN to access both of those links.


scripts/blackduck.sh
```diff
@@ -0,0 +1,16 @@
...
+    --detect.project.name="github.com/repaygithub/cactus" \
+    --detect.project.version.name="master" \
+    --detect.code.location.name="cactus" \
+    --detect.project.version.phase="RELEASED" \
+    --blackduck.url="https://repay.blackducksoftware.com" \
+    --blackduck.api.token=${BLACKDUCK_TOKEN} \
+    --detect.project.version.distribution=SAAS \
+    --detect.cleanup=true \
+    --detect.tools.excluded=SIGNATURE_SCAN \
+    --detect.excluded.detector.types=MAVEN
```
mikesoltow, line 16: What do these lines do?

mikesoltow (approved):
Seems fine to me.  I have a question, but I'll approve anyway.

wilysword (approved):

scripts/blackduck.sh
Dhalton: These two lines are disabling signature scanning and the maven detector. Signature scanning is a form of virus detection, but it always failed anytime I ran it in any of our projects. Maven is a build tool used mostly for Java projects so it shouldn't be necessary here. This is based on the detect script I found in ByteMover v2 [here](https://github.com/repaygithub/repay-bytemover-v2/blob/master/scripts/blackduck_javascript.sh#L15#L16).

PR #465, opened 2020-12-18
fix(cactus-web): fix FileInput misaligned text

NicolasSimmonds:
In the FormField story I created a knob to change the width of the form so you can test this more easily.

Check that the text keeps the same structure in the FileInputField regardless of the width of the form element

[CACTUS-446]

[CACTUS-446]: https://repayonline.atlassian.net/browse/CACTUS-446


modules/cactus-web/src/FieldWrapper/FieldWrapper.story.tsx
```diff
@@ -187,7 +188,7 @@ const ExampleForm = ({ withValidations }: { withValidations?: boolean }): ReactE
     >
       <Box
         as="form"
-        width="50vw"
+        width={text('widht', '50vw')}
```
Dhalton, line 191: Typo here

Dhalton (approved):
LGTM. There's a small typo I pointed out but I'm still gonna give my approval. 👍

modules/cactus-web/src/FieldWrapper/FieldWrapper.story.tsx
NicolasSimmonds: I was thinking on deleting this knob after it got approved since it was only for testing reasons. So after I get two approvals this si going out anyways.

mikesoltow: Rather than having this temporary knob, should we maybe add a "full (or mostly full) width" story to the FileInput?  That would also help Emily verify this, as she wouldn't be able to do anything after you delete the knob.

modules/cactus-web/src/FileInput/FileInput.tsx
```diff
@@ -637,16 +652,18 @@ const FileInputBase = (props: FileInputProps): React.ReactElement => {
...
-        variant="action"
-        id={id}
-        aria-describedby={describedBy}
-        disabled={disabled}
-        onClick={handleOpenFileSelect}
-      >
-        <BatchstatusOpen iconSize="small" />
-        {buttonText}
-      </TextButton>
+      {files.length !== 0 && (
```
mikesoltow, line 655: Can't this section be moved up into the lines 638-653 section?  `files.length !==0` seems redundant after `files.length === 0`.

mikesoltow (commented):
Just a few comments.

Also, it seems like the upload icon is _really_ far away from the text when the FileInput width is large (see screenshot).  Should we push those elements closer together?  I fiddled around and got what looked like a reasonable result when I did the following:
* Change `justify-content` to `center` in `EmptyPrompts`
* Remove the `width` prop from the `Flex` component inside `EmptyPromptsBase` (line 124).  Replace it with `m={3}`

That's just a suggestion though...up to you!
Screen Shot 2020-12-21 at 4 50 29 PM

mikesoltow (approved):
Changes look good...thanks!

PR #466, opened 2020-12-18
CACTUS-450 :: Use run-ete-tests CLI

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-450

I goofed a bit when I created the ticket; we don't run integration tests against the `theme-components` example app, so no need to make any changes there. Other than that, pretty self explanatory!

If you want to test locally you can run the integration tests for the `mock-ebpp` and `standard` example apps.


mikesoltow (approved):
LGTM

wilysword (approved):

PR #469, opened 2020-12-21
CACTUS-455 :: Modal Width Override

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-455

Pretty straightforward...to test you can bring up the Modal story and set the width using the width knob.


wilysword (approved):
Do you think that if `width` is given we should also set `box-sizing: border-box`?

Dhalton:
> Do you think that if `width` is given we should also set `box-sizing: border-box`?

That's probably a good idea. I could see that being a cause for some confusion...I'll add that in real quick. If anyone opposes just let me know

mikesoltow (approved):
LGTM

PR #470, opened 2020-12-31
CACTUS-451 ::  DateInput Improvements

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-451

This is a breaking change because one of the phrases (`showMonthYear`) was split into to `showMonth` and `showYear`, since the dropdowns are now separate.

Honestly, I'm not in love with my solution for the calendar's sliding motion; basically there are 3 calendars being rendered at all times, so I'm a bit concerned about performance there. If anyone has any ideas/suggestions on how that could be improved, I'm totally open to that.

Full disclosure here: The sliding motion when you click on the arrows to change the month doesn't seem to want to work in IE11. I spent some time trying to figure it out but couldn't come up with anything. I figure it's (hopefully) not the end of the world if the animation isn't shown in IE. Works fine in Chrome, FireFox, and Safari.

To test, bring up the DateInput stories and set some dates using the text fields. Bring up the calendar popup and use the arrows to change the month. Make sure the ACs on the ticket are being met, basically.


wilysword:
Looks pretty good, my initial testing only turned up a couple of issues.
- At least in Firefox, selecting a month/year using <space> instead of <enter> closes the entire dialog.
- When using the keyboard to open the month/year selection, it would be nice if focus moved automatically to the selection pane.
- There's a difference in behavior between mouse/keyboard: if you're scrolling through the months/years with the keyboard, the date value is updated with every change; if you click on a month/year, however, the selected date doesn't change (but it does take you to the correct month/year so you can click on a day to change the date). Not sure if this is by design or not.

Dhalton:
@wilysword 

Alright, I made some updates based on your feedback.

> * At least in Firefox, selecting a month/year using <space> instead of <enter> closes the entire dialog.

I haven't been able to reproduce this. I'm assuming you're on Linux? I'm not sure if that has anything to do with it, but I didn't see this issue on Firefox for Mac or Windows.

> * When using the keyboard to open the month/year selection, it would be nice if focus moved automatically to the selection pane.

This should be working now

> * There's a difference in behavior between mouse/keyboard: if you're scrolling through the months/years with the keyboard, the date value is updated with every change; if you click on a month/year, however, the selected date doesn't change (but it does take you to the correct month/year so you can click on a day to change the date). Not sure if this is by design or not (if it is I'm okay with it).

I think this was intentional before, but the prototype is updating the value when using the mouse, so I decided to do the same. The value in the DateInput should now update when selecting a month/year using your mouse. I also updated the behavior such that when the calendar opens, even if your date value is invalid (read: incomplete), the calendar will still open to month/year set in the value. For instance, if your date value is 02/##/#### (you only have the month of February set), the calendar will open to February of the current year. I think that behavior makes the most sense now that setting individual pieces of the date is easier to do and more probable to happen.

modules/cactus-web/src/DateInput/DateInput.test.tsx
```diff
@@ -320,20 +320,87 @@ describe('component: DateInput', (): void => {
...
+            onChange={handleChange}
+          />
+        </StyleProvider>
+      )
+
+      const portalTrigger = getByLabelText('Open date picker')
+      act((): void => {
+        fireEvent.mouseDown(portalTrigger)
+        fireEvent.click(portalTrigger)
+      })
```
mikesoltow, line 359: Just a thought, but I wonder if this block can be replaced with the following:

```suggestion
      userEvent.click(portalTrigger)
```

mikesoltow (commented):
I have to run for my house appointment, but I wanted to leave you with what I've noticed so far (aside from the fact that this looks way better!).  I'll finish my review later.

Quirk 1: Say you start on 1/31/2018.  If you click the up arrow when highlighted on the month, you see 2/31/2018 displayed.  The onChange handler gets triggered with a blank date.  Is that expected?  I'm not sure I can come up with something better, but I'm not sure it's super intuitive either.  Should the component automatically display some sort of error state that an invalid date has been selected?

Also, I noticed some weird behavior when you take the following steps:
1. Select January 31, 2021 using the calendar.  All is well.
2. Open the calendar back up and change the month to March using the dropdown.  All is still well: you can see that `onChange` is fired with `2021-03-31`.
3. Change the year to 2014 using the dropdown.  Note that the day value changes unexpectedly (see screenshot of the `onChange` calls).

Screen Shot 2021-01-06 at 11 29 11 AM

wilysword:
> I haven't been able to reproduce this
Now I can't reproduce it either, so either you fixed it accidentally or something else was going on, so I guess don't worry about it.

Dhalton:
> Quirk 1: Say you start on 1/31/2018. If you click the up arrow when highlighted on the month, you see 2/31/2018 displayed. The onChange handler gets triggered with a blank date. Is that expected? I'm not sure I can come up with something better, but I'm not sure it's super intuitive either. Should the component automatically display some sort of error state that an invalid date has been selected?

Went ahead and added this feature like we discussed. So now, if a user selects an invalid date, an error will be shown. This error will override any status messages set by `DateInputField`, if necessary.
 
> 1. Select January 31, 2021 using the calendar.  All is well.
> 2. Open the calendar back up and change the month to March using the dropdown.  All is still well: you can see that `onChange` is fired with `2021-03-31`.
> 3. Change the year to 2014 using the dropdown.  Note that the day value changes unexpectedly

This was a result of the list click handler setting the new date based on `focusDay` instead of `value`. It was fine to do that before, since selecting a year/month from the dropdown previously only opened up the month/year on the calendar but didn't set the value yet. Now that the value is being set via the dropdown, we have to generate the new date from the current value rather than `focusDay`, because it's entirely possible (and likely) that a user mouses over a day in the calendar on their way to opening up the dropdown, which updates `focusDay` and was throwing off the new value.

mikesoltow (commented):
Once again I have to leave before a full review, so I still don't have that much for you.

I did some thinking and research regarding the current animation solution, and I think I'm happy with it as a first pass.  I'm wondering if we can optimize it with a tech debt ticket or something like that.  I haven't actually implemented anything like this, but I wonder if we can use something like [react-transition-group](https://github.com/reactjs/react-transition-group).  I didn't get to dive too deeply, but it seems like using that you might do something like:
* Have only the visible Calendar rendered initially
* When a month arrow is clicked, render the new Calendar and unmount the currently visible one.  They each have appropriate enter/exit styles to simulate the slide animation on mount/unmount.

Maybe that's not possible, but perhaps worth looking into at some point.  For now, I don't notice any major problems with what you have.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -441,20 +495,23 @@ const CalendarDayBase = styled.button.attrs({ type: 'button' })`
...
     return (
       <CalendarDayBase {...rest} ref={ref}>
-        <span aria-hidden="false" hidden>
-          {longLabel}
-        </span>
+        {enableInteraction && (
+          <span aria-hidden="false" hidden>
+            {longLabel}
+          </span>
+        )}
```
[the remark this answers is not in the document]
Dhalton: Huh, I believe it's supposed to be for screen readers. I just checked and it looks like this wasn't working before, either. I'll see if I can get this fixed today.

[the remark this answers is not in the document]
Dhalton: I ended up just swapping out that `span` with `VisuallyHidden` which seems to have worked. Removing the `hidden` attribute just made all of the text visible in the calendar and for whatever reason VoiceOver was completely ignoring that element even though `aria-hidden` was set to false. 🤷‍♂️

wilysword:
The date's no longer changing randomly, but I found two things:

- Select the 31st of any month, then use the month drop down to change to February. If you use the mouse it'll change the day to 28 (or 29 for leap years), but if you use the keyboard it'll stay February 31st. I don't think this is a huge deal if you don't fix it, but it leads into the next issue:
- When you select an invalid date as I described above, the invalid date error message is not visible for me immediately, until I open the dialog again or move focus around a bit. Likewise once I've fixed the error, the error message doesn't disappear right away. Not sure if that was by design, but it does seem a little confusing.

wilysword (approved):
All my concerns have been addressed, I think.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1184,19 +1296,54 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
     }
   }
 
+  private handleLeftMonthArrowClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
+    const { month: focusMonth, year: focusYear } = this._getMonthData()
+    this.setState(
+      (state): Pick<DateInputState, 'focusDay' | 'value' | 'transitioning'> => {
+        const { value } = state
+        const pd = PartialDate.from(value, 'YYYY-MM-dd')
+        pd['setMonth'](focusMonth === 0 ? 11 : focusMonth - 1)
```
mikesoltow, line 1313: Doesn't matter too much, but I think you can do the following, right?

```suggestion
        pd.setMonth(focusMonth === 0 ? 11 : focusMonth - 1)
```

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1184,19 +1296,54 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
+        if (focusMonth === 0) {
+          pd['setYear'](focusYear - 1)
```
mikesoltow, line 1315: Same comment here

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1184,19 +1296,54 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
+
+  private handleRightMonthArrowClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
+    const { month: focusMonth, year: focusYear } = this._getMonthData()
+    this.setState(
+      (state): Pick<DateInputState, 'focusDay' | 'value' | 'transitioning'> => {
+        const { value } = state
+        const pd = PartialDate.from(value, 'YYYY-MM-dd')
+        pd['setMonth'](focusMonth === 11 ? 0 : focusMonth + 1)
+        if (focusMonth === 11) {
+          pd['setYear'](focusYear + 1)
```
mikesoltow, line 1335: And in this block

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1184,19 +1296,54 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
+          pd['setYear'](focusYear - 1)
+        }
+        pd.ensureDayOfMonth()
+        this.raiseChange(event, pd)
+        return { focusDay: pd.format(), value: pd, transitioning: 'slide-right' }
+      }
+    )
+    this._transitionTimeout = setTimeout(() => {
+      this.setState({ transitioning: false })
+    }, 400)
```
mikesoltow, line 1324: Why is this timeout longer than the right arrow click?

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1477,123 +1645,218 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
+                </IconButton>
+                <Flex justifyContent="center" alignItems="center">
+                  <MonthSelect
+                    className={isOpen || undefined}
+                    tabIndex={0}
+                    onClick={this.handleSelectMonthClick}
+                    aria-label={isOpen === 'calendar' ? phrases.showMonth : phrases.showCalendar}
+                    aria-roledescription="toggles between calendar and month selectors"
+                  >
+                    <span id={monthId}>{phrases.months[focusMonth].long}</span>
```
mikesoltow, line 1762: This bug was pre-existing, but we might as well fix it now.  To reproduce, open the DateInput story, click on the month section of the input, and try to type in `01`.  As soon as you type `0`, it will blow up with the following error:

```
Cannot read property 'long' of undefined
```

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1477,123 +1645,218 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
+            <StatusMessage
+              status="error"
+              id={statusId}
+              style={{
+                marginTop: '4px',
+                maxWidth: this._inputWrapper.current?.getBoundingClientRect().width,
+              }}
+            >
+              {phrases.invalidDateLabel}
+            </StatusMessage>
```
mikesoltow, line 1731: I could go either way, but I was envisioning that the `DateInputField` would handle the display of this in response to `onInvalidDate`, and that `DateInput` would just trigger the event for the wrapping component to do with what it will.  I lean toward that solution, as users might want more control over error message display if they're using a raw `DateInput` instead of our `AccessibleField` setup, but I can also see the value in the raw component just handling everything for you.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -873,35 +944,53 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
       })
     }
+    const { value, invalidDate } = this.state
+    const { onInvalidDate } = this.props
+    if (!value.isValid() && value.MM && value.dd && value.YYYY && !invalidDate) {
+      this.setState({ invalidDate: true })
+      if (onInvalidDate && typeof onInvalidDate === 'function') {
+        onInvalidDate(true)
+      }
+    } else if (value.isValid() && invalidDate) {
```
mikesoltow, line 981: One quirk with this is that, in datetime mode, an "invalid" error will appear after you type in or select a date but before you select a time.  I would think that we should probably shouldn't trigger it before the date/time is complete.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1477,123 +1645,218 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
-                    autoComplete="off"
-                    onChange={noop}
-                  />
-                )
+        <Flex alignItems="flex-start" justifyContent="center" flexDirection="column">
+          <InputWrapper
+            className={className}
+            role="group"
+            ref={this._inputWrapper}
+            aria-describedby={`${ariaDescribedBy} ${statusId}`}
```
mikesoltow, line 1662: This causes a broken aria reference is `ariaDescribedBy` is `undefined`

mikesoltow (commented):
Just a few more comments!  I still have to test on IE/mobile phones in BrowserStack, but most of my testing is done.

One generic comment is that the styles look a bit off in Firefox.  Not sure if there's an easy way to fix that.
Screen Shot 2021-01-12 at 5 46 57 PM

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1347,13 +1510,13 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
   }
 
-  private togglePortalView(): void {
+  private togglePortalView(triggerType?: 'month' | 'year'): void {
     this.setState(
-      (state): Pick<DateInputState, 'isOpen' | 'focusDay'> => {
-        const toggleToCal = state.isOpen === 'month-year'
-        const updates: Pick<DateInputState, 'isOpen' | 'focusDay'> = toggleToCal
+      (state): Pick<DateInputState, 'isOpen'> => {
+        const toggleToCal = state.isOpen === 'month' || state.isOpen === 'year'
```
mikesoltow, line 1516: Should we replace this variable by a new trigger type?  With the way it's currently implemented, if you open the month select, but don't select a month, and then click on the year select, the calendar is shown instead of the year select.  Instead maybe we can do the following below this line:

```javascript
const updates: Pick<DateInputState, 'isOpen'> = { isOpen: triggerType }
```
where `triggerType` is of type `'month' | 'year' | 'calendar'` (and is required)

Also, we'd have to do `this._shouldUpdateFocusDay = triggerType === 'calendar'` or something

mikesoltow (changes requested):
This morning I got around to testing on IE11 and mobile, and I found a few more things:

* The up/down arrows in the input field itself don't seem to be working in IE11
* When testing on a Samsung Galaxy S6 on BrowserStack, I ran into an issue on the `datetime` input.  After you select a date in the calendar, when you tap on the time picker to select a time, the keyboard covers up the input (see screenshot).  I found some links on SO (i.e. https://stackoverflow.com/questions/31535898/android-keyboard-hide-text-inputs), but they're all pretty hacky.  It's POSSIBLE we may just have to re-think the design so that the time picker in the calendar popup doesn't require the keyboard. (See https://material-ui-pickers.dev/ for a possible example.). If we do that, we'll need to discuss whether or not to do that as a follow-up ticket (would that mean two breaking changes in quick succession?), but hopefully it won't come to that.
* In IE11, the styles of the text in the input field are messed up in a similar way to Firefox.

Dhalton:
> One generic comment is that the styles look a bit off in Firefox. Not sure if there's an easy way to fix that.

Yeah, that's always been like that. I remember James having some trouble with it when he originally built the component. I'll see what I can do, though.

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: I just wanted to make sure the transition had finished before updating the state, I can set it to 300ms like the transition, though.

Dhalton: Oh, I see what you mean. That's a mistake. Nice catch!

modules/cactus-web/src/DateInput/DateInput.tsx
Dhalton: I like this idea better. I'll update this today.

mikesoltow (approved):
Changes look great to me, thanks!  I'll make a note to reach out to the designers about the mobile issue, but I'm good shipping this as-is.

PR #473, opened 2021-01-04
CACTUS-447 :: SonarQube Fixes

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-447

### Testing

* In Storybook, play around with a two-color theme.  Ensure that everything seems to be working normally when you edit both the primary and secondary colors.
* Ensure that the following components work as expected:
  * ActionBar
  * Button (disabled vs. non-disabled states)
  * CheckBoxGroup
  * ConfirmModal (icon and the variants)
  * DataGrid (top and bottom sections)
  * DateInput/DateInputField
  * FileInput/FileInputField
  * Footer (play around with adding/removing links and changing the screen size to ensure that the columns work correctly)
  * Layout
  * MenuBar

### DIFF
--- a/modules/cactus-theme/src/theme.ts
+++ b/modules/cactus-theme/src/theme.ts
@@ -141,6 +141,30 @@ const lightGray = `hsl(0, 0%, 90%)`
 const mediumGray = `hsl(0, 0%, 70%)`
 const darkGray = `hsl(0, 0%, 50%)`
 
+/** Notification Colors */
+const success = `hsl(145, 89%, 28%)`
+const error = `hsl(353, 84%, 44%)`
+const warning = `hsl(47, 82%, 47%)`
+const transparentSuccess = `hsla(145, 89%, 28%, 0.3)`
+const transparentError = `hsla(353, 84%, 44%, 0.3)`
+const transparentWarning = `hsla(47, 82%, 47%, 0.3)`
+const errorDark = `hsl(353, 96%, 11%)`
+const warningDark = `hsl(47, 96%, 11%)`
+const successDark = `hsl(145, 96%, 11%)`
+
+const status: StatusColors = {
+  background: {
+    success: transparentSuccess,
+    warning: transparentWarning,
+    error: transparentError,
+  },
+  avatar: {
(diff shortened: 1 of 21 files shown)


modules/cactus-theme/src/theme.ts
```diff
@@ -141,6 +141,30 @@ const lightGray = `hsl(0, 0%, 90%)`
...
+const status: StatusColors = {
+  background: {
+    success: transparentSuccess,
+    warning: transparentWarning,
+    error: transparentError,
+  },
+  avatar: {
+    success: `hsla(145, 89%, 28%, 0.3)`,
+    warning: `hsla(47, 82%, 47%, 0.3)`,
+    error: `hsla(353, 84%, 44%, 0.3)`,
```
wilysword, line 164: These appear to be the same as the `transparent*` colors, perhaps you should just reference those constants.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -54,8 +54,8 @@ const useVariant = (): Variant => {
   return 'top'
 }
 
-const getMenuItems = (menu: HTMLElement) =>
-  Array.from(menu.querySelectorAll(ITEM_SELECTOR)) as HTMLElement[]
+const getMenuItems = (menu: HTMLElement): HTMLElement[] =>
+  Array.from(menu.querySelectorAll(ITEM_SELECTOR))
```
wilysword, line 58: What's this issue? I think I'll need to make this change to my open PR and I'd like to make sure I do the right thing.

wilysword (approved):

modules/cactus-web/src/MenuBar/MenuBar.tsx
mikesoltow: Here's a screenshot of the error.  Taking a second look at it, I'm not sure this was actually a redundant cast.  However, the `querySelectorAll` method take a generic, so that's probably the best way to handle it.  I pushed up a change that does that.
Screen Shot 2021-01-05 at 1 43 53 PM

Dhalton (approved):
LGTM 👍

PR #475, opened 2021-01-04
CACTUS-466 :: Accordion Deletion Bug

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-466

When accordions were unregistered from an un AccordionProvider, the provider would still have the deleted accordion's ID in the `uncontrolledOpen` state.  This could cause bugs in other parts of the code.


Dhalton (approved):
LGTM 👍

wilysword (approved):

PR #478, opened 2021-01-08
CACTUS-468: trim spaces from hex color values

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-468

I think that all hex values go through that one function, so that should be the only change needed. I'm not sure if there's a place this can be manually tested, but I added some spaces randomly in a couple of the unit tests.


mikesoltow (approved):
LGTM

Dhalton (approved):

PR #479, opened 2021-01-11
CACTUS-463 :: Modal Fixes

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-463

### Testing
1. Change the modal's content to "This is a Modalaslkdfjlaskdjflkasdjflkasjdflkajdlkfjasldkfjlaskdjflkasjdflkj" & make sure the content does not overflow.
2. Set the modal's height/maxHeight using the knobs & make sure the values you set are being respected.


modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -78,9 +83,14 @@ export const ModalPopUp = styled(DialogOverlay)<ModalPopupProps>`
     margin: auto;
     max-width: ${(p) => !p.width && '80%'};
     ${width}
+    ${height}
+    ${maxHeight}
     outline: none;
     padding: 64px 24px 40px 24px;
     position: relative;
+    .modal-children {
+      word-break: break-word;
```
mikesoltow, line 92: This is still overflowing in IE11.  I think we need to replace it with the following, as we do elsewhere:

```suggestion
      overflow-wrap: break-word;
      word-wrap: break-word;
```

mikesoltow (changes requested):
Just a few comments, one in the code and one below.

It looks like the text will overflow if you set the max height (see screenshot).  I'm not sure what we want the behavior to be.  Should there be a scrollbar in the Modal?

Also, should we set a default max height (i.e. a percentage of the viewport, so it never takes up the full vertical space) or just leave that to the user?

Screen Shot 2021-01-11 at 5 44 15 PM

modules/cactus-web/src/Modal/Modal.tsx
Dhalton: Hmmmm, that's a problem then. I originally wanted to do this with `overflow-wrap`, but I kept running into issues with it being overridden. The weird thing is: The value it kept overwriting it with _was_ `break-word`, so `overflow-wrap: break-word` doesn't work here. I'll have to see if I can figure out a different way, unfortunately.

Dhalton:
> It looks like the text will overflow if you set the max height (see screenshot). I'm not sure what we want the behavior to be. Should there be a scrollbar in the Modal?

I would say a scrollbar is ideal in this situation.

> Also, should we set a default max height (i.e. a percentage of the viewport, so it never takes up the full vertical space) or just leave that to the user?

I'm not opposed to setting a default. The max height I've seen used in channels is 90%, should we go with that?

modules/cactus-web/src/Modal/Modal.tsx
Dhalton: I think the problem had something to do with some conflicting Flex styles...I added a content div fo the children and applied the styles there, which seems to have worked.

wilysword (approved):
Honestly, I think it looks ugly with a scrollbar. It'd look better if the scrollbar were on the inner div, but for some reason I can't get it to respect height/max-height when I do that. Therefore I'm approving since there doesn't seem to be an alternative.

wilysword:
Actually I did think of one other thing to try and it worked, so if you want to play around with it:
- add `flexWrap="nowrap"` to the inner Flex component
- add `overflow-y: auto` to the .modal-children class
- add the height/maxHeight to the .modal-children class, and change the default max height to something like `60vh` (and probably remove them from the modal div)
- (you can also make a new styled component for the inner div with all those styles)

I'll let my approval stand, but if you want to play around with which scroll looks better to you I'd appreciate it.

Dhalton:
@wilysword 

I agree with you that the scroll bar looks better on the inner flex. I have a couple of questions, though...

> * add the height/maxHeight to the .modal-children class

I can do this, but that would mean whatever value is passed for height/maxHeight will be the height of the inner flex instead of the entire modal. I feel like that could be confusing to developers, don't you?

If we _do_ go with the solution above, wouldn't that mean that we no longer need to set the modal to `box-sizing: border-box` like you mentioned, since the height isn't going to be applied there anyway?

wilysword:
> I feel like that could be confusing to developers, don't you?

That's probably true, but I don't much like the alternatives. There's the way you have it right now which is fairly intuitive but ugly, my suggestion which is not very intuitive, or we could just not control the height at all and let the developers control the height by passing in a div or something with the styles they want. Maybe we could make it a little more intuitive, though, by calling the props `innerHeight` and `innerMaxHeight` or something.

> If we do go with the solution above, wouldn't that mean that we no longer need to set the modal to box-sizing: border-box like you mentioned

I suppose, though it would still make the width prop behave more intuitively.

Dhalton:
> Maybe we could make it a little more intuitive, though, by calling the props `innerHeight` and `innerMaxHeight` or something.

I could get on board with that. I'll do this.

> I suppose, though it would still make the width prop behave more intuitively.

There's some logic in the css that sets it to `box-sizing: border-box` when width prop is passed so we should be good on that front.

modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -93,17 +102,17 @@ export const ModalPopUp = styled(DialogOverlay).withConfig({
...
-    overflow: auto;
-    -ms-overflow-style: -ms-autohiding-scrollbar;
     outline: none;
     padding: 64px 24px 40px 24px;
     position: relative;
     .modal-children {
       max-width: 100%;
       overflow-wrap: break-word;
       word-wrap: break-word;
+      overflow-y: auto;
```
wilysword, line 112: Should probably make this just `overflow`, rather than `overflow-y`; I just remembered that one of the reasons this came up was because channels was planning on putting some kind of document in a modal that was going to be tall & wide.

wilysword (approved):
Looks even better now.

mikesoltow (approved):
This looks awesome!

PR #480, opened 2021-01-11
CACTUS-456 :: Pre-Release Script

mikesoltow:
Not much that you can really test here.  If you happen to notice anything wrong with the code, I'd definitely appreciate the heads up.  Otherwise, the main thing to look over would be the two commits to the changelog.  I used this script to release the latest round of pre-releases, and the generated changelog updates are included in this PR.  You can also try running it with `yarn release` if you'd like.  It should notify you that it will generate a pre-release since you are not on `master`.


wilysword (approved):
Don't see any problems from my limited knowledge.

mikesoltow:
Found one bug when doing a main branch release, and I pushed that up.  Looks good to go from my end.

Dhalton (approved):
Looks good as far as I can tell 👍

PR #483, opened 2021-01-12
CACTUS-462: make Alert close button visible in IE

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-462

Seems like the issue was with flex-basis and/or flex-grow, somehow shrinking the button to 0 width. I took the opportunity to fix the text overflow as well, since it was on my mind from the Modal PR.


Dhalton (approved):

mikesoltow (approved):
Looks good

PR #484, opened 2021-01-13
CACTUS-464: change keypress to keydown to avoid IE bug

wilysword:
UAT note: https://repayonline.atlassian.net/browse/CACTUS-464

I don't really remember why I went with keypress in the first place, but after some testing I found that keydown worked just as well in all browsers (I tested on button, div, and anchor with and w/out href), as long as it's paired with `onKeyUp={preventAction}` to prevent duplicate click events. As such, I decided to change all the keypress events, even the ones that weren't susceptible to the IE bug, just to make it less confusing for future devs.


Dhalton (approved):

mikesoltow (approved):
LGTM

PR #486, opened 2021-01-15
CACTUS-461: Blackduck fixes 1

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-461

Doesn't seem like two days worth of work, does it? I spent the rest of the time trying to upgrade two of the security dependencies, but it inexplicably made the visual tests fail even though they passed locally. After a day of that I just gave up, before I started to feel the urge to commit suicide.


mikesoltow (approved):
Ah man.  You have my condolences.  What you have here looks fine, though!

Dhalton (approved):

PR #487, opened 2021-01-15
CACTUS-469: fix z-index conflicts between components

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-469

The two conflicts I found were the Toggle/MenuBar, which originated the ticket, and the Modal overlay which didn't cover the Sidebar/ActionBar.

Also semi-standardized the values:
- Drop-down-type popups are all 1000
- Most layout stuff is 100
- Overlays are 100 + or -, depending on whether they cover everything or everything but the Sidebar

There's no real reason for the 1000, just that that's what most of them already were.


mikesoltow (approved):
LGTM

Dhalton (approved):
I'm good with it!

PR #488, opened 2021-01-19
CACTUS-472: ActionBar ARIA labels

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-472

BREAKING CHANGE: aria-label prop is now required on ActionBar.Panel & ActionBar.Item

I looked at the other components where a similar change might make sense (Sidebar.Button, BrandBar.Item); for the former, I decided that if devs are springing for the lower-level components they might need greater flexibility; and for the latter, I didn't want to require a prop for an optional situation that only happens on tiny screens, if at all.


Dhalton (approved):
LGTM

mikesoltow (approved):
LGTM

PR #489, opened 2021-01-20
CACTUS-458 :: Fix sub-menu collapse bug

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-458

I think the issue was caused by the `box.setFocus` call occurring before the `setExpanded` operation actually completed.  In a class component, you'd get around this by putting `box.setFocus` in the callback argument for `setState`.  Since hooks are used here, I spun up a quick hook that accomplishes the same thing (calling it once in a `useEffect` hook under the hood).  This negated the need for the delay feature of `setFocus`, as this accomplishes something similar, just in a more targeted way (firing the focus call after the `expanded` state changes instead of on the next render).


wilysword:
Unless I'm missing the issue, that's what the `delay` argument is already supposed to do. Did you try modifying how `delay` works before you tried this solution?

mikesoltow:
> Unless I'm missing the issue, that's what the `delay` argument is already supposed to do. Did you try modifying how `delay` works before you tried this solution?

Thanks for pointing that out...I thought delay was working a bit differently for some reason.  Let me dig in to see if I can figure out what's going on.

wilysword:
Rereading the ticket, since it happens on _close_ I think you might have it backwards: expanded is set to false first, causing the focus element to blur and the menu to close, and _then_ it attempts to set the focus to the button but it's already hidden. (Just a guess, and I have no idea why the order of operations would be different in different browsers.)

mikesoltow:
> Rereading the ticket, since it happens on _close_ I think you might have it backwards: expanded is set to false first, causing the focus element to blur and the menu to close, and _then_ it attempts to set the focus to the button but it's already hidden. (Just a guess, and I have no idea why the order of operations would be different in different browsers.)

There's something weird going on with the delay where, after clicking the `button` to collapse and seeing the initial blur events (with `currentTarget` being the sub-menu `ul` and `relatedTarget` being the `button`), I see additional blur events with the current/relatedTargets reversed.  Then a final round of blur events with `currentTarget` of the sub-menu `ul`, this time with `relatedTarget` of `null`.

I tried re-working the `delay` mechanism into something that can take a custom dependency (so you can specify a variable that the focus should wait for instead of just telling it to focus on the next render).  That seems to work as well, but I haven't fully tested it.  If you have time, let me know what you think (it's at https://github.com/repaygithub/cactus/compare/collapse-alternative), or I can push this up tomorrow assuming everything still looks good.

mikesoltow:
@wilysword I pulled in the solution I mentioned above into this PR.  It seems to be working well, although I'm doing a few more cross-browser tests right now to confirm.

mikesoltow:
Actually, don't review this yet (unless you want to comment on the general approach)...something is still broken.

wilysword:
That sounds kind of like the first blur is causing the menu to close, then the click event happens and it's trying to reopen it, though I have no idea why that would happen. My best advice is to find every call to `toggle` and throw a console.log next to it, see what's firing in what order.

I'm not positive, but I don't think your proposed solution should work: the delay is necessary when expanding, but when _hiding_ I think the delay would actually break it because if the active element (in the submenu) is hidden focus goes to the body, and the entire menu closes. (That's just my thought experiment of what I think would happen, I didn't actually try it.)

mikesoltow:
This should be ready for review now...I think 🤞

modules/cactus-web/src/helpers/state.ts
```diff
@@ -0,0 +1,24 @@
+import { SetStateAction, useCallback, useEffect, useRef, useState } from 'react'
+
+type StateSetterWithCallback<S> = (v: SetStateAction<S>, callback: () => void) => void
+
+const noop = () => undefined
+
+export const useStateWithSetterCallback = <S>(
```
wilysword, line 7: I'd put this in `helpers/react.ts`, and call it just `useStateWithCallback`.

modules/cactus-web/src/helpers/state.ts
```diff
@@ -0,0 +1,24 @@
...
+  initialState: S | (() => S)
+): [S, StateSetterWithCallback<S>] => {
+  const [state, setState] = useState<S>(initialState)
+  const storedCallback = useRef<() => void>(noop)
+
+  const setStateWithCallback = useCallback((v: SetStateAction<S>, callback: () => void) => {
```
wilysword, line 13: If you make the callback optional, this can be used in cases where sometimes you want a callback, and sometimes you don't.
```suggestion
  const setStateWithCallback = useCallback<StateSetterWithCallback<S>>((v, callback = noop) => {
```

modules/cactus-web/src/helpers/usePopup.ts
```diff
@@ -61,27 +62,27 @@ function usePopup(
...
+          focusHint = null
         }
-        setExpanded(isExpanded)
-      }
-      // A closed element can't focus by index or text search.
-      if (!isExpanded && typeof focusHint !== 'object') {
-        focusHint = null
-      }
-      if (focusHint !== undefined) {
-        box.setFocus(focusHint, focusOpt)
```
wilysword, line 84: With the changes you've made here, setFocus is only called if the state changes where before, it could be called even if the state was the same.

wilysword (commented):
I've got some code organization suggestions, and noted one functionality change which you may or may not wish to address.

Honestly, though, it makes no sense to me that this works. If the element becomes hidden it _should_ lose focus which _should_ close the menu, but somehow it doesn't. Baffling. It _does_ work, though, so I'll approve as soon as you decide how to address my comments.

mikesoltow:
@wilysword Okay, after a week, I think I've finally addressed your comments.

wilysword (approved):

Dhalton (approved):
Works great!

PR #491, opened 2021-01-21
CACTUS-476: add integration tests for DateInput

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-476

Since I couldn't think of a way of working the datetime/time inputs into the example app, I made a "hidden" page specifically for the integration tests. With that done, I felt it made more sense to move the existing DateInput tests to that file as well.

Finally, I also fixed an issue where IE (I think it's actually a polyfill library) was adding an extra character to the month names.


Dhalton (approved):
LGTM

mikesoltow (commented):
I like these a lot.  Just one suggestion for an additional assertion.

mikesoltow (approved):
LGTM

PR #492, opened 2021-01-25
CACTUS-480: prevent duplicate options in Select

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-480

Two parts to this:
1. Making sure values can only be added programmatically if `canCreateOption` is true.
2. Do not display duplicates from the extra options (does not prevent duplicates among passed-in options).

No stories can test this as-is, but if you want to be thorough you can the modifications in the reproduction section of the ticket to test this manually; beyond that I also added a unit test to cover it.

Mike suggested this was a breaking change; is that still the case if the previous behavior was more of a bug than a feature?

### DIFF
--- a/modules/cactus-web/src/Select/Select.test.tsx
+++ b/modules/cactus-web/src/Select/Select.test.tsx
@@ -1172,6 +1172,24 @@ describe('component: Select', (): void => {
       expect(listbox).toHaveTextContent('superior')
       expect(listbox).toHaveTextContent('globe')
     })
+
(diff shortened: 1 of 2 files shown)


Dhalton (approved):
As far as whether or not this is a breaking change goes, I guess if it's a bug then we could get away with doing a minor release.

mikesoltow:
> As far as whether or not this is a breaking change goes, I guess if it's a bug then we could get away with doing a minor release.

I can get behind this

mikesoltow (approved):
LGTM

PR #493, opened 2021-01-26
CACTUS-485: MenuBar Focus Flash Bug

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-485

Mac (for some reason) does not like focusing buttons on click, which led to the list being focused instead and triggering the "focus to first menuitem" code. The best solution is to swap out `button` for another element: in this case I chose `span` for its similarity to `a`, which is probably the most commonly used `as` prop for `MenuBar.Item`.

I wasn't expecting anything to change visually, and even looking at the diff I'm not sure why it did, or why the MenuBar storyshot _didn't_ change.


Dhalton (approved):
Clever solution. Works like a charm.

wilysword:
I really don't understand why browsers still do this kind of weird stuff. If I were writing a browser today, I'd write every element as "generic box with default CSS and some default behavior"; like a `<button>` is pretty much just a div with role=button, tabindex=-1, and onkeypress=(e)=>{if(e.key==='Enter')e.target.click()}.

mikesoltow (approved):
LGTM

PR #494, opened 2021-01-27
CACTUS-484: prevent content stretching on overflow

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-484

Used this as a reference: https://css-tricks.com/preventing-a-grid-blowout/
It's weird how often the solution is just to add a min width... Though it would be an interesting challenge, I am nonetheless glad I don't have to implement the CSS spec.

There will need to be a ticket on the channels board for updating cactus version and making sure it fixes the problem everywhere (I only really checked on the landing page) once this is released. If the basic fix (min width) doesn't fix the problem everywhere, we may need to add the overflow fix as well.


Dhalton (approved):
LGTM - tested it out in Channels and it looks much better.

mikesoltow (approved):
LGTM

PR #496, opened 2021-01-27
Cactus 494: iconButton variants

daniloPenaR:
[CACTUS-494](https://repayonline.atlassian.net/browse/CACTUS-494)

Note: 
When I was setting up the success and warning variants into `inverseVariantMap` I didn't see any inverse styles for those new variants. 
So I put the same styles in both `variantMap` and `inverseVariantMap`


Dhalton:
> When I was setting up the success and warning variants into `inverseVariantMap` I didn't see any inverse styles for those new variants.
> So I put the same styles in both `variantMap` and `inverseVariantMap`

Danilo asked me about this yesterday but I wanted to get the rest of the team's opinions on it. We have to set _something_ in the `inverseVariantMap` under the `success` and `warning` keys (or update the code so that it doesn't use the inverse map if it's using one of those variants) but there aren't any inverse styles for those variants in the style guide.

Do we even need the inverse variants? If we decide on keeping them, we should probably turn it into a full-fledged feature and make sure any applicable components use it & that designers are accounting for it. Honestly, though, I don't think it's used and I would probably be okay with removing code related to inverse components.

I think we've had discussions around this before but nothing really came of it so I'm curious to hear your guys' thoughts.

Dhalton:
Aside from the inverse variant stuff, this looks really good, Danilo! I'll be ready to approve when we get that sorted out.

wilysword:
Aside from the standard-type button, I'm not sure _any_ of our components have proper inverse designs. Personally I think we should leave these variants with no inverse now, and if we want to add more inverse functionality later it should be a concerted effort across the whole library.

wilysword (approved):
Matches the style guide, looks good to me. I'm curious, though, do you know why two of the screenshots show the bright yellow and the last one shows the correct dark yellow?

Dhalton (approved):
Seems like we will probably leave the inverse designs alone for now, so I'll go ahead and get my approval on this. Nice work, Danilo.

PR #499, opened 2021-01-29
CACTUS-493: Fix I18nSection remount bug

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-493

This works by delaying the context change until the new language is loaded. Apparently Mari doesn't think this will work for channels, but since the only alternatives I can think of are the previous remounting behavior (which could cause lost state), and flashing the content key names (default behavior for missing translations) while the sections load; I think they'll just have to work around it on their end if they don't like any of those options.

I don't know any way of testing this other than sticking `console.log` statements in the code. Specifically for the remounting bug, I just added this component as a child of an `I18nSection` in the standard example app:
```
class X extends React.Component {
  componentDidMount() {
    console.log('X MOUNTED')
  }
  render() {
    return null
  }
}
```
If you run it and change the language on the `master` branch, you should see the `X MOUNTED` log twice. Change to this branch, refresh the page and change languages, you should only see it once.

### DIFF
--- a/modules/cactus-i18n/src/BaseI18nController.ts
+++ b/modules/cactus-i18n/src/BaseI18nController.ts
@@ -133,7 +133,7 @@ export default abstract class BaseI18nController {
     return id
   }
 
-  protected negotiateLang(lang?: string, strict = false): string[] {
+  public negotiateLang(lang?: string, strict = false): string[] {
     if (lang === undefined) {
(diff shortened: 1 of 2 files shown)


Dhalton (approved):
LGTM, works well.

mikesoltow (approved):
LGTM

PR #500, opened 2021-01-29
CACTUS-498 :: Alert Close Button Variants

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-498

Pretty quick & easy. I'm not sure why the storyshots didn't update...I guess it wasn't able to detect the color change on such a small piece of a component. 🤷

### DIFF
--- a/modules/cactus-web/src/Alert/Alert.tsx
+++ b/modules/cactus-web/src/Alert/Alert.tsx
@@ -66,8 +66,15 @@ const typeVariant = (props: AlertProps): ReturnType<typeof css> | undefined => {
   }
 }
 
+const statusToVariantMap: { [K in Status]: 'success' | 'warning' | 'danger' | 'action' } = {
+  success: 'success',
+  warning: 'warning',
+  error: 'danger',
+  info: 'action',
+}
+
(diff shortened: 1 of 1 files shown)


wilysword (approved):

wilysword:
This got me to thinking: besides the infamous "inverse" variants, I wonder if some components might also need "dark" variants. To me it's kind of weird that the warning IconButton is so much darker than success or error, but in this case it looks better than they do. So you'd have the normal variants for a white background, the dark variants for a light background, and the inverse variants for a dark background. Something to think about for the future, maybe.

mikesoltow:
> This got me to thinking: besides the infamous "inverse" variants, I wonder if some components might also need "dark" variants. To me it's kind of weird that the warning IconButton is so much darker than success or error, but in this case it looks better than they do. So you'd have the normal variants for a white background, the dark variants for a light background, and the inverse variants for a dark background. Something to think about for the future, maybe.

Yeah we were going back and forth on this.  I felt that just having a "dark" variant on all alerts looked fine, but channels/designers liked the button colors for success/error 🤷   So I wouldn't be opposed to revisiting this at some point.

mikesoltow (approved):
LGTM

PR #502, opened 2021-01-29
Cactus 496: Dimmer component

daniloPenaR:
[Cactus-496](https://repayonline.atlassian.net/browse/CACTUS-496?atlOrigin=eyJpIjoiZjFmMjE5ZjM2NTliNDUxNWFmYTM1M2JiZjY4MzE1NWIiLCJwIjoiaiJ9)


modules/cactus-web/src/Dimmer/Dimmer.story.tsx
```diff
@@ -0,0 +1,107 @@
...
+const LOGO =
+  'https://repay-merchant-resources.s3.amazonaws.com/staging/24bd1970-a677-4ca7-a4d2-e328ddd4691b/repay_logo_new.jpg'
+
+export const ContentDimmer = (): React.ReactElement => {
+  const [open, setOpen] = useState(false)
+  return (
+    <>
+      <Layout>
+        <BrandBar logo={LOGO} />
+        <Flex mx={15} justifyContent="start" alignItems="flex-start" flexDirection="column">
```
wilysword, line 26: The Layout's content needs to go in the `Layout.Content` component, or the styles get thrown off. I actually think you don't need the Layout or BrandBar, though, the Flex alone would do it for all these stories.

modules/cactus-web/src/Dimmer/Dimmer.story.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <h1>Page Dimmer</h1>
+        <p>
+          Mauris eu felis fringilla tortor scelerisque tincidunt et quis risus. Proin dui arcu,
+          tincidunt nec rhoncus eu, blandit ut odio. Donec ac tristique felis. Morbi vel urna
+          commodo, efficitur orci in, condimentum augue. Morbi neque felis, consequat sit amet mi
+          eget, imperdiet consectetur augue. Sed dignissim congue ex at vestibulum. Etiam rutrum,
+          mauris rutrum maximus congue, lectus sem molestie est, eget gravida orci ligula tempus
+          dolor.
+        </p>
+        <Button onClick={() => setOpen(true)}>Activate dimmer!</Button>
```
wilysword, line 72: When you focus this button and then activate it, the button retains focus meaning that although you can't click it, you can still activate it with the keyboard. When the dimmer opens, it should probably blur the active element (unless the active element is within the dimmer).

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -0,0 +1,66 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled from 'styled-components'
+
+export interface DimmerProps {
+  active: boolean
+  page?: boolean
```
wilysword, line 7: Mike and I had a discussion about this page prop (https://repay.slack.com/archives/CABGTU4S3/p1612199551048700), and I think it would be best to remove it (and also remove the `DimmableContent` component). Basically, we don't actually have a use case for that and it's generally best to leave out code that doesn't have a foreseeable use. (If you want to keep it, though, you can wait for a second opinion.)

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return <Container>{children}</Container>
+}
+
+const Dimmer: DimmerType = (props): React.ReactElement => {
+  const { active, page, children } = props
+  return (
+    <DimmerStyled active={active} page={page}>
+      {active && children}
+    </DimmerStyled>
+  )
```
wilysword, line 24: If you're going to use the `active` prop to conditionally render, I'd just go with conditionally rendering the whole thing:
```suggestion
  return active && (<DimmerStyled page={page}>{children}</DimmerStyled>)
```

modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -47,31 +48,33 @@ const Modalbase: FunctionComponent<ModalProps> = (props): React.ReactElement =>
...
-            alignItems="center"
-            flexWrap="nowrap"
-          >
-            {children}
-          </Flex>
-        )}
-      </DialogContent>
-    </ModalPopUp>
+    <Dimmer active={isOpen} page>
+      <ModalPopUp
```
wilysword, line 52: This isn't working right, because the `ModalPopUp` has it's own built-in overlay. What you need is to get the styles from Dimmer, but apply them to DialogOverlay:
```
export const ModalPopUp = styled(Dimmer).withConfig({
  shouldForwardProp: [omitted for brevity]
})<ModalPopupProps>`
  ...any addtional styles not covered by the Dimmer, etc.
`.withComponent(DialogOverlay)
```
and then put the rest of this code back how it was.

modules/cactus-web/src/Dimmer/Dimmer.story.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const ModalWithDimmer = (): React.ReactElement => {
+  const [open, setOpen] = useState(false)
+
+  return (
+    <Layout>
+      <Flex mx={15} justifyContent="start" alignItems="flex-start" flexDirection="column">
+        <h1>Modal with dimmer</h1>
+        <Button onClick={() => setOpen(true)}>Activate dimmer!</Button>
+        <Dimmer active={open} page>
```
wilysword, line 93: You don't need the Dimmer around the Modal component, it should be part of the Modal. You don't need this story at all, in fact, since there are already plenty of stories with modals in them.

wilysword (changes requested):

Dhalton:
I was about to submit my review but it looks like Glen beat me to the punch. I was going to mention a lot of the things he did, though, so I'll just +1 everything Glen said 😄

modules/cactus-web/src/Dimmer/Dimmer.tsx
daniloPenaR: Dhalton gave a +1 to everything you said, so, I'm going to remove it hahaha

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  render(): React.ReactElement {
+    const { active, children } = this.props
+    return active ? <DimmerStyled>{children}</DimmerStyled> : <></>
+  }
+}
+
+export const DimmerStyled = styled.div`
+  position: fixed;
+  display: flex;
+  background: rgba(46, 53, 56, 0.9);
```
Dhalton, line 32: We really should be using the colors from the theme here. It looks like the style guide specifies `darkestContrast`, so the way to set that using the theme would be:
`background: ${(p) => p.theme.colors.darkestContrast};`

modules/cactus-web/src/Dimmer/Dimmer.mdx
```diff
@@ -0,0 +1,71 @@
+---
+name: Dimmer
+menu: Components
+---
+
+import Dimmer from './Dimmer'
+import { LiveProvider, LiveEditor, LiveError, LivePreview } from 'react-live'
+import DateInput from './DateInput'
```
Dhalton, line 8: Looks like this line is preventing the docs site from building. Once this is gone, you can test the docs site by running `yarn docs start` from the root. I'd recommend giving it a quick look over as the docs site tends to operate according to its own rules.

Dhalton (changes requested):
Looks much better! Just a couple of things I noticed but I'll be ready to approve when those are taken care of.

modules/cactus-web/src/Dimmer/Dimmer.tsx
daniloPenaR: But if I set the Bg on that way and use the `opacity` property, the content will be affected by that opacity too.

modules/cactus-web/src/Dimmer/Dimmer.mdx
daniloPenaR: Thanks! I'll check it now.

modules/cactus-web/src/Dimmer/Dimmer.tsx
Dhalton: Ah, good point. I wonder if it would be worth it to add a new version of `darkestContrast` with this opacity to the theme in the future then. 🤔 Mike, Glen, what do you guys think?

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  z-index: 100;
+`
+
+Dimmer.propTypes = {
+  active: PropTypes.bool.isRequired,
+}
+
+Dimmer.defaultProps = {
+  active: false,
+}
```
wilysword, line 50: This is redundant with the static attributes on your class.

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  public static defaultProps = {
+    active: false,
+  }
+  componentDidUpdate(prevProps: DimmerProps): void {
+    if (prevProps.active === false && this.props.active === true && document.activeElement) {
+      try {
+        ;(document.activeElement as HTMLElement).blur()
+      } catch {}
+    }
+  }
```
wilysword, line 22: You don't have to change anything, but just for future reference you can do this in a functional component as well with the `useEffect` hook.
```
const { active } = props
React.useEffect(() => {
  if (active && document.activeElement) {
    ;(document.activeElement as HTMLElement).blur()
  }
}, [active])

**Comment by wilysword** [comment]
Oh wow, we almost forgot something extremely important: you also need to export this in the `index.ts` file:
```
export { DateInputField } from './DateInputField/DateInputField'
export { default as Dimmer } from './Dimmer/Dimmer'
```

Dhalton (approved):
LGTM, thanks for making those changes, Danilo

wilysword (approved):

PR #507, opened 2021-02-08
CACTUS-497: support complex options in Select component

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-497

The updated docs should have everything you need to know about how to use it.

Also changed IDs from using both label & value to just value, which is still supposed to be unique and should be less mutable (e.g. from translations and the like). Not sure if that would could as a breaking change or not, since it could be used in scripts.


Dhalton (approved):
LGTM

mikesoltow (commented):
Code looks pretty good.  Just one readability suggestion.  My initial UAT looked good as well, but I still want to check all browsers.  I need to head out for the night, so I'll finish that up in the morning.

mikesoltow (approved):
LGTM!

PR #508, opened 2021-02-08
CACTUS-504 :: ColorPicker

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-504

I had to create a component inline for the palette icon, so we'll definitely want to move that to the icon library after this is merged.

There is a small issue with Saturation element in IE11; you can't drag the pointer. You can click somewhere on the Saturation element and the pointer will move there, so I don't think it's a huge deal, but I want to point that out. I figure people will probably use the inputs to set the color for the most part, anyway. I tried to find some info on it but [this](https://github.com/casesandberg/react-color/issues/795) was the most similar issue I could find and there's no update on it so far.


modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,574 @@
...
+      }
+    }
+  }
+
+  const handleCancelClick = (event: React.MouseEvent) => {
+    if (onBlur && typeof onBlur === 'function') {
+      const cactusEvent = new CactusFocusEvent('blur', eventTarget, event)
+      onBlur(cactusEvent)
+    }
+    setOpen(false)
```
mikesoltow, line 429: In addition to closing the popup, don't we also need to reset the state to where it was?  Otherwise, the value has changed, but `onChange` was never triggered.

mikesoltow (commented):
Looks great!  The main thing for me is a question about the Cancel behavior (see the relevant comment in the code).  I'm about to start cross-browser tests, but I wanted to leave you with the feedback I have so far first.

For the screenshots, should we have it click the button to open the popup before the screenshot is taken?

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,574 @@
...
+    height: 16px;
+    border-radius: 50%;
+    top: 7px;
+    left: 7px;
+    border: 1px solid ${(p) => p.theme.colors.mediumGray};
+    background-color: ${(p) =>
+      p.$saturation
+        ? `hsl(${p.$hsl.h}, ${p.$hsl.s * 100}%, ${p.$hsl.l * 100}%)`
+        : `hsl(${p.$hsl.h}, 100%, 50%)`};
+  }
```
mikesoltow, line 239: I noticed this console warning generated by `styled-components`:

```
Over 200 classes were generated for component styled.div with the id of "sc-exqJjn".
Consider using the attrs method, together with a style object for frequently changed styles.
Example:
  const Component = styled.div.attrs(props => ({
    style: {
      background: props.background,
    },
  }))`width: 100%;`

  <Component />```
  
  Perhaps we want to control `background-color` with `attrs` instead?

**Review by mikesoltow — COMMENTED** [review]
Cross-browser testing looked good.  Just one additional comment regarding a console warning I saw.

File: modules/cactus-web/src/ColorPicker/ColorPicker.tsx (hunk 1 above)
Comment (mikesoltow) [reply modules/cactus-web/src/ColorPicker/ColorPicker.tsx:429]: I noticed afterwards that this weird behavior only happens on the uncontrolled version (since the `useEffect` hook only resets the value if one was passed in).  I still think it would be nice to reset the value, however, as otherwise Cancel and Apply aren't really different from each other in the uncontrolled case.

File: modules/cactus-web/src/ColorPicker/ColorPicker.tsx [hunk 3]
```diff
@@ -0,0 +1,574 @@
...
+    <>
+      <InputWrapper
+        id={id}
+        ref={inputRef}
+        $disabled={disabled || false}
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+      >
+        {!disabled ? (
+          <EditableInput
```
wilysword, line 490: What exactly is an EditableInput and why are we using it here? From what I'm seeing it doesn't seem to add anything over a regular input.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
Dhalton: Ah yeah you're right. I was mostly using the "Controlled" story so that's probably why I missed it. Thanks for letting me know. I will update this today

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
Dhalton: Hmmmm 🤔 do you know how this was triggered? Haven't seen this...I'll see if I can figure out how to reproduce it.

modules/cactus-web/src/ColorPicker/ColorPicker.story.tsx
```diff
@@ -0,0 +1,60 @@
...
+} as Meta
+
+const storyParams = {
+  cactus: { overrides: { alignItems: 'start', paddingTop: '32px' } },
+}
+
+const eventLoggers = {
+  onChange: (e: any) => {
+    console.log(`onChange '${e.target.name}':`)
+    console.log(e.target.value)
```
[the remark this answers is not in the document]
Dhalton: The value is an object so when I concatenated it with the string it just logged out "[Object] [object]" or something like that

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
wilysword: Why change the background color at all? Maybe instead have the element with a transparent background color, and a very thick border + solid outset box shadow?

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,574 @@
...
+        ) : (
+          <input value={`#${state.hex}`} disabled />
+        )}
+        <IconButton
+          disabled={disabled}
+          iconSize="small"
+          ml="auto"
+          label={phrases.triggerLabel}
+          onClick={handleIconButtonClick}
+        >
```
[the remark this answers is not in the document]
Dhalton: No real reason other than the fact that I didn't realize it was there and that I should. 😉 I'll update it to use the helper

modules/cactus-web/src/ColorPicker/ColorPicker.story.tsx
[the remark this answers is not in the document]
wilysword: You don't have to concatenate, you could log it like
```
console.log(`onChange '${e.target.name}':`, e.target.value)
```
However, now that I realize the `value` is a complex nested object, I suppose it's probably more readable to have it on a separate line.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
Dhalton: Whoa...mind = blown 🤯. That's a good idea, Glen. I'll give that a shot.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
Dhalton: From the [docs](https://casesandberg.github.io/react-color/)...

> Use EditableInput to display an input / label that acts as the single source of truth until the input is blurred.

I felt I needed to use it because of that bit about it acting as the single source of truth until it's blurred, but honestly it was a pain to work with. Maybe I could try to pull off the same behavior using our inputs...I will try.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
wilysword: Alright, to keep the behavior similar to how it is now, you'll probably want to ignore most if not all of the `wrapperProps`.

Dhalton:
> Would it make more sense to put the palette icon on the left, like with the date picker?

Yeah we could do that. I was going off of the design but as the ticket says it doesn't have to be exact. I'll run the idea by the designers and see what they think.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
wilysword: I don't even understand what that means, and [looking at the code](https://github.com/casesandberg/react-color/blob/master/src/components/common/EditableInput.js#L59) it still fires `onChange` events normally. Since browsers only allow a single input to have focus, don't _all_ inputs become the single source of truth while they're focused? The user, at least, can't make any other changes.

wilysword:
Not sure why, but when I change the color by typing in the main input (popup closed): as you said it doesn't fire `onChange` unless the value is valid, but it's firing the event with the _previous_ valid value.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,574 @@
...
+  const handleOuterHexChange = (
+    change: { [k: string]: string },
+    event: React.ChangeEvent<HTMLInputElement>
+  ) => {
+    const hex = change[phrases.hexLabel]
+    if (isValidHex(hex)) {
+      const color = tinycolor(hex)
+      saveColor(color)
+      if (onChange && typeof onChange === 'function') {
+        const cactusEvent = new CactusChangeEvent(eventTarget, event)
```
wilysword, line 418: Here's the problem. The box will update the target value on the next render, but you're raising the event right away so you'll need to manually update the value here.

Dhalton: Thank you for pointing this out, appreciate it. That could've been bad!

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
wilysword: Okay, looking at it a bit more I think all it really means is that it ignores the `value` prop while it's focused, so it's like a Frankenstein controlled & uncontrolled input. I'm not sure that's a good idea, personally, but you can do whatever's easiest.

Dhalton: So I've been experimenting with this and I think we are going to have to leave them as `EditableInput`s. We need that hybrid controlled/uncontrolled behavior. Take a hex input for example: If I try to use a native HTML `input`, there are two ways to go about writing the change handler, but only one would work as far as I can tell:

1. I update the state immediately on change, but then I'm left with an invalid value and no way to update it since every change I try to make just keeps generating invalid color values.
2. I wait for the hex value to be valid before updating the state, but that would mean I still need to ignore the value passed to the input until either the hex value is valid or the input is blurred, otherwise there's no way to type in the input b/c the state isn't updating and the value is therefore not changing.

So I'd still have to implement the exact same behavior that the `EditableInput` does.

wilysword: I think the problem with (1) is that you're trying to use a single state value to be two things: the state of the text input, _and_ the state of the color picker as a whole. Like I said, though, however is easiest for you I'm fine with.

modules/cactus-web/src/ColorPicker/ColorPicker.test.tsx
```diff
@@ -138,6 +129,42 @@ describe('component: ColorPicker', () => {
...
+
+      userEvent.clear(blueInput)
+      userEvent.type(blueInput, '255')
+      fireEvent.blur(blueInput)
+
+      const cancelButton = getByText('Cancel')
+
+      userEvent.click(cancelButton)
+
+      expect(onChange).not.toHaveBeenCalled()
```
mikesoltow, line 164: It might also be a good idea to add an assertion here to verify that the value was reset

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -360,15 +364,11 @@ export const ColorPickerBase: React.FC<ColorPickerProps> = (props) => {
...
 
-  const [state, setState] = useState<Color>({
-    hsl: { h: 0, s: 1, l: 0.5 },
-    hsv: { h: 0, s: 1, v: 1 },
-    rgb: { r: 255, g: 0, b: 0 },
-    hex: 'FF0000',
-  })
-  const [open, setOpen] = useState<boolean>(false)
+  const [state, setState] = useState<Color>(initialState)
+  const [openedState, setOpenedState] = useState<Color>(initialState)
```
mikesoltow, line 371: I wonder if this should be a ref?  Not sure we need to re-render when we change it.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -380,21 +380,25 @@ export const ColorPickerBase: React.FC<ColorPickerProps> = (props) => {
...
-      hexInput.focus()
-    }, 0)
+  }, [value, expanded])
+
+  const handleIconButtonClick = () => {
+    if (!expanded) {
+      // Save state popup was opened with in case we need to reset on cancel
+      setOpenedState(state)
+      const hexInput = document.querySelector('.hex-wrapper')?.querySelector('input') as HTMLElement
+      toggle(true, hexInput)
```
wilysword, line 399: This focus hint isn't working in the story because it's selecting the input from the disabled color picker. You should harden the selector so it'll work even if multiple pickers are on the page, e.g.
```
const hexInput = document.querySelector(`#${popupProps.id} input`)
```

wilysword:
I think you need to manually delete the extra storyshots you generated before.

mikesoltow (commented):
Just a few more minor comments

mikesoltow (changes requested):
Change look good, but found one little styling hiccup.  If you type in CCCCCC into the input field, then the text is wider than the input (see screenshot).  Increasing the input's width from 4.5 to 5 rem seemed to fix it.
Screen Shot 2021-02-11 at 12 43 04 PM

wilysword:
I think we're almost there, but there's still something funky going on with the cancel. To reproduce:

1. Type something in the main hex input, it fires the change event
2. Shift+tab to the icon button, space to activate
3. Use the mouse to pick a random hue/saturation
4. Press the cancel button, and the value goes back to what it was _before_ step one. (Only the value in the input, a change event is not fired.)
5. If in step two you use the mouse to open the picker instead of the keyboard, cancel will restore to the correct value.

Dhalton:
@wilysword There you go saving my butt again...thanks for pointing that out. Should be working now

wilysword (approved):

mikesoltow (commented):
Looking good.  Just one more thing on IE11 (see screenshot).  I think you can get rid of that by ditching the `height: 478px;` line in `ColorPickerPopup`

mikesoltow:
Just realized I didn't actually upload the screenshot.  Here you go.

Screen Shot 2021-02-12 at 1 27 27 PM

mikesoltow (approved):
LGTM

PR #509, opened 2021-02-09
Cactus 474 navbar variants

daniloPenaR:
[Cactus - 474](https://repayonline.atlassian.net/browse/CACTUS-474)


modules/cactus-web/src/MenuBar/MenuBar.story.tsx
```diff
@@ -66,7 +69,7 @@ export const BasicUsage = (): React.ReactElement => {
       }
     }
     return (
-      <Component key={ix} {...props}>
+      <Component key={ix} variant={variantSelection} {...props}>
```
Dhalton, line 72: Can we make this a separate story instead of a knob? So we can get a screenshot of both variants for the storyshot tests

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -40,10 +42,74 @@ interface MenuBarProps {
...
+export type MenuBarVariants = 'light' | 'dark'
+
+type VariantMap = { [K in MenuBarVariants]: FlattenInterpolation<ThemeProps<CactusTheme>> }
+
+const variantMap: VariantMap = {
+  light: css`
+    [role='menubar'] > li > [role='menuitem'] {
+      ${(p) => textStyle(p.theme, 'body')};
+      color: ${(p) => p.theme.colors.darkestContrast};
+      font-weight: 600;
```
Dhalton, line 59: I would put any styles that don't depend on the variant back into the `Nav` that's defined on line 350 under the same css selector. In this case I believe it would be `textStyle`, `font-weight`, `text-transform`, and the `box-shadow` styles.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -40,10 +42,74 @@ interface MenuBarProps {
...
+      text-transform: uppercase;
+
+      background-color: ${(p): string => p.theme.colors.white};
+
+      border: ${(p) => border(p.theme, 'transparent')};
+      border-bottom: 3px solid ${(p) => p.theme.colors.lightContrast};
+
+      &:hover,
+      &[aria-expanded='true'] {
+        box-shadow: ${shadowTypes[2]} hsla(200, 96%, 35%, 0.3);
```
Dhalton, line 69: Ok this color we *do* have defined in the theme with the opacity you need. It's under the `transparentCTA` key and we should definitely use it. The reason being that the way this is set up now, the theme could be using a different color but the box shadow will always show up with the blue CTA color.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -40,10 +42,74 @@ interface MenuBarProps {
...
+      background-color: ${(p): string => p.theme.colors.base};
+      border: ${(p) => border(p.theme, 'transparent')};
+
+      &:focus,
+      &:hover {
+        border: ${(p) => border(p.theme, 'lightgray')};
+      }
+
+      &:hover {
+        box-shadow: ${shadowTypes[2]} hsla(200, 96%, 35%, 0.3);
```
Dhalton, line 96: Same comment about the `transparentCTA` color here

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -257,8 +332,7 @@ const navClickHandler = (event: React.MouseEvent<HTMLElement>) => {
 }
 
 MenuBar.displayName = 'MenuBar'
-MenuBar.propTypes = { 'aria-label': PropTypes.string }
```
Dhalton, line 260: Why are you removing the propTypes from the menu bar here?

Dhalton (changes requested):
Looking good! I just have a few suggestions for you.

wilysword:
This is looking quite a bit better than the old stuff, I have to say. A few comments on where the styles don't match the style guide, though:

- The padding on the menu items is 20x8 instead of 24x16
- The margin-left on the down arrows is 16px instead of 8px
- The "On Page" styles don't appear to be implemented for either variant (the selector for that is `[aria-current='true']`, and there's no story that has that right now, so to test you can make a temporary change to one of the stories)

On the light variant:
- On hover, text stays dark instead of changing to `callToAction`
- On focus, text shifts down 1px, probably because the border is changing size
- border-bottom is always 3px, but it's supposed to be 2px when the theme is set to "thin" and 3px only when the theme is set to "thick"
- The edges of the border are angled, leading to little gaps between the menu items; I'm not certain this will work, but one possible fix would be to move the on-hover box shadow up to the `li` element, and then implement the bottom border as an inset box shadow (see the `insetBorder` helper function)

On the dark variant:
- The scroll buttons need to have their background/foreground colors changed as well (if you don't know what I'm talking about, increase the "Breadth" knob to 15 or so, and the buttons should appear on either side of the screen)
- Hovering over one item increases the height of the entire menubar because the border is changing size

For the border changing size issues, my best suggestion is that if you need just the bottom-border, it may be best to try an inset box shadow instead (though that doesn't always work). Failing that, try to make sure that the border is always the same size and either matches the background color or is transparent, and then just changing the color on hover/focus. If you can't get either of those to work, then you'll have to adjust the padding when the border changes, so the total width/height will stay the same.

Screenshot of a scroll button:
scrollbutton

wilysword:
Here's one more thing I saw with the light variant (in the Layout story): the light contrast border needs to go all the way from side to side, not just under the menu items. There again, it'd probably be best to use an inset box shadow if you can.
menubar-border

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -464,7 +555,7 @@ const MenuButton = styled.span.attrs({ tabIndex: -1 as number, role: 'menuitem'
   width: 100%;
   height: 100%;
 
-  ${NavigationChevronDown} {
+  ${NavigationArrowDown} {
     width: 8px;
     height: 8px;
     margin-left: 16px;
```
Dhalton, line 561: This `margin-left` needs to be changed to 8px

Dhalton:
Looking good, but I'm still seeing the issue with the menu changing size because of the border on the dark variant. I played around with the `insetBorder` function but couldn't get that to work. Did you try Glen's other suggestion to make the border the same color/transparent and change the color on hover? I have a feeling that would work.

daniloPenaR:
@Dhalton It works, but I'm facing a new inconvenient. Now when I focus a list Item this happens, and if I drop the border bottom, it will start changing the size
Screen Shot 2021-02-11 at 5 13 01 PM

Dhalton:
> @Dhalton It works, but I'm facing a new inconvenient. Now when I focus a list Item this happens, and if I drop the border bottom, it will start changing the size

Ah, I see. Well if that's the case, then why don't we adjust the padding when it's hovered? Seems like that might be the last option.

wilysword:
I think we can get it to work with the border, I'll take a look tomorrow.

wilysword:
Okay, it turns out you were right about having to adjust the padding (I'd forgotten that focus & hover had different border sizes). I did notice that what you were doing with the ScrollButton was interfering with the disabled styles, so I made a tweak to that component that makes it easier to style. I also made the same modification to the `border` helper as we made earlier to the `insetBorder` helper. Here's my suggestions (you can pick the whole commit, or just some of the changes): https://github.com/repaygithub/cactus/commit/9b82be0534f17ce3b2a36008360a2cd6d4a4900c

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -273,34 +352,32 @@ export { TypedMenuBar as MenuBar, MenuBarList, MenuBarItem }
...
-      color: ${(p) => p.theme.colors.callToAction};
-      border-bottom-color: ${(p) => p.theme.colors.callToAction};
-    }
-    &:focus {
-      border: ${(p) => border(p.theme, 'callToAction')};
-    }
-    &[aria-current='true'] {
-      font-weight: bold;
+      z-index: 100;
+      box-shadow: ${shadowTypes[2]} ${(p) => p.theme.colors.transparentCTA};
```
wilysword, line 377: What you're doing here is equivalent to the `boxShadow` helper that already exists.
```suggestion
      ${(p) => boxShadow(p.theme, 2)};
```

wilysword (commented):
Looking good, just a couple of minor things left. On my comments about the scroll buttons, this is what it looks like in the submenus:
scrollbutton-underline

Dhalton (approved):
LGTM, nice work

wilysword (approved):

PR #511, opened 2021-02-11
docs(cactus): storybook documentation 1.0

daniloPenaR:
Link: [Cactus-506](https://repayonline.atlassian.net/browse/CACTUS-506)

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,4 +236,6 @@ These are other use-cases where you must set up a knob for a prop from the paren
 
 Knobs are a powerful way to explore the component behavior through their properties. However, there are specific cases where it is more convenient to show those behaviors by splitting them into different stories. In the main `Basic usage` story, it is not necessary to include knobs for all of the props.  Here are some cases that you would consider creating a new story instead of using knobs: if the component has different variants, (such standard, warning, success), a better choice could be to create a new story showing all the variants on the same canvas, so the user could compare them. Another case where you must consider creating a new story is if the props modify the component by making significant changes to its appearance, or if the component has specific optional functionalities or integrations with other small components like buttons or icons.
 
-If you want to look for more information about how to implement the knobs addon, please visit the [documentation](https://storybook.js.org/docs/react/writing-stories/introduction).
+#### Usage
+
+If you want to look for more information about how to write stories, please visit the [documentation](https://storybook.js.org/docs/react/writing-stories/introduction).
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ npm install --save-dev @repay/cactus-fwk
 
 **Before contributing, please read our [guidelines for contributing](./CONTRIBUTING.md)**
 
-You will need to install [Node.js](https://nodejs.org/en/) runtime and [Yarn](https://yarnpkg.com/en/docs/install) for package management. Next clone the repository and install the dependencies.  **Be sure to do this from the repository root.**
+You will need to install [Node.js](https://nodejs.org/en/) runtime and [Yarn](https://yarnpkg.com/en/docs/install) for package management. Next clone the repository and install the dependencies. **Be sure to do this from the repository root.**
 
 ```
 yarn install
@@ -166,7 +166,7 @@ To publish a release, open Terminal or command prompt and call:
 yarn release
 ```
 
-This command will install all dependencies, build packages, and then call [Lerna](https://lerna.js.org/) to handle the release process.  Lerna will automatically determine which packages have changed and will prompt you for the type of version bump (major/minor/patch).   It will then publish the updated packages, create commits with `package.json` changes, and create git tags.
+This command will install all dependencies, build packages, and then call [Lerna](https://lerna.js.org/) to handle the release process. Lerna will automatically determine which packages have changed and will prompt you for the type of version bump (major/minor/patch). It will then publish the updated packages, create commits with `package.json` changes, and create git tags.
 
 We also have a `yarn release:beta` command that will do roughly the same thing, but will pass additional options to Lerna that are needed for beta versions/pre-releases.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -193,3 +193,103 @@ will send them to you via onetimesecret.
...
+- **`Text:`** To get a text from the user.
+- **`Boolean:`** To get a boolean value from the user.
+- **`Number`:** To get a number from the user.
+- **`Select`:** To get a value from a list of options. 
+- **`Radios`:** To get a value from a list of radio buttons.
+
+[More info here.](https://github.com/storybookjs/storybook/tree/master/addons/knobs#available-knobs)
+
+The knobs must have been created to give the user access to the different props of a component. For example, a button component that includes a `variant` prop with a list of variants like *standard*, *action*, *danger*, *warning* and *success*, those variants must be accessible and controllable from the Knobs section of the button's story.
+These are other use-cases where you must set up a knob for a prop from the parent component:
```
wilysword, line 228: In Markdown, you need blank lines to separate paragraphs, so you need a blank line before and after this line. (See https://github.com/repaygithub/cactus/blob/f7f7235d63b200588e180c5cb0d908b386b6e156/README.md for what it currently looks like.)

README.md
```diff
@@ -193,3 +193,103 @@ will send them to you via onetimesecret.
...
+– To Modify text from labels, inputs, textareas, headers and/or other contents.
+– To enable or disable user's actions like clicks. 
+– To change a specific CSS property like `justify-content`, `align-items`, `flex-direction`, `width`, `height`
+– To hide or show a specific component from the story.
```
wilysword, line 232: Looks like you're using dashes here instead of hyphens.
```suggestion
- To Modify text from labels, inputs, textareas, headers and/or other contents.
- To enable or disable user's actions like clicks. 
- To change a specific CSS property like `justify-content`, `align-items`, `flex-direction`, `width`, `height`
- To hide or show a specific component from the story. 
```

README.md
```diff
@@ -193,3 +193,103 @@ will send them to you via onetimesecret.
...
+[More info here.](https://github.com/storybookjs/storybook/tree/master/addons/knobs#available-knobs)
+
+The knobs must have been created to give the user access to the different props of a component. For example, a button component that includes a `variant` prop with a list of variants like *standard*, *action*, *danger*, *warning* and *success*, those variants must be accessible and controllable from the Knobs section of the button's story.
+These are other use-cases where you must set up a knob for a prop from the parent component:
+– To Modify text from labels, inputs, textareas, headers and/or other contents.
+– To enable or disable user's actions like clicks. 
+– To change a specific CSS property like `justify-content`, `align-items`, `flex-direction`, `width`, `height`
+– To hide or show a specific component from the story. 
+
+##### Usage
```
wilysword, line 234: I'm not sure about this section. It's useful information, but if the target audience is future developers I think they could probably figure this out just by looking at some of the existing stories. Since you already wrote it I would normally say we should just keep it; but if there's ever another change as significant as the recent major version bump, we'd have to maintain this documentation and I'm not sure it's worth it. I'd like to hear some other opinions on this one before you make any changes, though.

README.md
```diff
@@ -193,3 +193,103 @@ will send them to you via onetimesecret.
...
+
+#### Actions
+The Actions add-on logs the events triggered by the user, such as onClick, onFocus, or onBlur.
+
+#### Knobs
+
+The [knobs](https://www.npmjs.com/package/@storybook/addon-knobs) are controls that allow the user to control and edit the component’s props dynamically, exploring the component behavior. These are the knobs that are available: 
+
+- **`Text:`** To get a text from the user.
+- **`Boolean:`** To get a boolean value from the user.
```
wilysword, line 220: Put the colon outside the backticks so it matches the last three items in the list.
```suggestion
- **`Text`:** To get a text from the user.
- **`Boolean`:** To get a boolean value from the user.
```

wilysword (commented):
Looks good. I think it's still missing a section, though, perhaps after the section on knobs: it needs to say something about when to make a knob vs when to make a completely different story. (I'd say stories should be split if (a) there are too many props/knobs; or (b) if changing a prop results in a significant visual difference. That's just a guideline, though, and some smaller components you can fit multiple variants in a single story side-by-side.)

README.md
```diff
@@ -216,20 +216,23 @@ The Actions add-on logs the events triggered by the user, such as onClick, onFoc
...
+- To hide or show a specific component from the story. 
+
+Knobs are a powerful way to explore the component behavior through their properties. However, there are specific cases where is more convenient to show those behaviors by splitting them into different stories. The main story that you must include is the components `Basic usage`, it is not necessary to include all the knobs per prop into that main story, is there when you can decide when to make a knob for a prop or when to make a new different story to show a specific behavior, there are some cases that you would consider at the moment of being choosing between a knob or a new story; if the component has different variants, (such standard, warning, success), a better choice could be to create a new story showing all the variants on the same canvas, so the user could compare them. Another case where you must consider creating a new story is if the props modifies the component by making significant changes to its appearance, or if the component has specific optional functionalities or integrations with other small components like buttons or icons.
```
wilysword, line 235: Just add a header to this section, e.g. "Knobs vs. Stories", and I think we'll call it good.

README.md
```diff
@@ -193,3 +193,106 @@ will send them to you via onetimesecret.
...
```
mikesoltow, line 235: Some more rewording suggestions:

```suggestion
Knobs are a powerful way to explore the component behavior through their properties. However, there are specific cases where it is more convenient to show those behaviors by splitting them into different stories. In the main `Basic usage` story, it is not necessary to include knobs for all of the props.  Here are some cases that you would consider creating a new story instead of using knobs: if the component has different variants, (such standard, warning, success), a better choice could be to create a new story showing all the variants on the same canvas, so the user could compare them. Another case where you must consider creating a new story is if the props modify the component by making significant changes to its appearance, or if the component has specific optional functionalities or integrations with other small components like buttons or icons. 
```

README.md
mikesoltow: Haha...I actually steered Danilo in this direction, but now that you mention it, I take your point about the maintenance.  Maybe we could replace this with a link to Storybooks docs, i.e. https://storybook.js.org/docs/react/writing-stories/introduction

Sorry, Danilo!

mikesoltow (commented):
This looks great!  I mostly just have a few suggestions for rewording things.  But the substance of these docs is excellent.

wilysword (approved):

README.md
```diff
@@ -224,75 +224,16 @@ The [knobs](https://www.npmjs.com/package/@storybook/addon-knobs) are controls t
...
-    } as Meta
-
-    export const FooStory = () => {
-        return <FooComponent 
-            label={text('label', 'This is a knob')} 
-            disabled={boolean('disabled', false)} 
-        /> 
-    }
-```
+If you want to look for more information about how to implement the knobs addon, please visit the [documentation](https://storybook.js.org/docs/react/writing-stories/introduction).
```
mikesoltow, line 239: I would just make the following change, and then I think we're good:

```suggestion
#### Usage

If you want to look for more information about how to write stories, please visit the [documentation](https://storybook.js.org/docs/react/writing-stories/introduction).
```

mikesoltow (commented):
Looks great!  Just one more little comment, then I'm ready to 👍

mikesoltow (approved):

PR #512, opened 2021-02-11
CACTUS-503: added how-to guide for ETE testing

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-503

I think one or two links in this will suffer from the same problem as https://repayonline.atlassian.net/browse/CACTUS-505. I did play around with a couple things to see if I could fix it, but since there's already a ticket for that I didn't want to spend too long on it.


mikesoltow (approved):
This looks great!

Dhalton (approved):

PR #514, opened 2021-02-11
CACTUS-509 :: DateInput Transition

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-509

It works in IE11 now but to make it work I had to slow the transition down a bit and had to pass a longer `timeout` to `CSSTransition` than the actual transition itself takes.

It looks _decent_ in IE11, but obviously it's not nearly as smooth as more modern browsers. Part of it is browserstack (I tested on my windows machine at home and it's ever-so-slightly better than what you see in browserstack), but also I think animations just suck in IE. It just seems a lot less responsive.

If you guys would rather me just remove the functionality in IE altogether just let me know. I wanted to give you guys a chance to see it for yourselves, though.


wilysword (approved):
That is annoyingly slow, but it didn't seem much slower than any other interaction on the component. Personally I don't think we need the sliding effect at all, but then I'm not much for that kind of eye candy in general. It'd probably be easier to either have it for all browsers or have it for none.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -1779,78 +1819,44 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
...
-                            </CalendarDay>
-                          )
-                        )}
-                      </div>
-                    </Calendar>
+                  <CalendarSlideWrapper>
+                    <TransitionGroup className="calendar-transition-group">
+                      <CSSTransition
+                        key={this.state.calendarKey}
+                        timeout={isIE ? 800 : 300}
```
mikesoltow, line 1826: For non-IE browsers, I'd lower this to `200` or so.  At `300`, I get this weird effect where multiple calendars stack on top of each other if I click through the dates really fast.

mikesoltow (commented):
Nice work.  I'm fine keeping it in IE, as the current interaction isn't snappy either, but I think you can lower the timeouts a bit.  I was able to lower the `transitionTime` to `350ms` and the `timeout` to `400`.  I also have one comment about the non-IE timeout.

mikesoltow (changes requested):
So I was just about to approve this, but then I noticed something: once you use the side arrows a couple of times, you can't use the month/year drop-downs because the calendar will turn blank when you make your selection.  This doesn't happen if you use the drop-down right away...only if you use the side arrows first.  I have to hop off in a minute, so I can't dig into it right now, but I'm happy to help you look tomorrow if you'd like.

Dhalton:
@mikesoltow huh, that's weird. I'll investigate tomorrow. Thanks for the heads up

mikesoltow (approved):
Looks good now...thanks!

PR #518, opened 2021-02-15
CACTUS-502 :: Dependency Updates

mikesoltow:
The main changes of consequence were an update to Reach and updates to the build script.  To test, I'd do the following:

#### Reach Updates

Test the following components, ideally in more than one browser (hint hint IE11):

* Layout - You can kill several birds with one stone here...not only do you have the MenuBar but you have MenuButton and SplitButton in the ActionBar Panel.
* Accordion
* BrandBar
* DateInput
* MenuBar
* MenuButton
* Modal
* SplitButton
* Tooltip

#### Build Script Updates

Run `yarn cleanup && yarn build`.  Then run `yarn w mock-ebpp start`.  Check out several pages just to make sure everything works.  Again, testing in more than one browser to be safe would be a good idea.


wilysword:
Not sure what the deal is, but when I fired up `yarn docs dev` and tried to go to one of the docs pages I get a 404. The strange part is that the 404 page has a list of "valid" URLs, and the URL I'm currently on is on the list. Not sure if I did something wrong in install/build or what.
gatsby404

I didn't have any problems testing in Storybook.

mikesoltow:
> Not sure what the deal is, but when I fired up `yarn docs dev` and tried to go to one of the docs pages I get a 404. The strange part is that the 404 page has a list of "valid" URLs, and the URL I'm currently on is on the list. Not sure if I did something wrong in install/build or what.

Try running `rm -rf node_modules && yarn && yarn cleanup && yarn build && yarn docs start`.  I initially was reproducing your issue, but after going nuclear, it seemed to work.

mikesoltow:
@wilysword Actually I was still able to reproduce it on another page.  If you refresh the page, it seems to work, but when you first go to it, it doesn't seem to work.  I'll do a test deployment in the morning, as if the deployed version just works, then I won't sweat it too much.

wilysword (approved):

daniloPenaR:
Screen Shot 2021-02-18 at 11 24 12 AM
That's happening when I test the modal on storybook from IE11 (browserstack)

daniloPenaR:
> Screen Shot 2021-02-18 at 11 24 12 AM
> 
> That's happening when I test the modal on storybook from IE11 (browserstack)

I made a quick inspection on the element, and seems that the problem is with a `margin:auto;`, I'm not sure if that was a change made by me lol, but I think we could change or drop that property.

mikesoltow:
> That's happening when I test the modal on storybook from IE11 (browserstack)

Good catch, but that appears to be an existing bug.  It happens in the deployed storybook as well (https://repaygithub.github.io/cactus/stories/cactus-web/?path=/story/modal--basic-usage).  I did notice that earlier and created a ticket (https://repayonline.atlassian.net/browse/CACTUS-518).  While it should be a quick change, I'd probably prefer to keep it separate just so we don't need to do another round of cross-browser testing for this PR.

daniloPenaR (approved):

PR #519, opened 2021-02-17
CACTUS-512 :: Grayscale Theme

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-512

This was pretty straightforward, but let me know if I missed anything.

### DIFF
--- a/modules/cactus-theme/tests/theme.test.ts
+++ b/modules/cactus-theme/tests/theme.test.ts
@@ -23,7 +23,7 @@ function themeAccessibility(themeName: string, theme: CactusTheme): void {
       ).toBeGreaterThanOrEqual(4.5)
     })
 
-    Object.keys(theme.colorStyles).forEach((name: keyof typeof theme['colorStyles']): void => {
+    Object.keys(theme.colorStyles).forEach((name): void => {
       if (name.includes('transparent')) {
         test(`colorStyle ${name}`, (): void => {
           const color = new Color(theme.colorStyles[name].backgroundColor)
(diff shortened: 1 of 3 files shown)


modules/cactus-theme/tests/theme.test.ts
```diff
@@ -54,18 +54,43 @@ describe('@repay/cactus-theme', (): void => {
   themeAccessibility('cactus theme', cactusTheme)
 
   test('generates a theme with primaryHue', (): void => {
-    const theme = generateTheme({ primaryHue: 100 })
+    const theme = generateTheme({ primaryHue: 200 })
     expect(theme.colors).toMatchObject({
-      base: 'hsl(100, 96%, 11%)',
-      callToAction: 'hsl(100, 96%, 35%)',
+      base: 'hsl(200, 96%, 11%)',
+      callToAction: 'hsl(200, 96%, 35%)',
```
daniloPenaR, line 60: Just out of curiosity, Why did you have to change `primaryHue` from 100 to 200?

daniloPenaR (approved):

modules/cactus-theme/tests/theme.test.ts
mikesoltow: Good question!  It's because using a hue of 100 actually makes some combos not accessible.  We weren't testing accessibility on this setup before, so it didn't matter.  However, I added accessibility tests for this case, both for the non-grayscale and grayscale cases, so that we could make sure that grayscale wasn't messing with accessibility in any way.

Dhalton (approved):
Nice, LGTM!

PR #521, opened 2021-02-18
CACTUS-514 How-to Formik + Cactus

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-514

So according to the Divio nomenclature, I think this came out as more of a reference guide than a how-to guide. How-tos are supposed to have numbered steps, basically, and the only steps I could come up with were (1) install Formik and (2) write a form; both of which are covered quite well already in the Formik docs. So instead I mostly covered how to combine Formik with our form elements, and any quirks or unexpected surprises I found while doing so.


Dhalton (approved):
Really well-written and clear, nice!

daniloPenaR (commented):
Well explained! great information. 
The examples are very appropriated. 
Looks good to me.

daniloPenaR (approved):

PR #524, opened 2021-02-19
CACTUS-508 :: Update Jest Sonar Package

mikesoltow:
Blackduck listed `jest-sonar-reporter` as a high Operational Risk since it is now inactive.  This PR replaces it with `jest-sonar`.

There's not much to test.  Just make sure the tests still run and that `test-report.xml` is still being generated in the various module folders.

### DIFF
--- a/modules/cactus-fwk/package.json
+++ b/modules/cactus-fwk/package.json
@@ -41,7 +41,7 @@
     "collectCoverageFrom": [
       "src/**/*.{ts,tsx}"
     ],
-    "testResultsProcessor": "jest-sonar-reporter"
+    "reporters": ["default", ["jest-sonar", { "outputDirectory": ".", "outputName": "test-report.xml" }]]
   },
   "devDependencies": {
     "@babel/core": "^7.12.16",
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -4,6 +4,7 @@
 **/dist
 !actions/create-pull-request/dist
 **/coverage
+**/sonar-report.xml
```
wilysword, line 7: Do you still need to have added that, since it looks like you're changing the file name in the config?

wilysword (approved):
One question, but it looks to me to be working.

daniloPenaR (approved):

mikesoltow:
@daniloPenaR and @wilysword : would you each mind re-reviewing?  I made the following changes:
* `yarn.lock` got borked in my first commit somehow.  I de-borked it, but to make everything cleaner I amended the original commit (so you'll need to `get reset --hard origin/cactus-508-blackduck` from your local copy of this branch instead of `git pull`)
* `lodash` released a fix over the weekend for a security vulnerability Blackduck had flagged, so I pushed up a commit for that as well.

wilysword (approved):

daniloPenaR (approved):

PR #527, opened 2021-02-22
Cactus 505:  Gatsby Links

daniloPenaR:
(CACTUS-505)[https://repayonline.atlassian.net/browse/CACTUS-505?atlOrigin=eyJpIjoiNDIxZDE2NWViYzdlNDJkYTkxMzg0OGMwNWEzY2RmZDUiLCJwIjoiaiJ9]

### DIFF
--- a/docs/Components/API Documentation.md
+++ b/docs/Components/API Documentation.md
@@ -24,7 +24,7 @@ The spaces are defined as `[0, 2, 4, 8, 16, 24, 32, 40]`, this may change, but i
 - `4` (`=16px`) vertical spacing for a single hierarchy change such as between title and content to different form groups
 - `5`, `6`, `7` are used to further indicate hierarchy or context changes
 
-Additional information about how to use space can be found in the [Shared Styles](https://repaygithub.github.io/cactus/design-system/shared-styles/) section of the Cactus Design System.
+Additional information about how to use space can be found in the [Shared Styles](https://www.figma.com/proto/J9TeJ5homSkfSwV1sCFLoF/REPAY-STYLEGUIDE-V-1?scaling=min-zoom&node-id=2623%3A0) section of the Cactus Design System.
 
 ### Variants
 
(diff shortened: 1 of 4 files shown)


wilysword (commented):
Everything seems to be working locally, but will this fix the problem in the public/published docs? @mikesoltow is there a way to test that? For example, the first two links on https://repaygithub.github.io/cactus/components/ are broken.

mikesoltow:
> Everything seems to be working locally, but will this fix the problem in the public/published docs? @mikesoltow is there a way to test that? For example, the first two links on https://repaygithub.github.io/cactus/components/ are broken.

Feel free to run `yarn cleanup && yarn build && yarn docs publish`.  I recently updated the docs from `master`, so you wouldn't be seeing Danilo's code currently.  I'm fine with you publishing this branch as long as you go back and re-publish from `master` when you're done (so Emily can still UAT).

daniloPenaR:
> Everything seems to be working locally, but will this fix the problem in the public/published docs? @mikesoltow is there a way to test that? For example, the first two links on https://repaygithub.github.io/cactus/components/ are broken.

@wilysword  Are you referring to `Theming` and `API Documentation`? Because I have just proved those links and It seems to work. I'll try cleaning all the cache data of my browser.

wilysword (approved):
Despite some weirdness with my browser, this seems to check out. Very good job.

mikesoltow (changes requested):
This is working great!  I just found one link that slipped through the cracks: click on Cactus Web > API Docs and then click on the "@repay/cactus-theme" link.

daniloPenaR:
> This is working great! I just found one link that slipped through the cracks: click on Cactus Web > API Docs and then click on the "@repay/cactus-theme" link.

🤔 I thought it was working, going to see what's happening.

mikesoltow (approved):
Nice!  Great job figuring this out.  I think at some point we may want to migrate all links to this method, since it seems to be better supported by Gatsby and more predictable, but we can do that later.

PR #530, opened 2021-02-22
CACTUS-525: added color styles to Tag component

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-525

Technically I just needed to set the default `standard` color style, but who knows what the future holds, and what other use cases may pop up? Since it was just as easy to allow any color style, I figured why not.

### DIFF
--- a/modules/cactus-web/src/SelectField/SelectField.story.tsx
+++ b/modules/cactus-web/src/SelectField/SelectField.story.tsx
@@ -29,6 +29,7 @@ export const BasicUsage = (): React.ReactElement => (
       warning={text('warning', '')}
       error={text('error', '')}
       autoTooltip={boolean('autoTooltip', true)}
+      multiple={boolean('multiple', false)}
       comboBox={boolean('comboBox', false)}
       canCreateOption={boolean('canCreateOption', true)}
       disableTooltip={select('disableTooltip', [false, true, undefined], false)}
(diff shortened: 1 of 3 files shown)


Dhalton (approved):
LGTM

mikesoltow (commented):
Looks good.  Only suggestion I have is to add a colorStyle knob on the Tag story, as that's the easiest way for Emily to UAT the new prop.

wilysword:
As I was thinking about it more, I think for it to be really usable we'd need to modify the border color as well; would you rather take it out, or leave it kind of incomplete? (Or I guess a third option would be to come up with border colors, probably based on the Alert component or something.)

mikesoltow:
> As I was thinking about it more, I think for it to be really usable we'd need to modify the border color as well; would you rather take it out, or leave it kind of incomplete? (Or I guess a third option would be to come up with border colors, probably based on the Alert component or something.)

🤔  Curious to get your take, @Dhalton , but I think that's a good point.  The borders don't actually look terrible to me...but that's probably because they're so tiny.  To be good citizens, we should probably make them legit.  I'd lean toward taking this out for now, but I don't feel very strongly.  It would be easy enough to add with designer input if we need it later.

Dhalton:
Yeah, good point about borders. I guess my two cents would be to remove it since we don't have the designs to fully support that feature. Like Mike said, it'd be easy to add it back if people are asking for it later on. I like that you were planning ahead, though!

Dhalton:
My approval still stands. 👍

mikesoltow (approved):

PR #532, opened 2021-02-24
Cactus-507 :: SonarQube Fixes

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-507

Pretty straightforward...just fixed a few code smells and added some tests for some files we were missing coverage for.


daniloPenaR (approved):

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -1336,8 +1336,13 @@ class SelectBase extends React.Component<SelectProps, SelectState> {
       } else if (child?.props) {
         const value = child.props.value
         if (typeof value === 'string' || typeof value === 'number') {
-          const { altText: altProp, children, id, 'aria-label': ariaLabel } = child.props
-          const rawLabel = children || children === 0 ? children : value
+          const {
+            altText: altProp,
+            children: childrenOfChild,
```
wilysword, line 1341: Most people call those "grandchildren" 😉.

wilysword (approved):

modules/cactus-web/src/Select/Select.tsx
Dhalton: Lol funny that you point this out, because I sat for a while trying to come up with a name that doesn't sound stupid. Grandchildren 🤦 I honestly might just update that because I hate this name.

Dhalton: Updated. I swear I just make things needlessly complicated sometimes...

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -16,7 +16,8 @@ class Dimmer extends React.Component<DimmerProps> {
   componentDidUpdate(prevProps: DimmerProps): void {
     if (prevProps.active === false && this.props.active === true && document.activeElement) {
       try {
-        ;(document.activeElement as HTMLElement).blur()
+        // eslint-disable-next-line
+        (document.activeElement as HTMLElement).blur()
```
mikesoltow, line 20: Do we actually want to leave the semi-colon in?  See https://github.com/prettier/prettier/issues/6247

I know SonarQube complains, but we can always mark these in SonarQube as false positives.

PR #533, opened 2021-02-24
CACTUS-527: Change font size on TextInput

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-527

All the storyshots changed because the previous height (32) was actually not correct according to the style guide, it should have been 35. Instead of hardcoding the heights, I added padding.


daniloPenaR (approved):
Nice approach to solve this. Thanks.

Dhalton (approved):
Works great, LGTM!

PR #536, opened 2021-02-26
Cactus 530 grid blowout

daniloPenaR:
[CACTUS-530](https://repayonline.atlassian.net/browse/CACTUS-530)


mikesoltow (changes requested):
This looks good, but would you mind adding a story under `Grid` that reproduces this issue (thus demonstrating to Emily/our screenshots that it's fixed)?  Thanks!

daniloPenaR:
@mikesoltow  I added the new story, `Prevent Grid Blowout`, however at this moment I'm having troubles with the tests; it seems like there is a `DateInput` test that is failing. But I'm not sure why, because I didn't work with that component as well.

mikesoltow:
@daniloPenaR I'm not sure why the test started failing, but it's failing in `master` too.  In fact, looking at the test, it looks like it _should_ have been failing all along...so I'm not sure why it was succeeding before.  In any case, if you want to sneak a little fix in, change line 323 in `DateInput.text.tsx` to `fireEvent.click(within(getByLabelText('Select a month')).getByText('March'))`.  You'll have to add `within` to the `import { ... } from '@testing-library/react` statement at the top.

wilysword (approved):

mikesoltow (approved):
🎉

PR #537, opened 2021-03-02
Cactus 532: Divider Component

daniloPenaR:
[CACTUS-532](https://repayonline.atlassian.net/browse/CACTUS-532)

This component is pretty simple. If I missed something, please let me know.


modules/cactus-web/src/Divider/Divider.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components'
+
+import { border } from '../helpers/theme'
+
+export const Divider = styled.hr`
```
Dhalton, line 5: Something I just though of: We support the margin props on pretty much all of our components, so should we also add support for them here? I could see that being useful. Looking for opinions on this, though. What do you guys think?

Dhalton (commented):
LGTM but I just have a question I wanted to ask the rest of the team before I approve

modules/cactus-web/src/Divider/Divider.tsx
mikesoltow: I'm not opposed to adding that in.  I think that, in general, I would prefer that the width be constrained by the parent's padding than the divider's margin...but it doesn't hurt to add it in just in case.

modules/cactus-web/src/Divider/Divider.tsx
```diff
@@ -0,0 +1,9 @@
...
+  border: ${(p) => border(p.theme, 'lightContrast')};
```
mikesoltow, line 6: This line looked thicker than 1px to me, so I did some digging and found this: https://stackoverflow.com/questions/4151743/how-can-i-change-the-thickness-of-my-hr-tag

Basically it recommends changing this to the following.

```suggestion
  border: none;
  height: ${(p) => p.border === 'thin' ? '1px' : '2px'};
  color: ${(p) => p.theme.colors.lightContrast};
  background-color: ${(p) => p.theme.colors.lightContrast};
```

The attached screenshot shows both the new styles (top divider) and the styles as you have them (bottom divider).  Notice how the new styles are thinner.

Screen Shot 2021-03-02 at 3 28 16 PM

mikesoltow (changes requested):

Dhalton (approved):
LGTM, nice work!

mikesoltow (approved):
LGTM

PR #540, opened 2021-03-02
Change order of visual/integration tests

mikesoltow:
What ⬆️  says

### DIFF
--- a/examples/mock-ebpp/tests/ui-config-integration.test.ts
+++ b/examples/mock-ebpp/tests/ui-config-integration.test.ts
@@ -44,8 +44,8 @@ test('should fill out and submit the entire form', async (t): Promise<void> => {
   await t.click(queryByText('Use Cactus Styles'))
   await fillTextField('month', '2')
   await fillTextField('day of month', '28')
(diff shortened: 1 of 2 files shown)


wilysword (approved):

Dhalton (approved):

examples/mock-ebpp/tests/ui-config-integration.test.ts
```diff
@@ -45,7 +45,7 @@ test('should fill out and submit the entire form', async (t): Promise<void> => {
   await fillTextField('month', '2')
   await fillTextField('day of month', '28')
   await fillTextField('year', '2019')
-  await t.expect(getByLabelText('year').value).eql('2019')
+  await t.expect(queryByLabelText('year').value).eql('2019')
```
wilysword, line 48: I don't think this is the problem, I think it's that it's not focusing quick enough on the previous line, and only the '019' is actually registering. I tried a different fix (https://github.com/repaygithub/cactus/pull/541/commits/67054bc254ed2cb77127e26281fb44fce2674696) of adding another delay, so we can see how they go, I guess.

PR #547, opened 2021-03-09
CACTUS-547 :: Improve Link Contrast

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-547

I added a new `Breadcrumb.Active` component to render non-clickable text representing the current page.  This maintains the current `as` behavior of `Breadcrumb.Item` (without forcing users to do `as={Text}` for the current page).  Plus, if the user wants the current page link to be clickable for some reason, then they can do what they do today and use `<Breadcrumb.Item active>`.  The rendered output for that is unchanged.


Dhalton (approved):
🚀

PR #548, opened 2021-03-11
CACTUS-552 :: Multiselect Order

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-552

### Testing
1. Bring up storybook and go to the Select/With Multiselect story
2. Clear out all of the values in the multiselect
3. Select options in reverse order compared to the order they're listed in the options. For example, select "Bisbee", then select "Avondale".
4. Make sure the options are displayed in the order you chose them, regardless of what order they appear in the options list.
5. Go through each of the other Select stories and make sure you're still able to select values for a normal select, a combobox, and multiselect with combobox. Make sure the order is maintained for multiselect/combobox when you create new options as well.


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -117,18 +117,17 @@ function willTruncateBlockShow(
...
 const ValueSwitch = (props: {
-  options: ExtendedOptionType[]
+  selected: ExtendedOptionType[]
   placeholder: string | undefined
   extraLabel: string
   multiple?: boolean
 }): React.ReactElement => {
-  const selected = props.options.filter((o): boolean => o.isSelected)
-  const numSelected = selected.length
+  const numSelected = props.selected.length
```
mikesoltow, line 125: I don't think you need to call `props.selected` everywhere, right?  You destructured it on line 120.

mikesoltow (approved):
Looks good to me!  Just one minor nit you may want to pick.

modules/cactus-web/src/Select/Select.tsx
Dhalton: Sure thing!

Dhalton: I actually didn't destructure it, that's just the type definition for the `ValueSwitch` props. But it's being destructured now 😄

mikesoltow: Haha...my bad.  I think the diff was clipped 😄

Dhalton: Np!

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -1427,6 +1427,26 @@ class SelectBase extends React.Component<SelectProps, SelectState> {
...
+        }
+      })
+    } else {
+      const option = selected.find((option) => option.value === value)
+      if (option) {
+        optionsInOrder.push(option)
+      }
+    }
+    return optionsInOrder
+  }
```
wilysword, line 1448: I think this whole function could be simpler; there should be a variable called `this.optionsMap` that's a value-to-extended-option mapping, so I think you could just do something like this (for an array value):
```
return value.map((val) => this.optionsMap.get(val))
```

Dhalton: 🤯

Dhalton: If only it were that simple...TypeScript is complaining that the result from `this.optionsMap.get()` could be `undefined`. Trying to figure out how I can tell TS that `val` is known to be a key of the options map.

wilysword: You might want to filter the results anyway, since people can pass in `value`s that aren't valid options.

Dhalton: Ah, true. Yeah I will just have to work around it. It should still clean things up a bit.

wilysword: I also just learned some versions of Typescript have an operator for asserting truthiness, so if you're still having problems maybe it could help:
```
const x: number | undefined = 0
// The exclamation point tells the compiler you know the value isn't undefined.
const y: number = x!
```

Dhalton: Alright I just pushed up a change, let me know if that's more along the lines of what you're thinking.

Dhalton: Huh, I didn't know about that little TS feature. I'll have to keep that in mind for future reference.

wilysword (approved):
It might be good to change the `defaultMultiValue` in `Select.story.tsx` so that this will show up in storyshots. Certainly not a requirement, though.

PR #549, opened 2021-03-15
CACTUS-551 :: Redundant onChange in Select

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-551

The code change was pretty straightforward.  I added a few tests as well to increase our coverage around these edge cases.

As I was tinkering with the tests, I noticed that we were calling `rerender` a lot in that file when we didn't need to, so I decided to quickly pull out all of those calls (a couple were actually useful, so I didn't take out _all_ of them).  I put those changes in a separate commit, so if you want to see my changes without that refactor, just look at the diff from my second commit.


daniloPenaR (approved):
I tested it on my browser and seems to work well.

Dhalton (approved):
LGTM

PR #550, opened 2021-03-15
CACTUS-544 :: Notification Component

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-544

### Testing
Open the Notification "Basic Usage" story and make sure the alert is positioned where it's supposed to be when you change the values of the knobs.


daniloPenaR (approved):
Nice work

modules/cactus-web/src/Notification/Notification.story.tsx
```diff
@@ -0,0 +1,86 @@
...
+          Message
+        </Alert>
+      </Notification>
+    </>
+  )
+}
+
+BasicUsage.parameters = {
+  beforeScreenshot: async (page: Page) => {
+    await page.click('button[id="open-notification"]')
```
wilysword, line 44: I don't know if the ID selector is actually more efficient, but 🤷.
```suggestion
    await page.click('#open-notification')
```

modules/cactus-web/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { ScreenSize, SIZES, useScreenSize } from '../ScreenSizeProvider/ScreenSizeProvider'
+
+interface Position {
+  vertical: 'top' | 'bottom'
+  horizontal: 'left' | 'center' | 'right'
+}
+
+interface NotificationProps {
+  open: boolean
+  position?: Position
```
wilysword, line 14: What's the reasoning for having this be an object instead of two separate props? Personally, I think `<X vertical="top" horizontal="left" />` is both easier to read and looks better than `<X position={{ vertical: 'top', horizontal: 'left' }} />` (the double brackets annoy me for some reason, same when I have to use the `style` prop). If you don't like just calling them `horizontal` and `vertical`, maybe `positionX` & `positionY`.

modules/cactus-web/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const getSpacing = (screenSize: ScreenSize): number => {
+    return screenSize === SIZES.extraLarge ||
+      screenSize === SIZES.large ||
+      screenSize === SIZES.medium
+      ? 40
+      : screenSize === SIZES.small
+      ? 24
+      : 16
+  }
```
wilysword, line 64: 😕 ❓ These aren't exactly CPU-intensive calculations, why are you making functions for them instead of calculating directly? For the second one especially, it might be better to  just pass the screen size as a prop and do the calculation in `getNotificationPosition`: then you could reference `theme.space` instead of hard-coding pixel numbers. (I admit I hard-code pixels way too much, I really need to get better at that...)

wilysword (changes requested):
The third one is the "requested change", the other two are just suggestions.

modules/cactus-web/src/Notification/Notification.tsx
Dhalton: This was how it was stubbed out in the suggested API on the ticket but I don't have a problem making it two props.

modules/cactus-web/src/Notification/Notification.tsx
Dhalton: It just felt cleaner to keep them in functions instead of in the body of the component. I like your suggestion for using `theme.space`, though. Will update.

wilysword (approved):

PR #551, opened 2021-03-15
Cactus 521 : Header Component

daniloPenaR:
[CACTUS 521](https://repayonline.atlassian.net/browse/CACTUS-521)

Let me know how can I achieve a better typification of the components or subcomponents, if necessary. I typed many parameters as `any`.


modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export const Header: HeaderType = ({ children, bgColor = 'lightContrast', ...rest }) => {
+  const childrens = Children.toArray(children)
+
+  return (
+    <StyledHeader bgColor={bgColor} {...rest}>
+      <HeaderColumn>
+        {childrens.find(
+          (child: any) => child.type.displayName === 'HeaderBreadcrumbRow' && <>{child}</>
```
mikesoltow, line 45: To avoid the `any` here, you could do the following if you want.  First try adding the following two lines above line 41:

```typescript
type ChildType = React.ReactChild | React.ReactFragment | React.ReactPortal
type ChildElement = React.ReactElement<any, React.ComponentClass | React.FunctionComponent>
```

Then, you can change this line to the following:

```suggestion
          (child: ChildType) =>
            (child as ChildElement).type.displayName === 'HeaderBreadcrumbRow' && <>{child}</>
```

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+import React, { Children, ReactNode } from 'react'
+import styled from 'styled-components'
+import { margin, MarginProps } from 'styled-system'
+
+import Text from '../Text/Text'
+
+export type BackgroundColorVariants = 'lightContrast' | 'white'
+interface HeaderProps extends MarginProps, React.HTMLAttributes<HTMLDivElement> {
+  bgColor?: BackgroundColorVariants
+  children?: ReactNode
```
mikesoltow, line 11: You shouldn't need to specify `children` here, as it's already provided by `React.FC`.

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export type HeaderType = React.FC<HeaderProps> & {
+  Item: typeof HeaderItem
+  BreadcrumbRow: typeof HeaderBreadcrumbRow
+  Title: typeof HeaderTitle
+}
+
+export const HeaderItem: React.FC<React.HTMLAttributes<HTMLDivElement>> = ({ children }) => {
```
mikesoltow, line 20: The type signature here is not valid.  Since this element is not a div, it shouldn't accept div props.  If you want it to render a div, you could change it to simply: `export const HeaderItem: React.FC<React.HTMLAttributes<HTMLDivElement>> = (props) => <div {...props} />`.  If not, then I'd change this to:

```suggestion
export const HeaderItem = ({ children }: { children?: ReactNode }): React.ReactElement => <>{children}</>
```

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return <>{children}</>
+}
+
+export const HeaderTitle: React.FC = ({ children }) => {
+  return (
+    <Text as="h1" textStyle="h2" my="0">
```
mikesoltow, line 26: I goofed on the ACs.  Originally I thought this had to render an `h1` tag for accessibility, but it turns out that you can start at any h-level (you just don't want to skip levels).  Therefore, it's probably simpler to change this to the following:

```suggestion
    <Text as="h2" my="0">
```

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+      {children}
+    </Text>
+  )
+}
+
+export const HeaderBreadcrumbRow: React.FC<React.HTMLAttributes<HTMLDivElement>> = ({
```
mikesoltow, line 32: Same comment here as for the item.

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+export const HeaderColumn = styled.div<{ mainColumn?: boolean }>`
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  padding-top: ${(p) => !p.mainColumn && '24px'};
+  ${(p) => `
+    ${p.theme.mediaQueries?.small}{
+      &:first-child {
+        flex-direction: column; 
+      }
```
mikesoltow, line 82: This block is probably redundant.

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+      &:last-child {
+        flex-direction: row;
```
mikesoltow, line 84: Could you add `align-items: center;` here?  The vertical alignment of `Header.Item`s is a bit off.

modules/cactus-web/src/Header/Header.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    expect(container).toMatchSnapshot()
+  })
+  test('Should render the Header with Title, Breadcrumb and Items', () => {
+    const { container } = render(
+      <StyleProvider>
+        <Header>
+          <Header.BreadcrumbRow>
+            <Breadcrumb>
+              <Breadcrumb.Item href="/">Label</Breadcrumb.Item>
+              <Breadcrumb.Item href="/" active>
```
mikesoltow, line 43: Could you change this to the following?  The last item will typically not be a link.

```suggestion
              <Breadcrumb.Active>
```

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+Header.propTypes = {
+  bgColor: PropTypes.oneOf(['white', 'lightContrast']),
+}
+
+export const HeaderColumn = styled.div<{ mainColumn?: boolean }>`
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  padding-top: ${(p) => !p.mainColumn && '24px'};
```
mikesoltow, line 77: I don't see the `mainColumn` prop being passed anywhere.  Also, the Header.Title is not vertically centered on mobile screens because of this style setting.  Instead you could change this to `:last-child` (and then override it for the non-mobile view).

mikesoltow (changes requested):
Looks pretty good for the most part, but there are a few tweaks we might want to make.  Most of my comments are up to you, but there are a couple of things we should definitely fix (mobile padding-top, vertical alignment of Header.Items).

modules/cactus-web/src/Header/Header.tsx
daniloPenaR: Thanks, @mikesoltow ! I wonder if you know about some useful links (tutorials, blogs, etc...) where I could take a deep dive into Typescript and React Types (In addition to the documentation)

modules/cactus-web/src/Header/Header.story.tsx
```diff
@@ -0,0 +1,108 @@
+import * as icons from '@repay/cactus-icons'
```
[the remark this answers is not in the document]
daniloPenaR: This is very practical and saves us lines of code, thanks!

modules/cactus-web/src/Header/Header.tsx
daniloPenaR: you're right, at the beginning I was using a `mainColumn` prop but I decided to remove it because I started working with `:last-child` and `:first-child`. Probably I messed this change by undoing a temporal commit (I use temporal commits during the development process to keep a track of my work before the definitive commit and pushing)  
Thanks!

modules/cactus-web/src/Header/Header.tsx
daniloPenaR: @mikesoltow That's a lot better than what I've done here. There is a problem here though when I implement your suggestion: neither the BreadcrumbRow and the Items are being rendered. 

That's weird, I will take a look to find why is this happening.

modules/cactus-web/src/Header/Header.tsx
daniloPenaR: @mikesoltow I just noticed an issue there with our approach, lol. When a user creates a  `Header` 
 without any `Header.Items` the `:last-child` styles will be applied to the first main column (I'm conditionally rendering the second column if there are `Header.Item`s )

I think that maybe my first approach, the `mainColumn` prop, turns out again as a possible solution to this, let me know if you have any comment about this.

modules/cactus-web/src/Header/Header.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const { getByTestId } = render(
+      <StyleProvider>
+        <Header bgColor="white" data-testid="headerComponent"></Header>
+      </StyleProvider>
+    )
+
+    const rawHeader = getByTestId('headerComponent')
+    const headerStyles = window.getComputedStyle(rawHeader)
+    expect(headerStyles.backgroundColor).toBe('rgb(255, 255, 255)')
+  })
```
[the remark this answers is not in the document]
daniloPenaR: @mikesoltow I could render a complete `Header` component with all of its sub-components and take a single snapshot inside this test too. I'm not completely sure about the main differences between the snapshot tests vs screenshot tests but I think a snapshot is very useful in the development process (I execute the unit tests while I'm creating or modifying a component for paying attention to snapshot changes).

It seems I could be a little confused about snapshot and screenshot tests, let me know if so, thanks!

modules/cactus-web/src/Header/Header.tsx
mikesoltow: I don't unfortunately.  I was playing around with this just using VSCode's IntelliSense.  I deleted the `any`s and looked at what it was complaining about.  It indicated that `child` was of type `React.ReactNode`.  I could then CMD+Click on the React types in VS code to be taken to the type definitions.  And I could see that the problem was that `React.ReactNodes` are not guaranteed to have a `displayName` on them (only some descendants of `ReactNode` have that property).  We need to make a [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) to tell TypeScript that we know the children will be an element that has a `displayName`.  Some more clicking in the type definition files led me to to the type we needed to assert on.

modules/cactus-web/src/Header/Header.test.tsx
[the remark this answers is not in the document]
mikesoltow: Good point about the development process.  If you find snapshot tests helpful during development, then we can certainly leave them in.  I typically don't develop alongside with snapshot tests, as I don't find the snapshots very readable.  I just keep checking Storybook for visual changes as I develop.  I _do_ practice TDD for behavior changes however (i.e. does the Select dropdown close when you hit ESC).

That being said, that's just my method.  If it helps you to develop alongside snapshot tests, then please leave them in!

modules/cactus-web/src/Header/Header.tsx
mikesoltow: Ah, good point!  I think the `mainColumn` approach is fine.  However, you should remove the last bit of `:last-child` styles as well.  The `flex-direction: row` and `align-items: center` styles should only be applied to the non-main column (in the larger screens only).

mikesoltow: Actually, one other comment on `align-items`: it looks like the breadcrumbs should be centered over the header title on the mobile view.  Therefore, you'll want `align-items: center` to apply to everything on mobile view, and `align-items: center` to apply to just the non-main column on the larger screens.

modules/cactus-web/src/Header/Header.test.tsx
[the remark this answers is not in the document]
mikesoltow: If you decide to not add these tests back in, then you'll need to delete the obsolete snapshots by running `yarn web test -u` to fix the build.

modules/cactus-web/src/Header/Header.tsx
mikesoltow: I think it might be a weird TypeScript thing.  Try changing it to:

```
export const HeaderItem: FC = ({ children }) => <>{children}</>
```

Dhalton (approved):
LGTM

mikesoltow (approved):
LGTM!  Thanks for the changes

PR #554, opened 2021-03-17
Cactus 550: accordion alignment

daniloPenaR:
[CACTUS-550](https://repayonline.atlassian.net/browse/CACTUS-550)

Following [this workaround](https://github.com/philipwalton/flexbugs/issues/231), the problem seems to be (obviously IE11) that a `min-height` property breaks other properties like `align-items`. 

Test: 

- Open Accordion, Basic usage story on IE11 
- Make sure that the elements inside the menu are centered.
- Test it on mobile and ipad views.


Dhalton (approved):
LGTM

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -240,6 +240,7 @@ export const AccordionHeader = styled(AccordionHeaderBase)`
   background: none;
   border: 2px transparent;
   outline: none;
+  height: 24px;
```
mikesoltow, line 243: This solution doesn't work if you have super-long content in the Accordion.  To test this, try using the Storybook knobs to create super-long text (see screenshot), and note that it overflows the top and bottom boundaries.  This doesn't happen in the deployed version.

Instead, I'd try a different workaround from the same thread you linked in the PR.  https://github.com/philipwalton/flexbugs/issues/231#issuecomment-362790042

This seems to work better in all cases.
Screen Shot 2021-03-18 at 12 40 37 PM

mikesoltow (changes requested):

modules/cactus-web/src/Accordion/Accordion.tsx
daniloPenaR: Doh, I didn't test it with a long string, going to look for another solution.

Dhalton:
Updates look good to me. Nice catch, Mike.

mikesoltow (changes requested):
If you look at the updated screenshot, you can see that the IconButtons no longer line up toward the end on the With Outline story.  This is probably because of the `width: 90%;` bit that was added.  What did that fix again?

mikesoltow (approved):
LGTM!

PR #558, opened 2021-03-19
CACTUS-549 :: Mobile Breadcrumbs

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-549

This was definitely a challenge, mostly because devs can use custom components for their links. You'll probably notice some workarounds that I implemented to get this to work, but if you have any ideas for how I could do this better, don't hesitate to suggest them; I know there are probably ways to implement this better that I didn't think of.

### Testing
1. Open the "Custom Item Elements" Breadcrumb story and set the viewport to a tiny screen size.
2. Make sure you can select a link using your mouse, both by clicking on the first breadcrumb and the ellipsis in the middle.
3. Make sure you can select a link using the keyboard, triggering the dropdown from both the first breadcrumb and the ellipsis.
4. Repeat steps 1-3 using the "Add more Breadcrumbs" story.
5. Set your viewport back to normal and make sure the breadcrumbs still work as they did before on larger screen sizes.
6. Do a quick check on the shape and box shadow of the dropdowns in the ColorPicker, DateInput, and MenuButton to make sure the refactoring didn't break anything there.


modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -36,13 +64,185 @@ export const BreadcrumbActive = (
...
+      const popup = document.querySelector('#breadcrumb-nav-popup') as HTMLUListElement
+      toggle(true, popup)
+    },
+    [toggle]
+  )
+
+  React.useEffect(() => {
+    // We can't pass onClick or onKeyDown handlers to BreadcrumbItems directly because they could be using
+    // custom components. To circumvent this, we can listen for click & keydown events on the document
+    // and manually trigger the event handlers when necessary.
```
mikesoltow, line 97: Could you tell me a bit more about this?  I was able to remove these `document` event handlers and use this with React Router (admittedly I haven't tested the old version that Channels uses yet...just the new one we use in our app generator), and everything seemed to work well.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -36,13 +64,185 @@ export const BreadcrumbActive = (
...
+                {Children.map(children, (child, index) => {
+                  const cloneProps: {
+                    mobileListItem: boolean
+                    handleItemMouseEnter: () => void
+                    itemIndex: number
+                    isSelected?: boolean
+                  } = {
+                    mobileListItem: true,
+                    itemIndex: index,
+                    handleItemMouseEnter: () => setSelectedIndex(index),
```
mikesoltow, line 206: If we use a popup type of `menu`, then we'll want to pass `role: 'menuitem'` into these cloned items.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -36,13 +64,185 @@ export const BreadcrumbActive = (
...
+  const childrenCount = Children.count(children)
+  const childrenArray = Children.toArray(children)
+  const breadcrumbNavId = 'breadcrumb-nav'
+
+  const isTiny = SIZES.tiny === useScreenSize()
+  const firstBreadcrumb = useRef<HTMLDivElement | null>(null)
+  const ellipsisButton = useRef<HTMLButtonElement | null>(null)
+  const popup = useRef<HTMLDivElement | null>(null)
+  const mainBreadcrumbList = useRef<HTMLUListElement | null>(null)
+  const { expanded, toggle, buttonProps, popupProps } = usePopup('dialog', {
```
mikesoltow, line 76: If we end up getting rid of the `document` event handlers, then you'll want to extract the `wrapperProps` from the returned value and pass it into the `Nav`.  This will allow `usePopup`'s close-on-blur functionality to work.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -36,13 +64,185 @@ export const BreadcrumbActive = (
...
+    id: breadcrumbNavId,
```
mikesoltow, line 77: I think we could simplify this by passing in a position function here instead of using a separate `usePositioning` hook.  See my suggested commit for a full example.  It seems to work 🤷

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -36,13 +64,185 @@ export const BreadcrumbActive = (
...
+  const [selectedIndex, setSelectedIndex] = React.useState<number>(0)
+
+  const maxDropdownWidth = mainBreadcrumbList.current?.getBoundingClientRect().width
+
+  const handleTriggerClick = React.useCallback(
+    (event: React.MouseEvent) => {
+      event.preventDefault()
+      setSelectedIndex(0)
+      const popup = document.querySelector('#breadcrumb-nav-popup') as HTMLUListElement
+      toggle(true, popup)
```
mikesoltow, line 89: Should we change `true` to `undefined`?  That will allow clicks to toggle the menu closed as well.

mikesoltow (changes requested):
Nice work...you're right that this was a tricky one!  I have a few suggestions, and i've left both comments in the code and below.  Most of them are suggestions/questions except for the handling of `Breadcrumb.Active`, which we'll definitely want to get in there.

I didn't have time to get to a couple of things that I noticed (z-index and dropdown radius), so forgive me for not digging into those more to provide better comments.

Finally, because I still have little hands-on experience with `usePopup` n' such, I decided to try out my suggestions myself and make sure they weren't total 💩 . You can find the diff at https://github.com/repaygithub/cactus/compare/cactus-549-mobile-breadcrumbs...mike-breadcrumb-suggestions?expand=1

* I goofed and forgot to wrap `BreadcrumbActive in an `li` in my Breadcrumb link PR.  Would you be able to do that for me?  (I do that in my suggested commit.)
* We'll want to add `z-index: 1000;` to the popup to match styles from other menu dropdowns, right?
* The border radius of the dropdown looks rounder than the menu button...but I didn't have time to dig into this.


https://github.com/repaygithub/cactus/compare/cactus-549-mobile-breadcrumbs...mike-breadcrumb-suggestions?expand=1

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
Dhalton: It comes back to whether or not the custom component they're using passes along any props it receives on to the elements they render. For components that do pass on their props, it works because the `handleTriggerClick` and `keyDownAsClick` handlers are passed through and allows the dropdown to be opened/closed that way. However, if the component doesn't pass along the props it receives (like the one in the custom item elements story), clicking on the first link just redirects you instead of opening the dropdown.

I don't think we want to require that any custom components must pass on their props, so what I'm thinking is we keep the document level event handlers and refrain from passing the `onClick` & `onKeyDown` handlers to the first breadcrumb (otherwise if the handlers _are_ passed through, the handler would get triggered twice and wouldn't open/close the dropdown). What do you think about that?

mikesoltow: Yeah, I think that makes sense.  I'd be fine with that.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -36,13 +64,185 @@ export const BreadcrumbActive = (
...
+        setSelectedIndex(0)
+        break
+      case 'End':
+        setSelectedIndex(childrenCount - 1)
+        break
+      case 'Enter':
+      case ' ':
+        const link = document.querySelector("[data-selected='true'] a") as HTMLAnchorElement | null
+        link?.click()
+        break
```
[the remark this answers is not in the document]
Dhalton: Yeah this is definitely one spot that I was hoping to get some feedback on. I don't like that this assumes there's always an anchor tag, but it's not quite that simple. Your suggestion doesn't work because the links in the dropdown never actually have focus. Only the dropdown div ever has focus, so hitting enter or space when the div is focused doesn't do anything.

I don't think I'm able to focus the links themselves inside the dropdown, because then I have no way of giving them an `onKeyDown` handler (b/c of custom components) and that means this code wouldn't run.

I've thought about trying to focus the `li`s in the dropdown instead, but you run into the same issue here, where hitting enter or space when the `li` is focused also doesn't do anything so I'd still have to find the link and trigger the click programmatically.

Open to any other ideas you might have, though, because I would like to find a less specific solution if possible.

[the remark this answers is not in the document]
Dhalton: For what it's worth, I did make sure that React Router (by default, at least) and Reach Router both use anchor elements internally for their `Link`s. Realistically, I don't know what other elements you'd want in there; I mean, they are supposed to take you to a different page, so an `a` element _should_ be used. But like I said, I do realize the potential problem this could cause and I would like to avoid doing this if it is possible.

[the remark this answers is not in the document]
mikesoltow: I'd need to dig into this more to see if this is feasible, but is it possible to mimic what the menu bar is doing?  It looks like it calls `setFocus(1, { shift: true })` on any scroll forward key, where `setFocus` comes from `usePopup` (and can be defined with a custom focus control, I believe).  I don't know what the implications of custom elements are, but presumably they are no different from the constraints the MenuBar would have to deal with, since the menu items often take the form `<MenuBar.Item as={Link} to="/a-route">`.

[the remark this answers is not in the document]
Dhalton: I'll take a peek at that code and see if I can figure out how it works there and apply it to this component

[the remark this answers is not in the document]
Dhalton: Works great! I'll go with that, thanks for the suggestion!

modules/cactus-web/src/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -33,6 +33,149 @@ describe('Breadcrumb should have label and href', (): void => {
...
+              <Breadcrumb>
+                <Breadcrumb.Item href="www.github.com">Link 1</Breadcrumb.Item>
+                <Breadcrumb.Item href="www.github.com">Link 2</Breadcrumb.Item>
+                <Breadcrumb.Item href="www.github.com">Link 3</Breadcrumb.Item>
+              </Breadcrumb>
+            </ScreenSizeContext.Provider>
+          </StyleProvider>
+        )
+
+        const firstBreadcrumb = getAllByText('Link 1')[0]
```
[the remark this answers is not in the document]
mikesoltow: Shoot!  That is lame 😢

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -127,88 +112,78 @@ const BreadcrumbBase = (props: BreadcrumbProps): React.ReactElement => {
...
+                    (child as ChildElement)?.type.displayName !== 'Breadcrumb.Active' &&
+                    (child as ChildElement)?.props.active !== true
+                  ) {
+                    const cloneProps: {
+                      mobileListItem: boolean
+                      handleItemMouseEnter: () => void
+                      role: string
+                    } = {
+                      mobileListItem: true,
+                      handleItemMouseEnter: () => setFocus(index),
```
mikesoltow, line 182: I'm not sure that, semantically, we want the links to focus on hover.  See my next comment for an alternate suggestion.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -287,7 +258,7 @@ const BreadcrumbPopupList = styled.ul<{ $maxWidth?: number }>`
     padding-top: ${(p) => p.theme.space[2]}px;
     padding-bottom: ${(p) => p.theme.space[2]}px;
 
-    &[data-selected='true'] {
+    &:focus-within {
```
mikesoltow, line 261: This style doesn't work on IE, unfortunately, so we get an outline on hover (see screenshot).  While I normally wouldn't want to bend over backwards to get IE mobile styles correct, I think this combined with the "focus on hover" probably warrants reworking it.  I think we actually might want to bring back the `data-selected` attribute.  That attribute could be set to true on hover or on focus, which you had before, and which mimics the behavior of the MenuButton.
Screen Shot 2021-03-24 at 4 27 06 PM

mikesoltow (commented):
This is looking great!  Everything is working well functionally.  Just a few more comments.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -127,88 +112,78 @@ const BreadcrumbBase = (props: BreadcrumbProps): React.ReactElement => {
...
             </li>
             {childrenArray[childrenCount - 1]}
             <BreadcrumbPopup
-              anchorRef={firstBreadcrumb}
               popupRef={popup}
               expanded={expanded}
               {...popupProps}
               onKeyDown={handlePopupKeyDown}
-              aria-owns="nav-popup-list"
-              aria-activedescendant={`mobile-list-item-${selectedIndex}`}
```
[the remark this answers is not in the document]
Dhalton: According to the W3C spec, it's used when a container element is focusable but its children are not focusable, so it replaces the need for a focusable child. Now that the links inside the list are actually receiving focus, we shouldn't need it. The user agent can tell which descendant is active by the focus.

[Relevant W3C aria-activedescendant Docs](https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology#:~:text=The%20aria%2Dactivedescendant%20attribute%20is,elements%20are%20not%20natively%20focusable.)

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -127,88 +112,78 @@ const BreadcrumbBase = (props: BreadcrumbProps): React.ReactElement => {
...
-                    handleItemMouseEnter: () => void
-                    itemIndex: number
-                    isSelected?: boolean
-                  } = {
-                    mobileListItem: true,
-                    itemIndex: index,
-                    handleItemMouseEnter: () => setSelectedIndex(index),
+                  if (
+                    (child as ChildElement)?.type.displayName !== 'Breadcrumb.Active' &&
+                    (child as ChildElement)?.props.active !== true
```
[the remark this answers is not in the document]
Dhalton: Yeah they're links but they'd still just take you back to the same page you're on, right (assuming `active` is set correctly)? I don't see any value in including them in the dropdown if that's the case, especially since, like you mentioned, it should be visible outside the dropdown. I don't feel super strongly about it either, though.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
Dhalton: Yeeahhhh you're right this is probably bad practice I'll update it

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
Dhalton: Darn you, IE! Yeah I'll bring back the `data-selected` attribute and do the styling like I was before...I ran into some issues with that before where my active index and the focused element were somehow out of sync but I can probably figure it out.

wilysword (approved):

mikesoltow (approved):
Changes look good.  Thanks, Dhalton!

PR #560, opened 2021-03-23
Cactus 560 menubar click

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-560

As noted in Slack, you'll note that the bottom box shadow is no longer present.  However, I think losing it is not the worst thing in the world, and I didn't think it would be wise for me to spend any more time on this.

Note that I changed the Layout stories to use the new Header component to more accurately reflect our apps' structure.  While doing that, I found the `border-box` bug Glen brought up in Slack, so I went ahead and fixed that as well.  As noted in the ticket's UAT, there's another bug in Safari for the MenuBar.  I'll be creating a follow-up ticket shortly for that.


daniloPenaR (approved):
Wow, how tricky can a simple CSS detail turn in!

And I'm sorry I mess that box-sizing property when I was creating the header component >.<

wilysword (approved):

PR #563, opened 2021-04-01
CACTUS-571 :: TS Bugs

mikesoltow:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-571

In addition to fixing the issue in `index.js`, I changed all the imports in story files to use `index.ts` for our publicly exported components.  That will ensure that our stories use the same exact import that our users will, which will allow us to catch these errors.

Make sure `yarn web test:types` works and that all the stories still function.


daniloPenaR (commented):
Looks good to me. It seems that we got some problems with Github's CI (integration tests, surprise surprise), I don't know if any specific change is needed to solve that inconvenient, tell me whether if you need to do some additional commits or is ready for my approval :)

Thus, I suppose that for the next components/features we have to start using this import format, isn't it?

Dhalton:
LGTM, I went ahead and pulled down your branch to see what was going on with the types. It looks like there was a `retry` label you could set in the `labels` prop passed to FileInput that must've been removed at some point, so we just need to remove it from the FileInput container in `theme-components`. So in a way, the failing CI confirms that the types are being correctly exported. I'll wait to approve until that's resolved, though.

mikesoltow:
> Thus, I suppose that for the next components/features we have to start using this import format, isn't it?

@daniloPenaR yup!  And your question reminded me that I should change the `make-component` script to use that format, which I just did.  Thanks!

Dhalton (approved):
LGTM!

daniloPenaR (approved):
Nice!

PR #564, opened 2021-04-05
CACTUS-556 :: MenuBar Sub-Menu Width

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-556

So if you look [here](https://caniuse.com/mdn-css_properties_width_max-content) you'll see that caniuse says that `width: max-content;` is not supported in IE11, but it's working for me now. It wasn't working at first, when I was only applying the style to sub-menus, but when I started setting it on all menus it began to work. I'd say be extra thorough when testing IE11 since it doesn't look like it should be supported.

### Testing
1. Open the Layout "Basic Usage" story
2. Open the "Interlude: About Bears" menu and then open the "Long Bears" sub-menu.
3. Verify that the width of the sub-menu is 320px wide.


daniloPenaR (approved):
Looks good to me.

wilysword (approved):
Who knows what goes on in IE's digital mind?

PR #566, opened 2021-04-09
Cactus 569 :: header improvements

daniloPenaR:
[CACTUS-569](https://repayonline.atlassian.net/browse/CACTUS-569)

- [x] _Horizontal padding is added to the mobile version of the header_: 16px of horizontal padding
- [x] _The Header.Item component should have horizontal spacing on desktop mode and vertical spacing on mobile mode _ : 8px of Horizontal and vertical margin added
- [x]  _The Header.Item component should render a div_ : the `Header` item now returns a `<div>{children}</div>` and passing `{...props}`

- [x] _All props passed into the `Header.Title `component should be passed into the rendered_ : Now the props of the `Header.Title` are passed directly to the Text component via `{...props}`


mikesoltow (changes requested):
This looks mostly great.  In fact, all of the comments in the code are just minor suggestions: you can take them or leave them.

The only reason I've requested changes is that long headers seem to overflow the page on Firefox's mobile view.  The only thing I was able to do to fix it was to add a bunch of `max-width: 100%` lines to the `div` and `h2` inside `Header.Title`, but I feel like I'm missing something more fundamental..

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -73,9 +73,18 @@ export const HeaderColumn = styled.div<{ mainColumn?: boolean }>`
   flex-direction: column;
   padding-top: ${(p) => !p.mainColumn && '24px'};
   align-items: center;
+  > :not(:first-child) {
+    margin-top: ${(p) => !p.mainColumn && '8px'};
```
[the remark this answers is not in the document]
daniloPenaR: Here I am using the `:not(:first-child)` to prevent that the first element of the possible array of `HeaderItems` will not take the `margin-top` property. Otherwise, an extra space of 8px will be added between the `mainColumn` and the items column

Screen Shot 2021-04-12 at 12 20 26 PM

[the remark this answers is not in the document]
daniloPenaR: if I put something like this 
```
 > div {
    margin-top: ${(p) => !p.mainColumn && '8px'};
  }
```
An extra space will be added on the top of the first element
Screen Shot 2021-04-12 at 12 26 18 PM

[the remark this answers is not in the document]
mikesoltow: D'oh!  You are right!  My mind was not at the right level in the hierarchy...I wash thinking that the "first child" was the main column.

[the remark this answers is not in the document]
daniloPenaR: So I think I could skip the two comments below if you agree :)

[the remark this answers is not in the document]
mikesoltow: Fine by me!

mikesoltow (commented):
Changes look good in everything except IE.  For some reason, the header text is overflowing when it's too long (see screenshot).
Screen Shot 2021-04-13 at 1 30 11 PM

daniloPenaR:
> Changes look good in everything except IE. For some reason, the header text is overflowing when it's too long (see screenshot).
> Screen Shot 2021-04-13 at 1 30 11 PM

Gotcha, going to solve.

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -73,13 +73,31 @@ export const HeaderColumn = styled.div<{ mainColumn?: boolean }>`
...
+  }
 
   ${(p) => `
     ${p.theme.mediaQueries?.small}{
+      > div:not(:first-child) {
+        margin-top: ${!p.mainColumn && '0px'};
+      }
+      > div{
+        margin-left: ${!p.mainColumn ? '8px' : '0px'};
+      }
```
mikesoltow, line 87: I think we want to flip the "not first child" selectors.  As it stands, the first child is getting a top margin while the others aren't (see screenshot where the first button is lower than the next two items).  Also, all of the buttons are getting a left margin.  Not sure if that's intended or if you wanted the first item to have no left margin.

```suggestion
      > div {
        margin-top: ${!p.mainColumn && '0px'};
      }
      > div:not(:first-child) {
        margin-left: ${!p.mainColumn ? '8px' : '0px'};
      }
```

Screen Shot 2021-04-15 at 12 17 35 PM

mikesoltow (commented):
There's something something weird going on in the mobile view of IE11, but it's _probably_ not worth a fix (unless it takes 5 mins), as I don't think anyone uses IE11 on tiny screens.

Other than that, there's just my one comment.

modules/cactus-web/src/Header/Header.tsx
daniloPenaR: @mikesoltow Well, at the first time I omit the left-margin in the first item just to prevent aggregating too much space between the title/mainColumn and the first item

Dhalton:
I'm seeing some weird-ness in the iPad view...see screenshot below. It looks like we need to provide more space for the title/breadcrumbs. I think it'd make more sense for the header items to wrap onto the next line than for the title/breadcrumbs to do that. Or maybe if there are several header items like that they should just be rendered below the title. I'm having trouble locating the designs for the header, so I'm not sure what the designers recommend at that screen size, but I think either one of those suggestions would at least help clean it up.

Screen Shot 2021-04-15 at 3 53 11 PM

daniloPenaR:
there's the weirdness in the  IE11's Mobile view and I tried it to solve by adding many CSS properties as `flex` or `min-width` but I can't figure out what is happening... Maybe I am missing any petite detail or something

modules/cactus-web/src/Header/Header.story.tsx
```diff
@@ -73,12 +73,9 @@ export const WithMultipleItems = (): React.ReactElement => {
         </Breadcrumb>
       </Header.BreadcrumbRow>
       <Header.Item>
-        <Button variant="action">
-          <ActionsAdd /> Add new configuration
-        </Button>
+        <Text>You are configuring Merchant directory from OWE Demo Merchant</Text>
```
Dhalton, line 76: Why remove all of the other things that were being rendered here? This was the story that started the conversation about rendering the header items in a separate column...I can't check to see how it looks with the changes you made if the extra items aren't here anymore.

daniloPenaR: @Dhalton  Sorry! As we were discussing on Friday in a thread via `ui-team`, we don't have many use cases for three long items in the right column, and on that day we were testing how a long text like that does look in different screen sizes. However, I set the small view to aligning the items in the right part and in a single column. 

I have restored that use case, so you can take a look again 👍

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -34,26 +34,85 @@ export const Header: HeaderType = ({ children, bgColor = 'lightContrast', ...res
...
+        word-wrap: break-word;
+        max-width: 100%;
+        hyphens: auto;
+      }
+    }
+
+    ${p.theme.mediaQueries?.medium}{
+      flex-direction: row;
+      align-items: center;
+      flex: 0 1 auto;
```
Dhalton, line 111:
```suggestion
      flex: 0 1 auto; 
      > div:not(:first-child) {
        margin-top: 0px;
      }
```

Other than that I'm ready to approve this

Dhalton (approved):
I think you've addressed all of my concerns. As far as the weird overflow thing in IE11 on mobile goes, obviously it'd be nice to fix it, but at the end of the day I don't even think it's _possible_ to use IE on a mobile device, so I'm not too worried about it.

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -34,73 +34,101 @@ export const Header: HeaderType = ({ children, bgColor = 'lightContrast', ...res
...
-      > div{
-        margin-left: ${!p.mainColumn ? '8px' : '0px'};
+        margin-left: 0px;
       }
-      flex-direction: ${!p.mainColumn ? 'row' : ''};
-      align-items: ${!p.mainColumn ? 'center' : 'flex-start'};
+      align-items: flex-start;
+    }
+    flex: 1 1;
+    > h2 {
```
mikesoltow, line 76: This selector won't work if the use does something like `<Header.Title as="h3">`.  Also, if the Breadcrumbs have many items, it's possible for them to overflow on IE11 as well.  I think we can just change this selector to `> *` unless you can think of a reason why that wouldn't be good.

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -16,11 +16,11 @@ export type HeaderType = FC<HeaderProps> & {
   Title: typeof HeaderTitle
 }
 
-export const HeaderItem: FC = ({ children }) => <>{children}</>
+export const HeaderItem: FC = (props) => <div {...props} />
 
-export const HeaderTitle: FC = ({ children }) => {
+export const HeaderTitle: FC = ({ children, ...props }) => {
```
mikesoltow, line 21: Sorry for not realizing this sooner, but these `props` have a type of `{}` because you haven't defined a props interface for this component.  I think you can change `FC` to `FC<React.ComponentProps<typeof Text>>`, as that should give the same props interface that `Text` has (which is what we want, since we're rendering a `Text` component).

modules/cactus-web/src/Header/Header.tsx
```diff
@@ -16,11 +16,11 @@ export type HeaderType = FC<HeaderProps> & {
   Title: typeof HeaderTitle
 }
 
-export const HeaderItem: FC = ({ children }) => <>{children}</>
+export const HeaderItem: FC = (props) => <div {...props} />
```
mikesoltow, line 19: Similar to my previous comment, we'll want to add a type for `props` here.  For this one, you could use:

```typescript
FC<
  React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>
>
```

(I got this by hovering over a `div` element in VSCode and copying the type signature for the props.)

mikesoltow (commented):
Sorry...I found a couple more things!  The overall changes do look good though.

modules/cactus-web/src/Header/Header.tsx
daniloPenaR: Nice catch, I was interpreting this `...props` in the wrong way. Definitively need to practice and learn too much about Typescript's type declarations.

daniloPenaR:
@mikesoltow I'm currently running the `test:docker-update` so I hope the last changes will be here in the next hour. I cannot get logged into any browserstack account cause I'm having problems with the LastPass login too (they said they'll send me an email but nothing has come)...

mikesoltow (approved):
Looks good...thanks, Danilo!

PR #570, opened 2021-04-14
CACTUS-575 :: Fix FileInput onChange not Firing

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-575

### Testing
1. Pull down this branch and run `yarn web cleanup && yarn web build`
2. In Channels, copy & paste this line into the `docker-compose.yml` file in the admin container under the `volumes` section: `- ../cactus/modules/cactus-web/dist:/code/node_modules/@repay/cactus-web/dist`
3. Run `make init` in the Channels repo
4. When your channels instance is up and running, open FireFox and log in to the app there. Make your way to `https://owe.channels.dev.repay.localhost/admin/#/public-assets/create`
5. Upload a file and make sure the image appears on the page and that the submit button is enabled.
6. Complete step 5 in other browsers of your choosing.

### DIFF
--- a/modules/cactus-web/src/FileInput/FileInput.tsx
+++ b/modules/cactus-web/src/FileInput/FileInput.tsx
@@ -529,10 +529,10 @@ const FileInputBase = (props: FileInputProps): React.ReactElement => {
   }, [box, eventTarget, files])
 
   useEffect((): void => {
-    if (value) {
+    if (value && !files.some((f) => f.status === 'loading')) {
       updateFiles({ control: value })
     }
-  }, [value, updateFiles])
+  }, [value, updateFiles, files])
 
   const handleDrop = React.useCallback<React.DragEventHandler>(
     (event) => {


wilysword:
I think the effect (or something like it) is necessary, because that's how the input is "controlled". Without it, there might as well not even _be_ a value prop. (Although you could argue it doesn't make sense for a file input anyway, given that you can't really get File objects from anyplace else.) At the very least we need some way to clear the value in the event of a form reset, and I used that effect to do it in the ticket I'm working on right now. It doesn't have to work exactly like it does now, though.

Is redux-form passing the empty array _before_ the `onChange` event even fires?

Dhalton:
> Is redux-form passing the empty array _before_ the `onChange` event even fires?

No, it's not passed before the hidden file input `onChange` is triggered, if that's what you mean. It passes an empty array after you've chosen a file (so after the hidden file input `onChange` event fires) but before it finishes loading.

It seems like this PR definitely won't be merged in its current state, but I'll leave it open for now to make sure everyone else has a chance to see what's going on. If anyone's got any ideas for a better solution to get this fixed ASAP, let me know

Dhalton:
> It seems like this PR definitely won't be merged in its current state, but I'll leave it open for now to make sure everyone else has a chance to see what's going on. If anyone's got any ideas for a better solution to get this fixed ASAP, let me know

Scratch that, explaining what was happening to Glen gave me that "Aha!" moment. 🤦  Can't believe I didn't think to try this before, but I think what I have now should work without sacrificing the controlled-ness.

wilysword (approved):
While not ideal, it solves the issue and I don't think it will cause any other problems before we can revamp how the input works.

daniloPenaR (approved):
I tested it on Firefox and chrome and it is working. Thanks.

PR #571, opened 2021-04-15
CACTUS-567 Clear custom inputs using empty string

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-567

I was considering taking out the changes to FileInput in light of https://github.com/repaygithub/cactus/pull/570 and the subsequent discussion of changes we could make to the FileInput; I figured we could just make sure to include the ability to clear in the ACs for the new ticket. Since Dhalton changed how he was applying the fix, though, I decided I might as well keep this.

Select obviously has changes way beyond the apparent scope of the ticket; but when I was trying to find the correct place to modify, I discovered that the way value & options & extraOptions interacted didn't make any sense (some stuff was happening before render, some during render, and some after). So I moved it all to one place so it flows in an easily traceable, linear fashion.

- Replaced redundant type definitions (mostly various instances of `string | number`).
- Moved several functions (`buildOptsFromChildren`, `asExtOpt` renamed `addOption`, `getExtOptions` renamed `getNextState`) to the new OptionState class.
  * `buildOptsFromChildren` is almost identical, though I moved some of the `altText` handling to `addOption`.
  * `addOption` and `getNextState` were simplified, with some of the redundant elements they shared moved to one or the other.
  * I thought about moving the `extraOptions` state into this class as well, but that would have made it weirder to update the component's state when adding a new option.
 - Moved the code from `detectOptionsFromValue` into `getDerivedStateFromProps` (this is the one that really instigated the refactor: you change `state.value` _before_ render but then for some reason wait until _after_ render to update `state.extraOptions`? Not on my watch 😉).


Dhalton:
I haven't been able to track down why this is happening, but I think there may be something wrong with the FileInput. If I add a file, click the button to clear the files, then try to add another file, it won't add it. It looks like the `handleFileSelect` handler isn't being triggered in that case, for some reason.

Also, would you mind adding a reset button to one of (or a couple of) the Select stories? Or if you have a way for me to test that component without a reset button, that's fine too.

wilysword:
Hmm, that doesn't happen in my browser. I'm seriously considering just taking FileInput out of this PR and deferring that bit of functionality until https://repayonline.atlassian.net/browse/CACTUS-585 is ready, since I don't think there's an urgent need for it.

For Select I had thought the unit tests were sufficient, although if you really want a way to manually test it I can add that.

Dhalton:
> I'm seriously considering just taking FileInput out of this PR and deferring that bit of functionality until https://repayonline.atlassian.net/browse/CACTUS-585 is ready, since I don't think there's an urgent need for it.

Yeah, it may not make sense to try to fix that for the FileInput when it's going to be heavily refactored anyway.

> For Select I had thought the unit tests were sufficient, although if you really want a way to manually test it I can add that.

Fair enough, no need to add a reset button then. I trust your tests.

I'm more or less good with this, but I'll hold off on approving until you decide if you want to leave the FileInput changes or not.

Dhalton (approved):

mikesoltow (approved):
LGTM

PR #573, opened 2021-04-22
CACTUS-584 Remove redundant Select state

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-584

Every time I look at this component I just want to refactor it to death. (Or rather to life, I suppose.) Still, this is one little bit cleaner.


Dhalton (approved):
LGTM, works well.

daniloPenaR (approved):
I tested that and it works. 

A complex component, I like what you're doing there.

PR #574, opened 2021-04-23
feat(cactus-web) :: Footer link props

daniloPenaR:
Passed the same props of the Link component allowing `as`

[CACTUS-580](https://repayonline.atlassian.net/browse/CACTUS-580)

Please, let me know if there is something I can do better. 
Thanks!

### DIFF
--- a/modules/cactus-web/src/Footer/Footer.tsx
+++ b/modules/cactus-web/src/Footer/Footer.tsx
@@ -165,7 +165,7 @@ export const Footer: FooterType = (props) => {
           {dividedLinks.map((linkGroup, colIndex) => (
             <LinkCol key={colIndex} maxCols={dividedLinks.length}>
               {linkGroup.map((link, i) => (
-                <StyledLink key={i} {...(link as any)}/>
+                <StyledLink key={i} {...(link as any)} />
               ))}
             </LinkCol>
           ))}
@@ -176,7 +176,7 @@ export const Footer: FooterType = (props) => {
 }
 
 export const FLink: any = (props: any) => {
-  const {id, children, to} = props
+  const { id, children, to } = props
   const setLinks = useContext(FooterContext)
   const key = useId(id)
   useEffect(() => {
--- a/modules/cactus-web/src/Footer/Footer.tsx
+++ b/modules/cactus-web/src/Footer/Footer.tsx
@@ -8,13 +8,7 @@ import { useLayout } from '../Layout/Layout'
 import Link from '../Link/Link'
 import { ScreenSizeContext, Size, SIZES } from '../ScreenSizeProvider/ScreenSizeProvider'
 
-interface LinkProps {
-  children: React.ReactNode
-  to: string
-  id?: string
-}
-
-type LinkMap = Map<string, LinkProps>
+type LinkMap = Map<string, typeof Link>
 type FooterContextType = (u: (s: LinkMap) => LinkMap) => void
 
 interface LinkColProps {
@@ -120,8 +114,8 @@ const columnsMap: { [K in Size]: number } = {
 }
 
 // Divides array of links into smaller arrays of links for splitting them up in the list view
-const divideLinks = (links: LinkProps[], maxCols: number) => {
-  const divided: LinkProps[][] = []
+const divideLinks = (links: typeof Link[], maxCols: number) => {
+  const divided: typeof Link[][] = []
   const numPerCol = Math.floor(links.length / maxCols)
   const extra = links.length % maxCols
   for (let i = 0, linkIndex = 0; i < maxCols && linkIndex < links.length; i++) {
@@ -135,11 +129,11 @@ const divideLinks = (links: LinkProps[], maxCols: number) => {
   return divided
 }
 
-type FooterType = React.FC<FooterProps> & { Link: React.FC<LinkProps> }
+type FooterType = React.FC<FooterProps> & { Link: typeof Link }
 
 export const Footer: FooterType = (props) => {
   const { logo, className, children } = props
-  const [links, setLinks] = useState(new Map<string, LinkProps>())
+  const [links, setLinks] = useState(new Map<string, typeof Link>())
   const [maxCols, setMaxCols] = useState<number>(1)
   const screenSize = useContext(ScreenSizeContext)
 
@@ -171,7 +165,7 @@ export const Footer: FooterType = (props) => {
           {dividedLinks.map((linkGroup, colIndex) => (
             <LinkCol key={colIndex} maxCols={dividedLinks.length}>
               {linkGroup.map((link, i) => (
-                <StyledLink key={i} {...link} />
+                <StyledLink key={i} {...(link as any)}/>
               ))}
             </LinkCol>
           ))}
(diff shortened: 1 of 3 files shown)


modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -8,13 +8,7 @@ import { useLayout } from '../Layout/Layout'
...
 import { ScreenSizeContext, Size, SIZES } from '../ScreenSizeProvider/ScreenSizeProvider'
 
-interface LinkProps {
-  children: React.ReactNode
-  to: string
-  id?: string
-}
-
-type LinkMap = Map<string, LinkProps>
+type LinkMap = Map<string, typeof Link>
```
Dhalton, line 11: The types are a bit off throughout this component...the type of the links stored in this map isn't an instance of the `Link` component, like you have written here; it should be an array of the Link _props_. So we should be exporting `LinkProps` from the Link component file and using that in place of the `LinkProps` that were defined in this file.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -120,8 +114,8 @@ const columnsMap: { [K in Size]: number } = {
 }
 
 // Divides array of links into smaller arrays of links for splitting them up in the list view
-const divideLinks = (links: LinkProps[], maxCols: number) => {
-  const divided: LinkProps[][] = []
+const divideLinks = (links: typeof Link[], maxCols: number) => {
```
Dhalton, line 117: The type here should still be `LinkProps`, but it should be the `LinkProps` from the Link component file.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -120,8 +114,8 @@ const columnsMap: { [K in Size]: number } = {
...
+  const divided: typeof Link[][] = []
```
Dhalton, line 118: Same thing here

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -135,11 +129,11 @@ const divideLinks = (links: LinkProps[], maxCols: number) => {
   return divided
 }
 
-type FooterType = React.FC<FooterProps> & { Link: React.FC<LinkProps> }
+type FooterType = React.FC<FooterProps> & { Link: typeof Link }
```
Dhalton, line 132: This should still be `Link: React.FC<LinkProps>`, but again, the `LinkProps` should come from the Link component.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -135,11 +129,11 @@ const divideLinks = (links: LinkProps[], maxCols: number) => {
...
 
 export const Footer: FooterType = (props) => {
   const { logo, className, children } = props
-  const [links, setLinks] = useState(new Map<string, LinkProps>())
+  const [links, setLinks] = useState(new Map<string, typeof Link>())
```
Dhalton, line 136: Should still be `LinkProps`

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -171,7 +165,7 @@ export const Footer: FooterType = (props) => {
           {dividedLinks.map((linkGroup, colIndex) => (
             <LinkCol key={colIndex} maxCols={dividedLinks.length}>
               {linkGroup.map((link, i) => (
-                <StyledLink key={i} {...link} />
+                <StyledLink key={i} {...(link as any)} />
```
Dhalton, line 168: You're having to add this `as any` here because the component thinks you're trying to spread an instance of a component into the props, which wouldn't make sense. If you make the other changes I'm mentioning, you shouldn't need to do this.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -181,11 +175,12 @@ export const Footer: FooterType = (props) => {
   )
 }
 
-export const FooterLink: React.FC<LinkProps> = ({ children, id, to }) => {
+export const FLink: any = (props: any) => {
```
Dhalton, line 178: This shouldn't accept `any`. It should be accepting the `LinkProps`.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -199,6 +194,8 @@ export const FooterLink: React.FC<LinkProps> = ({ children, id, to }) => {
   return null
 }
 
+const FooterLink = FLink as typeof Link
```
Dhalton, line 197: I think this should just be `type FooterLink = typeof FLink`

Dhalton (changes requested):

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -181,11 +175,12 @@ export const Footer: FooterType = (props) => {
   )
 }
 
-export const FooterLink: React.FC<LinkProps> = ({ children, id, to }) => {
+export const FLink: any = (props: any) => {
+  const { id, children, to } = props
```
Dhalton, line 179: No real reason to destructure these props, as long as you just add the `props` to the dependency array for the `useEffect` hook, which I'll point out below.

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -181,11 +175,12 @@ export const Footer: FooterType = (props) => {
...
   const setLinks = useContext(FooterContext)
   const key = useId(id)
   useEffect(() => {
-    setLinks?.((links) => new Map(links.set(key, { children, id, to })))
+    setLinks?.((links) => new Map(links.set(key, props)))
   }, [key, children, id, to, setLinks])
```
Dhalton, line 184: No need to include the individual `children`, `id`, and `to` props in this dependency array anymore, just replace them with `props`.

wilysword:
I have basically the same comments as Dhalton, but I felt that trying to explain them all with words would be longer than just trying to show what I mean: https://github.com/repaygithub/cactus/commit/cfa89ff15e762243ee2b45a3170e255f5943360d

If you like those changes you can just take the whole commit.

Dhalton:
> I felt that trying to explain them all with words would be longer than just trying to show what I mean

That was probably a good call...I was struggling to put my comments into words. If I were you, Danilo, I'd steal this commit 😝

modules/cactus-web/src/Footer/Footer.tsx
wilysword: @Dhalton It still complains, actually, because the `as` prop is a pain no matter what you do. Easier to just leave it like this.

Dhalton: Huh, yeah I guess you're right. The warning was gone in my editor but I see now that it's back. TS must've just been taking its sweet time. 👍

daniloPenaR:
Thanks for all of your comments ( and commits 😄 ).

modules/cactus-web/src/Footer/Footer.tsx
wilysword: @Dhalton Using `typeof Link` here so Typescript will use the same `as` prop logic that it uses with the Link component.

wilysword (approved):

Dhalton (approved):
Changes LGTM

PR #577, opened 2021-04-27
CACTUS-582 :: Dark Variants

daniloPenaR:
[CACTUS-582](https://repayonline.atlassian.net/browse/CACTUS-582)


modules/cactus-web/src/IconButton/IconButton.tsx
```diff
@@ -61,6 +61,13 @@ const variantMap: VariantMap = {
       color: ${(p): string => p.theme.colors.successDark};
     }
   `,
+  dark: css`
+    color: ${(p): string => p.theme.colors.darkContrast};
+
+    &:hover {
+      color: ${(p): string => p.theme.colors.darkestContrast};
+    }
```
mikesoltow, line 69: Setting the hover to `darkestContrast` doesn't lead to much change on hover, as the colors are too close together.  How would you feel about using `callToAction`?   That's currently what we use for `Breadcrumb.Link`.

modules/cactus-web/src/IconButton/IconButton.story.tsx
```diff
@@ -127,38 +153,61 @@ export const BasicUsage = (): React.ReactElement => {
...
+        </Grid.Item>
+      )}
+      <IconButtonBase size="large" disabled isTiny={isTiny} />
     </Grid>
   )
 }
 
+export const BasicUsage = (): React.ReactElement => {
+  return (
+    <StyleProvider theme={cactusTheme} global>
```
mikesoltow, line 205: Can you remove the `StyleProvider` from here?  This overrides the ability to use the Cactus Theme tab to play with the theme colors on this story.  All stories are already wrapped in a provider.

mikesoltow (changes requested):
Just a few comments, but otherwise it looks good.

I'm wondering if all the random updated stories have to do with the new `node` docker image.  That's the only thing that would make sense (plus, it seems like only "default viewport" screenshots updated, so maybe the new image's "default viewport" has changed 🤷 ).

modules/cactus-web/src/IconButton/IconButton.story.tsx
daniloPenaR: Just have one question, if I drop the StyleProvider here, the ScreenSizeProvider will keep useful? I read on the docs that the screenSizeProvider must be nested inside the StyleProvider to show the size correctly.

> The possible sizes are directly from the media queries object ("small", "medium", "large", "extraLarge"), with the addition of "tiny" when none of the media queries match. Because it uses `theme.mediaQueries`, it should be nested inside a `StyleProvider`; otherwise it will always give the default size of "large".

And when I was implementing it, the ScreenSizeProvider was throwing me always the same size

wilysword (approved):

Dhalton:
LGTM just waiting for the build to finish before I approve

Dhalton (approved):

PR #580, opened 2021-04-27
CACTUS-586 Prevent scroll propagating from drop-downs

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-586

This could also reasonably be expected to apply to Modal; however, depending on the exact configuration of the modal, the scrolling may happen at the overlay, inside the modal, or perhaps even inside the contents. Given the difficulty predicting what the dev might want, I decided to export the low-level implementation functions and let them figure out what they want from their modal's behavior. If it becomes an issue, we can always fix it later; but at that point I'd probably want to remove the internal scroll container in favor of the dev providing their own (if they need it), which would be a breaking change.


Dhalton (approved):
I wasn't able to reproduce the problem with the DateInput, but I was able to with the Select, and this fixed the issue there, so I'll go ahead and approve. Good idea to export the hook/function.

daniloPenaR (approved):
I cannot reproduce it in the `DateInput` element too. But tested it on the `Select` component and this works.

PR #582, opened 2021-04-30
CACTUS-598 :: Breadcrumbs Ellipsis

daniloPenaR:
[CACTUS-598](https://repayonline.atlassian.net/browse/CACTUS-598)

I made many tests just to make sure the ellipsis is being activated when the screen size requires it. But for some reason, I distrust the object returned by the `getBoundingClientRect()` function, please let me know if you find something weird...


modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -83,6 +89,17 @@ const BreadcrumbBase = (props: BreadcrumbProps): React.ReactElement => {
     [toggle, setFocus]
   )
 
+  React.useLayoutEffect(() => {
+    if (
+      (pivotWidth && parentNavContainerWidth && pivotWidth > parentNavContainerWidth) ||
```
Dhalton, line 94: This condition should probably be checking `pivotWidth >= parentNavContainerWidth` instead of just `>`.

modules/cactus-web/src/Breadcrumb/Breadcrumb.story.tsx
```diff
@@ -53,24 +57,38 @@ CustomItemElements.parameters = {
   },
 }
 
+export const HugeListOfBreadcrumbs = (): ReactElement => (
+  <Flex>
```
Dhalton, line 61: Why wrap these in a `Flex`? All that's doing for me is preventing the component from realizing that it has enough space when you size back up from something like iPad view to desktop view.

Dhalton (commented):
Something is definitely off, but I haven't quite been able to pinpoint the source of the problem yet. It seems like there's a certain screen size, about at iPad view width, where the component isn't switching to the ellipsis version when it should be. See screenshot:

image

Note that I added an extra link to the "Huge list of breadcrumbs" story to cause this to happen.

It looks like, for some reason, the value of `pivotWidth` is less than what it is in reality. Again, I haven't been able to figure out why or how to fix it yet, though. I'll keep playing around with it tomorrow, but wanted to point it out so you're aware.

modules/cactus-web/src/Breadcrumb/Breadcrumb.story.tsx
Dhalton: If you want to keep the flex I'd say at least give it a fixed width, but honestly I don't know why it's necessary at all.

daniloPenaR: You're right, wrapped the Breadcrumbs into `Flex` from my first commit when I was trying to solve these issues in a different way, and I forgot to get rid of those wrappers with the newer approach. I will not preserve those wrappers, thanks for point it out.

modules/cactus-web/src/Layout/Layout.story.tsx
```diff
@@ -162,6 +163,16 @@ export const BasicUsage = (): React.ReactElement => {
       {hasActions && <StoryActionBar />}
       <Layout.Content>
         <Header bgColor={useDarkNav ? 'lightContrast' : 'white'}>
+          <Header.BreadcrumbRow>
+            <Breadcrumb>
+              {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13].map((i) => (
+                <Breadcrumb.Item href="/" key={i}>{`label ${i}`}</Breadcrumb.Item>
+              ))}
+              <Breadcrumb.Item>
```
Dhalton, line 171: This should be
```
<Breadcrumb.Item as={Link} to="/">
  Make a Payment
</Breadcrumb.Item>
```
otherwise you get console errors regarding having an `<a />` tag as a child of another `<a />` tag.

modules/cactus-web/src/Header/Header.story.tsx
```diff
@@ -59,16 +59,48 @@ export const WithBreadcrumbs = (): React.ReactElement => {
...
+export const WithALotOfBreadCrumbs = (): React.ReactElement => {
+  const bgSelection = select('Background-color', bgColorVariants, 'lightContrast')
+
+  return (
+    <Header bgColor={bgSelection}>
+      <Header.Title>{text('Header Title', 'Header')}</Header.Title>
+      <Header.BreadcrumbRow>
+        <Breadcrumb>
+          {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17].map((i) => (
+            <Breadcrumb.Item href="/">{`label ${i}`}</Breadcrumb.Item>
```
Dhalton, line 72: Need to set a `key` prop on these. Getting console errors about it.

modules/cactus-web/src/Breadcrumb/Breadcrumb.story.tsx
```diff
@@ -53,6 +53,18 @@ CustomItemElements.parameters = {
   },
 }
 
+export const HugeListOfBreadcrumbs = (): ReactElement => (
+  <Breadcrumb>
+    {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13].map((i) => (
+      <Breadcrumb.Item key={i} href="/">{`label ${i}`}</Breadcrumb.Item>
+    ))}
+    <Breadcrumb.Item>
```
Dhalton, line 61: This should be
```
<Breadcrumb.Item as={Link} to="/">
  Make a Payment
</Breadcrumb.Item>
```
otherwise you get console errors regarding having an `<a />` tag as a child of another `<a />` tag.

modules/cactus-web/src/Breadcrumb/Breadcrumb.story.tsx
```diff
@@ -53,6 +53,18 @@ CustomItemElements.parameters = {
   },
 }
 
+export const HugeListOfBreadcrumbs = (): ReactElement => (
+  <Breadcrumb>
+    {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13].map((i) => (
+      <Breadcrumb.Item key={i} href="/">{`label ${i}`}</Breadcrumb.Item>
```
Dhalton, line 59: Need to set a `key` prop on these. Getting console errors about it.

Dhalton (changes requested):
I'm almost ready to approve. I think my only concerns are with the stories.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -83,6 +86,26 @@ const BreadcrumbBase = (props: BreadcrumbProps): React.ReactElement => {
     [toggle, setFocus]
   )
 
+  const checkEllipsisVersion = React.useCallback(() => {
+    setEllipsisVersion(false)
+    const parentWidth = mainNavContainer.current?.parentElement?.getBoundingClientRect().width
+    const pivotWidth = pivotBreadcrumb.current?.getBoundingClientRect().width
+    if ((pivotWidth && parentWidth && pivotWidth >= parentWidth) || (isTiny && childrenCount > 2)) {
+      setEllipsisVersion(true)
+    }
```
wilysword, line 95: Couple of changes I recommend in this function:
```suggestion
    const parentWidth = Math.ceil(mainNavContainer.current?.parentElement?.getBoundingClientRect().width)
    const pivotWidth = Math.ceil(pivotBreadcrumb.current?.getBoundingClientRect().width)
    const ellipsisVersion = pivotWidth >= parentWidth || (isTiny && childrenCount > 2)
    setEllipsisVersion(ellipsisVersion)
```
The `Math.ceil` is to (hopefully) avoid rounding/truncation errors. Then instead of setting the state twice, I think you only really need to set it once.

modules/cactus-web/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -83,6 +86,26 @@ const BreadcrumbBase = (props: BreadcrumbProps): React.ReactElement => {
...
+  React.useLayoutEffect(() => {
+    checkEllipsisVersion()
+    const handleResize = () => {
+      checkEllipsisVersion()
+    }
+    window.addEventListener('resize', handleResize)
+    return () => {
+      window.removeEventListener('resize', handleResize)
+    }
+  }, [checkEllipsisVersion])
```
wilysword, line 107: It would be better to implement the `checkEllipsisVersion` function in here, since this is where it's being used:
```suggestion
  React.useLayoutEffect(() => {
    const checkEllipsisVersion = () => {
      // ...etc
    }
    checkEllipsisVersion()
    window.addEventListener('resize', checkEllipsisVersion)
    return () => {
      window.removeEventListener('resize', checkEllipsisVersion)
    }
  }, [childrenCount, isTiny])
```

wilysword (commented):
Just a couple of comments about code organization, plus what Dhalton said about story console warnings. Looks good, functionally.

wilysword (approved):

Dhalton (approved):

PR #583, opened 2021-05-06
CACTUS-603: fix floating point errors in `useScroll` hook

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-603

As best I can tell, everything in that function is now an integer. Integer math banzai!

### DIFF
--- a/modules/cactus-web/src/helpers/scroll.ts
+++ b/modules/cactus-web/src/helpers/scroll.ts
@@ -66,10 +66,13 @@ interface Rect {
   width: number
 }
 
-const getRect = (e: Element): Rect => e.getBoundingClientRect()
+const getRect = (e: Element): Rect => {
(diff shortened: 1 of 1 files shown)


modules/cactus-web/src/helpers/scroll.ts
```diff
@@ -66,10 +66,13 @@ interface Rect {
   width: number
 }
 
-const getRect = (e: Element): Rect => e.getBoundingClientRect()
+const getRect = (e: Element): Rect => {
```
Dhalton, line 69: Get riggity riggity rect, son! (Rick & Morty reference, had to point it out 😝 )

Dhalton (approved):
LGTM, tested in IE & Chrome and it definitely fixes the problem.

daniloPenaR (approved):

PR #588, opened 2021-05-11
CACTUS 608:: Switch ConfirmModal Buttons

daniloPenaR:
[CACTUS-608](https://repayonline.atlassian.net/browse/CACTUS-608)


modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -75,10 +75,10 @@ const ConfirmModalBase: React.FunctionComponent<ConfirmModalProps> = ({
         {children}
       </Flex>
       <Flex justifyContent="space-between" marginTop="40px" marginBottom={2}>
+        <TextButton onClick={onClose}>{cancelButtonText}</TextButton>
```
Dhalton, line 78: We still need to maintain the spacing between the buttons. Gotta remove the `marginRight="24px"` from `Button` below this line and add it onto this `TextButton` instead. While you're at it you might as well just use `mr={5}`, which does the same thing but has the added benefit of staying in-line with the design system if the spaces we have defined were to ever change.

Dhalton (changes requested):
Just the comment about the margin and then I'm good to approve.

Dhalton (approved):
LGTM thanks for making those changes

wilysword (approved):

PR #590, opened 2021-05-12
CACTUS-597: Prop Forwarding & `data-testid`

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-597

I did a commit per component instead of my usual squash, so it could be helpful to go through this one commit at a time. Some of the commits also have fairly detailed commit messages where things weren't completely straightforward, so you can also read those to get more info.

Kind of weird how the components needing to be changed are all bunched up at the beginning of the alphabet: made me think this ticket was going to take a lot longer than it ended up taking.


Dhalton (approved):

daniloPenaR (approved):
Awesome

PR #591, opened 2021-05-12
CACTUS-609 :: Safari MenuBar Fix

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-609

Honestly, I still don't really know exactly what was causing this. I think it was something to do with `AceEditor`, because it was only happening on pages that use it, but that's about all I could figure out. John seemed to think it was related to `transform: translateZ(0);` in the `ace_cursor` class, but I didn't see evidence of that. I do know, however, that this fixes it. 🤷‍♂️ 

Also, I had to increase the z-index on the user menu dropdown in `BrandBar`, so that it renders above the `MenuBar`.

### Testing
1. Check out this branch and run `yarn web cleanup && yarn web build`.
2. Open up the `docker-compose.yml` file in `repay-api` and add this line to the `volumes` section under `admin-ui`:
`- ../cactus/modules/cactus-web/dist:/code/node_modules/@repay/cactus-web/dist`
3. Run `make init` to build & run Channels locally.
4. When that's done building, open Safari and go to `https://owe.channels.dev.repay.localhost/admin`.
5. Log in and navigate to the UI Config page under "Merchant Config".
6. Make sure the MenuBar dropdowns open when you click them.
7. Make sure you can log out using the user menu dropdown.
8. Repeat steps 4-7 for other browsers like Chrome or FireFox. Testing Channels locally in IE11 is a pain because you get a ridiculous amount of cert errors that pop up and get in the way, but if you want to test there too, be my guest. Or you can just take my word for it when I say it works there.

### DIFF
--- a/modules/cactus-web/src/BrandBar/BrandBar.tsx
+++ b/modules/cactus-web/src/BrandBar/BrandBar.tsx
@@ -245,6 +245,7 @@ const MenuList = styled(ReachMenuList)`
   background-color: ${(p) => p.theme.colors.white};
 
   [data-reach-menu-item] {
+    z-index: 110;
     box-sizing: border-box;
     position: relative;
     display: block;
--- a/modules/cactus-web/src/MenuBar/MenuBar.tsx
+++ b/modules/cactus-web/src/MenuBar/MenuBar.tsx
@@ -372,6 +372,7 @@ const Nav = styled.nav<MenuBarProps>`
   align-items: stretch;
   position: relative;
   outline: none;
+  z-index: 100;
   [role='menubar'] > li > [role='menuitem'] {
     white-space: nowrap;
     padding: 24px 16px;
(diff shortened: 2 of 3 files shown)


daniloPenaR:
```- ../cactus/modules/cactus-web/dist:/code/node_modules/@repay/cactus-web/dist```

what is the main function of adding this into the volumes section in the `docker-compose` file?

I assume that this is to link the current local changes in our cactus-web repo. But I would like to read a better explanation (If it has)

daniloPenaR (approved):
Tested it and it seems to work in Chrome, Firefox and Safari.

Dhalton:
> what is the main purpose of adding this into the volumes section in the `docker-compose` file?
> 
> I assume that this is to link the current local changes in our cactus-web repo. But I would like to read a better explanation (If it has)

Sure, so mounting a volume on a docker container is a way to give the container access to files on your machine that it wouldn't normally have access to. The part before the `:` is the path you want to take files from and everything after the `:` is telling the container where to save that data inside the container. So you can see, in the first half (`../cactus/modules/cactus-web/dist`) we're pointing to our local build of cactus-web, and the second half (`/code/node_modules/@repay/cactus-web/dist`) we're telling the container to put that local build in the `node_modules` folder inside the container. This is how we tell the `admin-ui` container to use the local version of cactus-web instead of what's downloaded from npm, allowing you to test the fix. This is also the reason you have to run `yarn web build` beforehand -- because we need to build the cactus-web distributable so it can be pulled into the container.

Hope that makes sense

wilysword (approved):

PR #592, opened 2021-05-19
CACTUS-616 :: Header Description

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-616

I'm still not sure if the `text` prop really provides any value. I just added it because I thought it might be nice to offer a couple of different ways of setting it up, but maybe it's not even worth including it. I dunno, I could go either way. Let me know what you guys think.

I didn't add any unit tests 'cause I figure the storyshots should cover it anyway.

The Header docs were pretty broken -- no way to try it out on the page, props tables passing the wrong `displayName`s, that sort of thing. I fixed those up, though.

Still looks really bad in IE on mobile, but I guess that's just something we're living with, for now anyway.

### Testing
1. Pull down this branch and run the storybooks
2. Check out the "With Description" and "With Everything" stories. Make sure it all looks good. Add some text to make sure it wraps properly.
3. Change the screen size to make sure it looks good at all screen sizes.
4. Check it out in other browsers. Don't expect it to look good in IE/mobile.


modules/cactus-web/src/Header/Header.tsx
```diff
@@ -10,10 +10,15 @@ interface HeaderProps extends MarginProps, HTMLAttributes<HTMLDivElement> {
   bgColor?: BackgroundColorVariants
 }
 
+interface HeaderDescriptionProps extends HTMLAttributes<HTMLDivElement> {
+  text?: string
+}
+
```
daniloPenaR, line 16: About the `text` prop, well, the rest of the subcomponents of the `Header` receive the `children` prop at all. But I think this approach is very interesting.  are there other components that are actually using the `text` prop (or something similar)? That could be something like a standard across our components (the ones that require it). This prop looks fine from my perspective, if you wanna preserve it, fine by me.

Dhalton: I think there might be a couple of other places we do stuff like this where we give them a couple of options, I don't know if it'd be applicable enough across the board to make it like a standard though.

daniloPenaR (approved):
LGTM.

mikesoltow (approved):
LGTM.  FWIW, I'm not sure if using the `text` prop serves a purpose...but maybe I'm missing something?  If there's not a use case that the prop solves, my vote would be to remove it 🤷

Dhalton:
> I'm not sure if using the `text` prop serves a purpose...but maybe I'm missing something? If there's not a use case that the prop solves, my vote would be to remove it 🤷

Yeah, it really doesn't provide any extra functionality. I'll remove it tomorrow.

mikesoltow (approved):

PR #594, opened 2021-05-21
CACTUS-613 :: Flex Item Props

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-613

A couple of points of discussion:

I didn't allow the props on `Table` or `DataGrid`. My thinking was that these components are typically meant to take up the majority of the screen's width. I couldn't really think of any use-cases for them to be rendered inside a `Flex` and their width to be adjusted, but I'm open to adding the props on them if you guys can see a reason to do so.

I went back and forth a lot when deciding whether or not we should add the props to the form field components (things like `TextInputField`, `TextAreaField`, `SelectField`, `DateInputField`, etc.). Ultimately I decided not to, since most use-cases I've seen either involve the fields taking up the entire width of the form or the form being sectioned off into a couple of columns. In either of those cases, though, I feel that they'd probably always be rendered inside block-level components like `Card` or `Box`, which _do_ now support the flex item props and therefore shouldn't really need them. However, I would like some opinions on this. If you guys think it'd be worth the effort, I can go back through and add them there too.

Also, for some reason it looked like the snapshots for the `FileInput`/`FileInputField` needed to be updated...no idea why that is since this branch is up-to-date with master.

### Testing
I didn't add or update any stories for these props. To be honest, I don't think it'd be worth it to do that for each of the components I updated. If you want to test them out you can render the components with updates inside a `Flex` and play around with the props that were added.

### DIFF
--- a/modules/cactus-web/src/Accordion/Accordion.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.tsx
@@ -15,17 +15,9 @@ import React, {
   useState,
 } from 'react'
 import styled, { css, StyledComponentBase } from 'styled-components'
-import {
-  flexbox,
-  margin,
-  MarginProps,
-  maxWidth,
-  MaxWidthProps,
-  width,
-  WidthProps,
-} from 'styled-system'
+import { margin, MarginProps, maxWidth, MaxWidthProps, width, WidthProps } from 'styled-system'
 
-import { FlexItemProps } from '../helpers/flexItem'
+import { flexItem, FlexItemProps } from '../helpers/flexItem'
 import KeyCodes from '../helpers/keyCodes'
 import { omitMargins, omitProps } from '../helpers/omit'
 import { boxShadow, radius } from '../helpers/theme'
@@ -775,7 +767,7 @@ export const Accordion = styled(AccordionBase).withConfig(
   ${margin}
   ${width}
   ${maxWidth}
-  ${flexbox}
+  ${flexItem}
 ` as any
 
 Accordion.defaultProps = {
--- a/modules/cactus-web/src/Accordion/Accordion.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.tsx
@@ -678,14 +678,13 @@ const AccordionBase = (props: AccordionProps): ReactElement => {
     }
   }
 
-  const rest = omitMargins(restProps, 'width', 'maxWidth')
   const open = isManaged ? getManagedOpen() : unmanagedState.isOpen
 
   return (
     <div
       id={id}
       className={`${className} ${open && variant === 'outline' ? 'box-shadow' : ''}`}
-      {...rest}
+      {...restProps}
     >
       <AccordionContext.Provider
         value={{
@@ -754,7 +753,7 @@ const variantStyles = (props: AccordionProps): ReturnType<typeof css> | undefine
 }
 
 export const Accordion = styled(AccordionBase).withConfig(
-  omitProps<AccordionProps>('flex', 'flexBasis', 'flexGrow', 'flexShrink')
+  omitProps<AccordionProps>(flexItem, margin, 'width', 'maxWidth')
 )`
   box-sizing: border-box;
   width: 100%;
--- a/modules/cactus-web/src/Accordion/Accordion.tsx
+++ b/modules/cactus-web/src/Accordion/Accordion.tsx
@@ -15,10 +15,19 @@ import React, {
   useState,
 } from 'react'
 import styled, { css, StyledComponentBase } from 'styled-components'
-import { margin, MarginProps, maxWidth, MaxWidthProps, width, WidthProps } from 'styled-system'
+import {
+  flexbox,
+  margin,
+  MarginProps,
+  maxWidth,
+  MaxWidthProps,
+  width,
+  WidthProps,
+} from 'styled-system'
 
+import { FlexItemProps } from '../helpers/flexItem'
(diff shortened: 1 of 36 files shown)


mikesoltow:
Just starting my review now, but my first thought is that I agree with you about the Table and form field components.  I'm not strongly opposed to adding them if anyone feels differently, but I agree that there doesn't seem to be much need.

modules/cactus-web/src/FileInput/__snapshots__/FileInput.test.tsx.snap
```diff
@@ -13,7 +13,17 @@ exports[`component: FileInput should render a disabled file input 1`] = `
 
 .c4 {
   box-sizing: border-box;
+  -webkit-flex-direction: column;
+  -ms-flex-direction: column;
+  flex-direction: column;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
+  align-items: center;
```
mikesoltow, line 22: My guess is that if you restricted the flex properties as in my previous comment, this would go away...but I haven't been able to pinpoint why this is happening yet, so I could be wrong.

mikesoltow (commented):
Looks pretty good.  Just one comment/question.

wilysword:
I think the `*Field` components should support these props; there's at least one place in channels (the rules form) where I applied flex styles to the `FieldWrapper` component (which is basically a block-level component), and I think anything that uses `FieldWrapper` should also support them.

mikesoltow (approved):
Updates look good to me.  Thanks!

modules/cactus-web/src/Accordion/Accordion.tsx
```diff
@@ -751,7 +753,9 @@ const variantStyles = (props: AccordionProps): ReturnType<typeof css> | undefine
   }
 }
 
-export const Accordion = styled(AccordionBase)`
+export const Accordion = styled(AccordionBase).withConfig(
+  omitProps<AccordionProps>('flex', 'flexBasis', 'flexGrow', 'flexShrink')
```
wilysword, line 757: You can just pass the styled-system function here. In fact, you can also pass `margin`, `'width'`, and `'maxWidth'` and get rid of the `omitMargins` call on line 682. (Not a big deal if you don't, though.)
```suggestion
  omitProps<AccordionProps>(flexItem, margin, 'width', 'maxWidth')
```

modules/cactus-web/src/Alert/Alert.tsx
```diff
@@ -6,14 +6,15 @@ import styled, { css, ThemeProps } from 'styled-components'
...
+import { flexItem, FlexItemProps } from '../helpers/flexItem'
 import { omitProps } from '../helpers/omit'
 import { boxShadow } from '../helpers/theme'
 import IconButton from '../IconButton/IconButton'
 
 export type Status = 'error' | 'warning' | 'info' | 'success'
 export type Type = 'general' | 'push'
 
-interface AlertProps extends React.HTMLAttributes<HTMLDivElement> {
+interface AlertProps extends FlexItemProps, React.HTMLAttributes<HTMLDivElement> {
```
wilysword, line 17: If you're using this with `omitProps`, I think you should add this to `AlertStyleProps` instead. (I feel like it should be causing a Typescript error when you pass `...rest` to the div in AlertBase; not sure what's going on there.)

modules/cactus-web/src/DateInputField/DateInputField.tsx
```diff
@@ -54,6 +58,10 @@ function DateInputFieldBase(props: DateInputFieldProps): React.ReactElement {
       autoTooltip={autoTooltip}
       disableTooltip={disableTooltip}
       alignTooltip={alignTooltip}
+      flex={flex}
+      flexGrow={flexGrow}
+      flexShrink={flexShrink}
+      flexBasis={flexBasis}
```
wilysword, line 64: It's beyond the scope of this ticket, but AccessibleField also supports the margin props, I wonder if in the long term we might want to let it handle the margin props as well?

wilysword (commented):
Would `List` be considered a block-level element?

Looks good, I'll approve once ^ is answered.

Dhalton:
> Would `List` be considered a block-level element?

I never really thought of it as one, but I mean there's no harm in supporting the flex item props, so if there's a potential use-case for it, I'll just do it.

modules/cactus-web/src/Accordion/Accordion.tsx
Dhalton: Ah, yeah the styled-system function didn't exist yet at the time that I made these changes. I'll update it.

modules/cactus-web/src/DateInputField/DateInputField.tsx
Dhalton: To be honest, I'm not sure what you mean. Did you mean to say that in the long term we might want it to handle flex item props? It already handles margin props, right?

wilysword: `styled(DateInputFieldBase)` includes calls to `margin` and `width`, so those are processed and passed on as part of the `className` prop to AccessibleField. However, AccessibleField _also_ supports `margin`, so instead of processing the styles in DateInputField, we could forward the `margin` props to AccessibleField and let it process them--which is what you're doing with the `flexItem` props, and what led to my speculation.

wilysword (approved):

PR #597, opened 2021-05-26
Cactus 604 :: usermenuitem

daniloPenaR:
[CACTUS-604](https://repayonline.atlassian.net/browse/CACTUS-604)

- got rid of all the Reach elements in the Brandbar. 
- Implemented the `usePopup` hook to replace the Reach library
- Changed the UserMenuItem component for a similar one from MenuBar
- Now, the UserMenuItem component allows the `as` prop and it can be used as a `Link`
- Finally, breaking change: `onSelect` prop is no longer available for the UserMenuItem, use `onClick` instead

Thanks to Glen for all the support.


modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -242,15 +287,21 @@ const MenuButton = styled(ReachMenuButton)<ProfileStyleProp>`
...
   margin-top: 8px;
-  outline: none;
   border-radius: ${radius(8)};
   ${(p) => boxShadow(p.theme, 1) || `border: ${border(p.theme, 'lightContrast')}`};
   background-color: ${(p) => p.theme.colors.white};
+  list-style: none;
 
-  [data-reach-menu-item] {
+  [role='menuitem'] {
     z-index: 110;
```
wilysword, line 305: Should move this z-index style up to the `MenuList` styles instead of just the menuitem styles.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -263,13 +314,17 @@ const MenuList = styled(ReachMenuList)`
     outline: none;
     padding: 4px 16px;
     text-align: center;
-    &[data-selected] {
+    &:focus,
+    &:hover {
```
wilysword, line 318: Focus and hover styles need to be different. For hover, try this instead:
```
&:hover {
  color: ${(p) => p.theme.colors.callToAction};
}
```

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -120,59 +129,95 @@ const ActionBarUserMenu: React.FC<UserMenuProps> = ({
...
+
+const handleArrows = (event: React.KeyboardEvent<HTMLElement>, toggle: TogglePopup) => {
+  if (event.key === 'ArrowDown') {
+    toggle(true, 1, { shift: true })
+  } else if (event.key === 'ArrowUp') {
+    toggle(true, -1, { shift: true })
+  }
+}
+
+const positionPopup = (menu: HTMLElement, menuButton: HTMLElement | null) => {
```
wilysword, line 181: This position function is something I whipped up for the POC, but it's too naive for the final product: it doesn't handle scrolling or menu items that are wider than the button. This one should be a bit more up to the task (though you may still need to make modifications): https://github.com/repaygithub/cactus/commit/46d75df290e05ef6d10df5afa8e9a768d9cdd2a8

wilysword (changes requested):

modules/cactus-web/src/BrandBar/BrandBar.tsx
daniloPenaR: @wilysword  So here you are referring to the cases we have a huge menu list wider than the viewport height?

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -120,59 +129,95 @@ const ActionBarUserMenu: React.FC<UserMenuProps> = ({
   label,
   children,
   isProfilePage,
+  id = 'user-menu',
```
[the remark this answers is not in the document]
daniloPenaR: I guessed `id` is optional and we are passing it into the options object of the usePopup it was pretty much convenient to have a default. But I'm going to remove it if you think it's not necessary.

modules/cactus-web/src/BrandBar/BrandBar.tsx
wilysword: I don't think so, though I don't quite understand your question. I'm referring to two scenarios, mainly:
1. In the BrandBar -> Basic Usage story, there's a knob called "Action one". Enter some long text there, and see how the menu behaves.
2. In the Layout -> Basic Usage story, try opening the user menu and then scrolling down the page.

modules/cactus-web/src/BrandBar/BrandBar.tsx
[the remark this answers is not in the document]
wilysword: `usePopup` will generate a random ID if none is provided, so you don't need the hard-coded default.

modules/cactus-web/src/BrandBar/BrandBar.tsx
daniloPenaR: Gotcha, Thanks, well I am currently testing those new changes in order to find anything in those cases that requires any other modification.

modules/cactus-web/src/BrandBar/BrandBar.tsx
wilysword: Still didn't fix this one.

wilysword (commented):
Where did those `user-menu-wider-list` snapshots come from? There doesn't appear to be an associated story, so you should probably delete them. Aside from that and the hover styles issue, though, I think this is looking good.

mikesoltow (changes requested):
None of my comments in the code are blockers, but I marked this as "Request Changes" because of the following issues:

* If you click to open the menu, you can then use the arrows to navigate
* In Safari, a weird outline appears when you click to open the menu.  You need to style the `div` that takes `wrapperProps` with `outline: none`.
* The `:visited` style is different in the mobile menu.  I don't think it should change, as it looks weird to have one menu item styled differently than the others.
* If you change a menu item to be something really long, it works fine on desktop/tablet but overflows on mobile.
* You should be able to hide a menu using ESC, but that doesn't seem to work.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -1,23 +1,18 @@
...
 import { OrderHint, useAction } from '../ActionBar/ActionProvider'
 import Flex from '../Flex/Flex'
-import { getTopPosition } from '../helpers/positionPopover'
+import { keyDownAsClick, preventAction } from '../helpers/a11y'
+import { AsProps, GenericComponent } from '../helpers/asProps'
+import { getViewport, usePositioning } from '../helpers/positionPopover'
 import { border, boxShadow, insetBorder, radius, textStyle } from '../helpers/theme'
+import usePopup, { TogglePopup } from '../helpers/usePopup'
 import { Sidebar } from '../Layout/Sidebar'
+import { SidebarMenu as ActionMenuPopup } from '../MenuBar/MenuBar'
```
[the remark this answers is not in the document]
daniloPenaR: Did you mean, preserving the same import name `SidebarMenu`

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -120,59 +130,108 @@ const ActionBarUserMenu: React.FC<UserMenuProps> = ({
...
+    anchorRef: buttonRef,
+    updateOnScroll: true,
+  })
+  wrapperProps.onClick = React.useCallback(
+    (event: React.MouseEvent<HTMLDivElement>) => {
+      const target = event.target as HTMLDivElement
+      if (target.matches('[role="menuitem"], [role="menuitem"] *')) {
+        toggle(false, buttonRef.current)
       }
     },
```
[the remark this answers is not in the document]
daniloPenaR: You're right, Glen recommended to just put into the `popupProps`

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -38,23 +34,38 @@ interface ProfileStyleProp {
...
+    ? preventAction
+    : (e: React.KeyboardEvent<HTMLElement>) => {
+        original(e)
+        preventAction(e)
+      }
+  return (
+    <li role="none">
+      <MenuListItem {...propsCopy} tabIndex={-1} role="menuitem" ref={ref as any} />
+    </li>
+  )
```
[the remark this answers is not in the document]
daniloPenaR: But here in Brandbar we need to add the `tabIndex` property to enable the navigation with the arrow keys.

modules/cactus-web/src/BrandBar/BrandBar.test.tsx
```diff
@@ -53,10 +53,6 @@ describe('component: BrandBar', () => {
       fireEvent.keyDown(document.activeElement, { key: 'ArrowDown' })
       await animationRender()
 
-      // @ts-ignore
-      fireEvent.keyDown(document.activeElement, { key: 'ArrowDown' })
-      await animationRender()
-
```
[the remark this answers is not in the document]
mikesoltow: But why did this change?  Before we needed to use the arrow twice?

modules/cactus-web/src/BrandBar/BrandBar.tsx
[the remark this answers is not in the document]
mikesoltow: That happens in `MenuBar.tsx` as well, it just happens in the `MneuButton` component itself.  See https://github.com/repaygithub/cactus/blob/master/modules/cactus-web/src/MenuBar/MenuBar.tsx#L555

modules/cactus-web/src/BrandBar/BrandBar.tsx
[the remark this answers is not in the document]
mikesoltow: No, I was thinking that, if we created a helper file for the shared menu item stuff, then we put SidebarMenu in that file as well, since it's shared.  It makes more sense to me to have two components import it from a helper file than to have one component import it from another component's file...when those two components don't have a hierarchical relationship or anything like that.

It's not strictly necessary, but just a suggestion for organization.

modules/cactus-web/src/BrandBar/BrandBar.story.tsx
```diff
@@ -91,6 +73,6 @@ export const CustomItems = (): React.ReactElement => {
 
 CustomItems.parameters = {
   beforeScreenshot: async (page: Page) => {
-    await page.click('[id="user-menu-button"]')
+    await page.click(`[id="user-menu-button"]`)
```
mikesoltow, line 76: What does this change do?

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -130,13 +130,14 @@ const ActionBarUserMenu: React.FC<UserMenuProps> = ({
...
   isProfilePage,
-  id = 'user-menu',
+  id,
   ...rest
 }) => {
   const { buttonProps, toggle, popupProps, wrapperProps } = usePopup('menu', {
     id,
     focusControl,
     onWrapperKeyDown: handleArrows,
+    buttonId: 'user-menu-button',
```
mikesoltow, line 140: What does adding a `buttonId` do?

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -337,6 +337,13 @@ const ActionMenuList = styled.ul`
   padding: 0;
   margin: 0;
   outline: none;
+  word-break: break-word;
+  [role='menuitem'] {
+    &:focus,
+    &:hover {
+      ${(p) => p.theme.colorStyles.callToAction};
+    }
+  }
```
mikesoltow, line 346: This doesn't fix the problem, as the focus and hover styles need to be different.  To see an example, open up a MenuBar dropdown and scroll to an item with the arrow keys.  Then hover on the item above it with your mouse.  Notice how the two have different styles.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -337,6 +337,13 @@ const ActionMenuList = styled.ul`
   padding: 0;
   margin: 0;
   outline: none;
+  word-break: break-word;
```
mikesoltow, line 340: Remember that this doesn't work in IE11.  There are a couple of other places in this file that should be corrected as well.

```suggestion
  word-wrap: break-word;
  overflow-wrap: break-word;
```

mikesoltow (changes requested):
Looks better!  Just a few more items to address.

Also, it looks like the `:visited` style still hasn't been updated:
>The :visited style is different in the mobile menu. I don't think it should change, as it looks weird to have one menu item styled differently than the others.

modules/cactus-web/src/BrandBar/BrandBar.story.tsx
daniloPenaR: Oh sorry, I was performing some changes and playing around with the screenshot's parameters and when I wrote that again I put them on that way, going to undo that now.

modules/cactus-web/src/BrandBar/BrandBar.tsx
daniloPenaR: This `buttonId` adds an `id` property for the element that contains the `buttonProps`. If you don't define a name for that id, the Hook will add a random one, and this line
```javascript
await page.click('[id="user-menu-button"]')
```
in the story will throw us an error.

wilysword: You should still find a different selector; maybe add a className, if it doesn't already have a unique one (from `styled`)? The reason we don't want to hard-code IDs is because they're supposed to be unique. Although this specific component shouldn't appear more than once on the page, what if the dev wanted to use that ID somewhere in the page content?

wilysword (commented):
Styles on larger screen sizes look better, but the `:hover`/`:focus`/`:visited` styles on tiny/mobile screens are still off.

modules/cactus-web/src/BrandBar/BrandBar.tsx
mikesoltow: Ah, good catch Glen.  I agree.

daniloPenaR:
@mikesoltow @wilysword What do you guys think about using `'[role="button"]'` as a selector? My guess is that using className is not a good idea, because we would need to use the entire class, isn't it? just like `class="sc-gGiJkG hvvMSH user-menu-button"`

wilysword:
A classname selector is like `.sc-gGiJkG`, and it can be used with any one of the classes (although you have to know the class, of course, so the only way to use auto-generated styled-components classes is if you have the component itself). You don't have to use that selector, though, you can use any selector as long as the first element it finds is the button you want to click.

daniloPenaR:
@wilysword So is it fine if I preserve that `role` selector?

wilysword:
As long as it works, I'm fine with it.

mikesoltow (commented):
Just one more comment!

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -552,7 +512,7 @@ const buttonStyles = `
   }
 `
 
-const MenuButton = styled.span.attrs({ tabIndex: -1 as number, role: 'menuitem' as string })`
+export const MenuButton = styled.span.attrs({ tabIndex: -1 as number, role: 'menuitem' as string })`
```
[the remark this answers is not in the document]
daniloPenaR: Yeah, I'm fine with that, going to perform those changes immediately.

modules/cactus-web/src/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+  const original = propsCopy.onKeyUp
+  propsCopy.onKeyUp = !original
+    ? preventAction
+    : (e: React.KeyboardEvent<HTMLElement>) => {
+        original(e)
+        preventAction(e)
+      }
+  return (
+    <li role="none">
```
luisniebla, line 63: role=none? Why?

wilysword: So that screen readers ignore it.

wilysword (approved):

luisniebla (approved):

mikesoltow (commented):
It's looking great, but I still see a different `:visited` style on mobile screens.  Is that just me?  (I have this branch rebased off my dependencies branch because of some issues I'm having)

mikesoltow (commented):
Also, one more thing I noticed: the regular MenuBar no longer is displaying the underline on the nested menu items in tablet mode when you hover over them.

wilysword:
I figured out the `:visited` style is actually because in the story it uses `as={Link}` instead of `as="a"`. The `Link` styles are interfering with the menubutton styles.

stevemom-repay (commented):
It doesn't look like there were much of any test changes short of changing from `onSelect` to `onClick`. It seems like we could be adding some additional tests to cover some of these other changes.

daniloPenaR:
> It's looking great, but I still see a different `:visited` style on mobile screens. Is that just me? (I have this branch rebased off my dependencies branch because of some issues I'm having)

@mikesoltow This is how it looks at this moment (with a hover on the Link)
Screen Shot 2021-06-23 at 7 02 04 AM
Screen Shot 2021-06-23 at 7 02 29 AM


And this is how I've modified the default styles of the `:visited` `Link` to inherit the color property
Screen Shot 2021-06-23 at 7 05 58 AM

Do you want to be inherited, or maybe you prefer both mobile and larger screens to show the visited Link with a different color?

daniloPenaR:
> It doesn't look like there were much of any test changes short of changing from `onSelect` to `onClick`. It seems like we could be adding some additional tests to cover some of these other changes.

@stevemom-repay  Maybe I could mock the `onChange` and create some tests, I've learned new stuff about that with my current ticket lol, @mikesoltow what do you think?

mikesoltow:
> I figured out the `:visited` style is actually because in the story it uses `as={Link}` instead of `as="a"`. The `Link` styles are interfering with the menubutton styles.

Ah, you are right.  The same thing will happen on the regular menu bar, so this is not a new problem.

@daniloPenaR for this PR, I'd take out your recent commit (I'm sorry!).  We can deal with this in a separate ticket if we wish (by adding some contextual style to `Link` for when it's wrapped in our new MenuBarItem component)...or just direct users to use `as="a"` instead.

daniloPenaR:
@mikesoltow Gotcha, going to undo those changes and to solve de `MenuBar` `border-bottom` issue.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -228,49 +257,71 @@ const MenuButton = styled(ReachMenuButton)<ProfileStyleProp>`
...
+const ActionMenuList = styled.ul`
+  list-style: none;
+  padding: 0;
+  margin: 0;
   outline: none;
+  word-wrap: break-word;
+  overflow-wrap: break-word;
+  [role='menuitem'] {
+    &:focus {
+      background-color: ${(p) => p.theme.colors.lightContrast};
```
mikesoltow, line 318: Sorry for not catching this before, but you'll want to remove this background color on focus if you want to match the MenuBar styles.

mikesoltow (commented):
Just one more thing!  I thoroughly tested the rest in all browsers, and it looks good.

Regarding @stevemom-repay 's testing suggestion, I think that's a good point.  There don't appear to be a ton of behavior tests for this file, and I think it could also benefit from some more screenshot tests of cases where the menu is open and an item is selected or hovered.

That being said, this ticket has been open for quite a while.  My inclination is just to push it through and then backfill some of the tests in a tech debt ticket.  @stevemom-repay are you okay with that?  Feel free to push back if not, as this is your wheelhouse.

stevemom-repay:
> Just one more thing! I thoroughly tested the rest in all browsers, and it looks good.
> 
> Regarding @stevemom-repay 's testing suggestion, I think that's a good point. There don't appear to be a ton of behavior tests for this file, and I think it could also benefit from some more screenshot tests of cases where the menu is open and an item is selected or hovered.
> 
> That being said, this ticket has been open for quite a while. My inclination is just to push it through and then backfill some of the tests in a tech debt ticket. @stevemom-repay are you okay with that? Feel free to push back if not, as this is your wheelhouse.

I think backfilling with tests at this point is fine.

mikesoltow (approved):
Thanks!  LGTM 🎉

PR #601, opened 2021-06-01
CACTUS-606 :: Remove Default Header Margins

daniloPenaR:
[CACTUS-606](https://repayonline.atlassian.net/browse/CACTUS-606)

- Removed the margins by applying a simple `margin: 0` in the styles of the `Text` component. 
- Also, I got rid of those places where we were setting the `mb`, `m`,  `my` or other related margin props to zero.

I know that the solution came in a very simple way...Nevertheless, I'm curious about something,
In the hypothetical case I really wanted to make a very specific feature by referring to the `as` prop inside the Styled-component's definition, how can I achieve that by implementing the `AsProp` helper?
I mean, that is what I was trying to do:
```javascript 
      export const Text = styled('p')<TextProps & AsProp<GenericComponent>>`
        //...
        ${(p) =>
          ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].includes(p.as as string) &&
          `
              margin: 0;
        `}
      `
```
is there any way to implement the `... & AsProp<C extends GenericComponent='p'>` without crashing the Typecript compilation?


mikesoltow (approved):
This looks good to me!  As to your question, I think the ultimate answer is that you shouldn't need to do that.  In most cases, I imagine you really want to know the ultimate HTML element and not a custom component (i.e., you would want to do something to all `a`s that a user could pass, rather than testing for `Link || 'a'` or something like that).  If that's the case, CSS selectors already give you what you need.

If you _really_ needed it, though, I'm not sure how you'd access it, to be honest.

Dhalton (approved):
LGTM

PR #606, opened 2021-06-10
CACTUS-646: fix stutter on transition with/without scroll buttons

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-646

The problem is that for some reason IE has trouble with a scrollable container inside a flex container. The only way to fix that was to put a max height, but that was causing the flickering when the menu height was between the max and max - buttons. To fix it, I just had to adjust the max to account for whether or not the buttons are visible; in some of my tests that led to a single flicker, but since max-height is the only solution (out of many I tried) that works, and it's IE, I don't think I care.

Also simplified the style a bit.


Dhalton (approved):

mikesoltow (approved):
:+1: Finally tested this.  Looks good

PR #608, opened 2021-06-17
CACTUS-578 :: Mini Table

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-578

### Testing

1. Check out this branch and start storybook
2. Bring up the Table > Layout story.
3. Set the variant to `mini`. Uncheck the `fullWidth` checkbox. Make sure the styles match the style guide. Question: should we still adhere to the `fullWidth` prop if the variant is `mini`?
4. Open the DataGrid > Basic Usage story. Set the variant to `mini` and make sure the styles match what's shown in the style guide.

### DIFF
--- a/modules/cactus-web/src/DataGrid/BottomSection.tsx
+++ b/modules/cactus-web/src/DataGrid/BottomSection.tsx
@@ -3,6 +3,7 @@ import React, { useContext } from 'react'
 import styled, { useTheme } from 'styled-components'
 
 import { cloneAll } from '../helpers/react'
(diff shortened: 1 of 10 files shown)


wilysword (approved):
Looks fine to me; I also see no reason to do anything with the `fullWidth` prop, as it could still be quite meaningful: it fits to the width of the container, so assuming the mini table is inside a smaller container, fullWidth would still be relevant without looking weird.

daniloPenaR (approved):

PR #609, opened 2021-06-24
CACTUS-601 :: BrandBar Dropdown

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-601

I know we're trying to get away from Reach, so I didn't use any Reach components for this, but I didn't feel like it'd be in-scope for me to refactor Reach out of the UserMenu.

### UAT
1. Pull down this branch and run `yarn web start`.
2. Open the "BrandBar -> With Org Dropdown" story.
3. Open the dropdown and make sure it works as you'd expect and that it matches the designs.
4. Repeat step 3 at differing screen sizes.


wilysword:
Couple things just from the visuals:
- When you focus items in the dropdown using the keyboard, they have the default dotted outline. Not sure if we care.
- Do you actually need to align the dropdown when it's open like that? As long as it takes into account the edge of the screen, I'm not sure that's necessary (even if it _is_ shown that way in the style guide).
- On the "With Org Dropdown" story, if you select align -> left, the border in between the two items doubles up. I don't see anything one way or the other in the style guide, but if it's not too hard we should probably fix that.
- Because the logo is taller than the buttons, the vertical borders just kind of end. To be honest I'm not sure what the best solution is: it would look weird for them to extend above the buttons, it would look weirder to center the buttons; maybe the best solution is to leave it how it is now, but I thought I should mention it anyway in case someone else can think of something.
- The drop-down doesn't follow the hover styles from the style guide.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -73,18 +109,59 @@ type BrandBarType = React.FC<BrandBarProps> & {
...
-      <MenuButtonStyles />
       <Flex alignItems="flex-end" justifyContent={justify}>
-        {children}
+        <LogoWrapper>{typeof logo === 'string' ? <img alt="Logo" src={logo} /> : logo}</LogoWrapper>
+        {!isTiny && leftChildren}
       </Flex>
+      <MenuButtonStyles />
+      {isTiny ? (
+        <Flex flexDirection="column" width="100%">
+          <StyledDropdownRow
```
mikesoltow, line 141: This code seems to create an empty row on mobile if there are no BrandBar Items.  You can see it in the screenshot diff for the mobile layout story.  Notice how the bottom border of the BrandBar is now extra thick (because it's really two borders stacked vertically around an empty row).  So we probably don't want to render this if there are no children (either that or just remove the borders somehow).

mikesoltow (commented):
I haven't done a full review yet, but I did notice one thing while looking over the screenshot diffs.

Also, I posted a thread to Slack to get the designers' opinion on the height issue Glen brought up.

Dhalton:
@wilysword 

> * On the "With Org Dropdown" story, if you select align -> left, the border in between the two items doubles up. I don't see anything one way or the other in the style guide, but if it's not too hard we should probably fix that.

This one is kind of an interesting problem since the borders are defined on the Dropdown Button, but the border for the Item that it's next to are set on a `Box` in the story. I think it should be up to the developer when setting up the BrandBar to avoid setting a border if it's next to a dropdown/another Item. Would it be alright for me to just add that logic into the story? To say, "if the dropdown as left-aligned, don't set a right border on the org name Item"?

I like that more than adding some complex logic into the BrandBar to check if there are any siblings to the left/right of a dropdown, checking if those siblings have a border, and then removing a border on the Dropdown if they do.

wilysword:
Sure, if there's a way for the devs to avoid it that's fine.

Dhalton:
Alright I think this is in a pretty good spot now. I do have a couple of things to note:

1. I don't know how I feel about including the hover/focus styles in the dropdown's CSS. Seems like it may be better to leave it to the devs during implementation to do that, so we don't have to limit them by saying any selectable item in the dropdown must be inside an `li` element. The styles are there, but I'm interested to hear what others think about whether or not they should be there.
2. The ticket says that `align` should default to `right`, but I couldn't do exactly that, since I needed to be able to differentiate between regular `BrandBar.Item`s and ones that are aligned, since the ones with an align prop are rendered in a different section in the mobile view. So if you don't pass an `align` prop, it will _behave_ as if it defaults to `right`, but it doesn't _actually_ do that. Just FYI

wilysword:
I think you may have misunderstood what I meant when I said the drop-down doesn't have hover styles. I meant the drop-down button itself (when it's closed), not any particular items _in_ the drop-down.

wilysword:
Mobile view, "With Org Dropdown" story, when you choose "align": `left` you get an extra border:
left-align

Dhalton:
> I think you may have misunderstood what I meant when I said the drop-down doesn't have hover styles. I meant the drop-down button itself (when it's closed), not any particular items _in_ the drop-down.

Doh! Yeah, I definitely misunderstood. So then, in your opinion, would you agree with me that it'd be better to leave hover/focus styles for items in the dropdown to be implemented by the dev?

wilysword:
Yeah, as I understood it the drop-down was supposed to be a bit of a blank slate for whatever they want to put in it, makes sense to include styling in what the devs should handle.

Dhalton:
Alright, I rebased onto the `cactus-web-v7` branch and resolved the conflicts. Hopefully I didn't break anything Danilo did. The `UserMenu` seems to be working, but if you notice anything funky, let me know.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -33,40 +45,107 @@ interface ProfileStyleProp {
...
+        <Flex flexDirection="column" width="100%">
+          {(!!leftChildren.length || !!rightChildren.length) && (
+            <StyledDropdownRow
+              alignItems="flex-end"
+              justifyContent={
+                leftChildren.length && !rightChildren.length
+                  ? 'flex-start'
+                  : rightChildren.length && !leftChildren.length
+                  ? 'flex-end'
+                  : 'space-between'
```
wilysword, line 131: Unless it makes the styles significantly more complicated, I'd say just make these either `space-between` or `center` always.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +290,190 @@ const StyledUserMenu = styled.div`
...
+        target instanceof Node &&
+        !buttonRef.current?.contains(target) &&
+        !portalRef.current?.contains(target)
+      ) {
+        toggle(false)
+      }
+    }
+
+    document.body.addEventListener('click', handleBodyClick)
+    document.body.addEventListener('keydown', handleBodyKeyDown)
```
wilysword, line 323: Would it work/make sense to combine these into a single `focusin` handler?

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +290,190 @@ const StyledUserMenu = styled.div`
...
+
+    return () => {
+      document.body.removeEventListener('click', handleBodyClick)
+      document.body.removeEventListener('keydown', handleBodyKeyDown)
+    }
+  })
```
wilysword, line 329: Pretty sure if you omit the dependency array entirely it runs the effect on every render, which is probably not what you want here.
```suggestion
  }, [])
```

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +290,190 @@ const StyledUserMenu = styled.div`
...
+  )
+}
+
+const BasePopup: React.FC<PopupProps> = ({ buttonRef, popupRef, isOpen, ...props }) => {
+  const isTiny = SIZES.tiny === useScreenSize()
+  const positionPortal = React.useCallback<
+    (popover: HTMLElement, target: HTMLElement | null) => void
+  >(
+    (popover, target) => {
+      if (!target || !popover || !buttonRef.current) {
```
wilysword, line 393: `target` and `buttonRef` should refer to the same element here, I'd think.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +290,190 @@ const StyledUserMenu = styled.div`
...
+    color: ${(p) => p.theme.colors.callToAction};
+  }
+
+  & svg {
+    font-size: 12px;
+  }
+`
+
+const DropdownPopup = styled(BasePopup)`
+  display: ${(p) => (p.isOpen ? 'block' : 'none')};
```
wilysword, line 466: I'd suggest using the `aria-hidden` attribute instead of passing an `isOpen` prop solely for this style:
```suggestion
  &[aria-hidden='true'] {
    display: none;
  }
```

wilysword (commented):
Didn't notice any conflicts with the user menu stuff.

Used devtools to change "Userforce One" to something much shorter, and then when I opened the org drop-down it extended off-screen.

modules/cactus-web/src/BrandBar/BrandBar.tsx
Dhalton: Whoops, yeah that definitely wasn't intentional.

modules/cactus-web/src/BrandBar/BrandBar.tsx
Dhalton: Tried it, but it doesn't really work since you can click somewhere on the screen without focusing on anything, which doesn't trigger the event. I can, however, refactor it a bit to pull out the two functions' bodies into a single function and call it in both, so there isn't duplicated code there.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -308,15 +299,12 @@ const Dropdown: React.FC<DropdownProps> = ({ label, children }) => {
       }
     }
 
+    const handleBodyClick = (event: MouseEvent): void => {
```
wilysword, line 302: It doesn't matter much, I suppose, but can you not pass `closeOnOutsideEvent` directly to `addEventListener`?

wilysword (commented):
Drop down positioning is still off-screen when the username is very short.

Dhalton:
> Drop down positioning is still off-screen when the username is very short.

Any idea how I can prevent this? The problem isn't so much the popup as it is the content inside it. The popup stops at the right side of the screen but the content sometimes overflows. I'm not sure how to stop it from happening, especially if the content has a hard-coded width value. I removed the hard-coded width and that looked like it fixed it in Chrome, but apparently it's still an issue in whatever browser you're using.

wilysword (approved):
Nevermind, whatever you did fixed it, I just didn't notice somehow.

modules/cactus-web/src/helpers/asProps.ts
```diff
@@ -12,3 +12,16 @@ type Props<C> = C extends keyof JSX.IntrinsicElements
...
 export type AsProps<C> = { as?: C } & Props<C>
+
+type PolyProps<P, T extends React.ElementType> = T extends keyof JSX.IntrinsicElements
+  ? P & Omit<JSX.IntrinsicElements[T], 'as' | 'ref' | keyof P>
+  : T extends React.ComponentType<infer U>
+  ? P & Omit<U, 'as' | 'ref' | keyof P>
+  : never
+
+export interface PolyFC<P, D extends React.ElementType> {
+  <T extends React.ElementType = D>(p: { as?: T } & PolyProps<P, T>): React.ReactElement | null
```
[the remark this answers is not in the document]
wilysword: Which part are you asking about? `PolyFC<P, D>` is short for `PolymorphicFunctionalComponent<Props, DefaultComponentType>`, and does basically the same thing as the `AsProps` type, allowing you to pass an `as={Component}` prop and have Typescript recognize and validate the correct props for `Component`. For example, given the following component:
```
type WhenProps = { when: string; children: never }
const Poly: PolyFC<WhenProps, 'div'> = ({ when, as: Component = 'div', ...props }) => {
  return <Component {...props}>{when}</Component>
}

// This would render a div, and take all the normal div props:
<Poly when="now" id="i-am-a-div" />
// But this would render an Alert, and take all the normal Alert props:
<Poly as={Alert} when="soon" status="warning" />
```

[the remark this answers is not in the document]
daniloPenaR: Gotcha

daniloPenaR (approved):

mikesoltow (commented):
Here's a partial review, as I need to log off and start the kids' dinner/bedtime slog.

I'm still looking at the code, but no issues there so far.  I've done a _little_ UAT, but nothing extensive yet.  One issue I did notice, however, can be reproduced as follows:
1. Add several items to the end of the org dropdown menu in Storybook
2. View the story in an iPhone 5 viewport.
3. Note that you can't scroll to the items off-screen.

I don't think this issue is unique to this PR, as I recreated it with the Brand Bar user menu as well.  I think you or I would have to dig a little deeper to see if the issue resides inside one of the hooks or in their application.  If it's in the hooks, we can make another ticket.  It not, we may want to fix it here before applying it to other uses.  I'd be happy to take a deeper look in the morning.

modules/cactus-web/src/BrandBar/BrandBar.story.tsx
```diff
@@ -84,3 +142,173 @@ BasicUsage.parameters = {
...
+
+const BrandBarWithOrgDropdown = () => {
+  /*
+   * Had to pull this into a separate component so we could use the
+   * useScreenSize() hook.
+   */
+  const [currentOrg, setCurrentOrg] = React.useState<Org>(ORGS.favorites[0])
+  const [searchValue, setSearchValue] = React.useState<string>('')
+  const [searchedOrgs, setSearchedOrgs] = React.useState<Org[]>([])
+  const isTiny = SIZES.tiny === useScreenSize()
```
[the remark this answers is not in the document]
Dhalton: The `WithOrgDropdown` story renders a `Layout`, which renders `ScreenSizeProvider` internally, but I need to call the hook from a child of `ScreenSizeProvider`, so I can't call it directly in that story. I had to create a new component to render inside the provider to call the hook.

modules/cactus-web/src/helpers/asProps.ts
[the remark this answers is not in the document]
Dhalton: Honestly, I really don't know. It's a different way of doing the `as` props, I know that much, but as far as how it works goes, I couldn't say. @wilysword pretty much just gave me this interface when I came to him about some issues I was having with as props. I should've mentioned that. Maybe he can shed some light on it?

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -33,21 +45,42 @@ interface ProfileStyleProp {
...
+export const BrandBarItem: PolyFC<ItemProps, 'div'> = (props) => {
+  const { mobileIcon, as: Component, ...rest } = props
   const isTiny = SIZES.tiny === useScreenSize()
   if (isTiny && mobileIcon) {
-    return <ActionBar.Panel aria-label="" {...props} icon={mobileIcon} />
+    return <ActionBar.Panel aria-label="" {...rest} icon={mobileIcon} />
+  } else if (Component) {
+    return <Component {...(rest as any)} />
   }
   return <>{props.children}</>
```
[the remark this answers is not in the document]
Dhalton: Hmmm, that'd be tricky to do since rendering an actual element there like `span` or `div` really messes up the styles/layout of the BrandBar. 🤔  It'd be a larger refactor than you might think

Dhalton:
Thanks for mentioning the height/scrolling issue, @mikesoltow . I'll take a look at it in the morning since I'm about to log off for the day.

modules/cactus-web/src/BrandBar/BrandBar.tsx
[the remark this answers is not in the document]
mikesoltow: Ah I see.  🤔 we'll probably want to address it at _some_ point, but it's definitely not urgent.

mikesoltow (commented):
Here's a partial review after looking into the scrolling issue a bit.  I'll hop back on to finish up shortly.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +281,182 @@ const StyledUserMenu = styled.div`
...
+      document.body.removeEventListener('click', closeOnOutsideEvent)
+      document.body.removeEventListener('keydown', closeOnOutsideEvent)
+    }
+  }, [])
+
+  const handlePopupKeyDown = (event: React.KeyboardEvent) => {
+    const key = event.key
+    if (key === 'Enter' || key === 'Space') {
+      keyDownAsClick(event)
+    }
```
mikesoltow, line 315: The SPACE key isn't working for me for some reason...unless I'm doing it wrong (which is likely!)

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +281,182 @@ const StyledUserMenu = styled.div`
...
+    switch (key) {
+      case 'ArrowDown':
+        setFocus(1, { shift: true })
+        break
+      case 'ArrowUp':
+        setFocus(-1, { shift: true })
+        break
```
mikesoltow, line 322: One thing potentially missing is having the down arrow open up the dropdown when the button is focused.  The user menu currently offers that functionality, as do the MenuBar and Select components.

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +281,182 @@ const StyledUserMenu = styled.div`
...
+      case 'Space':
+        toggle(false)
+        buttonRef.current?.focus()
+        break
+    }
+  }
+
+  const handleTriggerClick = () => {
+    toggle()
+  }
```
mikesoltow, line 340: I also don't think this is necessary, as this is included in `buttonProps`.  The logic is slightly different in that event, but I think the end result is the same.

mikesoltow (commented):
Just finished looking over the code and doing UAT on Chrome.  I'll do some cross-browser testing this afternoon, but here are my comments so far.

modules/cactus-web/src/BrandBar/BrandBar.tsx
Dhalton: 🤦  The key isn't `Space` it's ` `. Nice catch

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -205,6 +281,182 @@ const StyledUserMenu = styled.div`
   outline: none;
 `
 
+const Dropdown: React.FC<DropdownProps> = ({ label, children }) => {
+  const buttonRef = React.useRef<HTMLButtonElement>(null)
+  const portalRef = React.useRef<HTMLDivElement>(null)
+  const isTiny = SIZES.tiny === useScreenSize()
+  const { expanded, toggle, buttonProps, popupProps, setFocus } = usePopup('dialog', {})
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
Dhalton: Yeah looks like you're right. I did the same and it looks like it's working on my end too.

modules/cactus-web/src/helpers/asProps.ts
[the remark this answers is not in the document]
wilysword: Yeah, although even I'm not sure _why_ it behaves a little differently than `AsProps`, since they're almost exactly the same in theory. Maybe because `PolyProps` omits the keys the dynamic component has in common with the static props?

mikesoltow (commented):
UAT all checked out for me.  Just the close-on-click issue that we talked about on Slack, and then we should be good.

Thanks for the changes!

modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -311,12 +316,21 @@ const Dropdown: React.FC<DropdownProps> = ({ label, children }) => {
       case 'Escape':
       case 'Enter':
       case ' ':
-        toggle(false)
-        buttonRef.current?.focus()
+        toggle(false, buttonRef.current)
```
mikesoltow, line 319: Should we maybe remove `' '` and `'Enter'`?  Since you call `keyDownAsClick`, presumably `handlePopupClick` will be called, which will already handle this.  Plus, including the `toggle` here means that we ignore their wishes if they pass a funky selector into `listItemSelector`.

mikesoltow (commented):
Look good...just one more comment 😆

mikesoltow (approved):
🎉 🥳 

Thanks, Dhalton!

mikesoltow:
@wilysword or @daniloPenaR do either of you want to take one last look?  I don't want to hold this up, but I want to make sure of the `listItemSelector` change since that came in after you approved.  Thanks!

daniloPenaR:
I tested it all again on `Chrome`, `Firefox` and `IE11`, and it seems to work pretty well.  Just noticed that on IE11 the popup doesn't get closed when I press 'Esc' key. I don't know if this could be a BrowserStack issue or an IE11's one, is that relevant though?

wilysword (approved):

mikesoltow:
> Just noticed that on IE11 the popup doesn't get closed when I press 'Esc' key. I don't know if this could be a BrowserStack issue or an IE11's one, is that relevant though?

@daniloPenaR I can't replicate this when I run it on IE11, so I'm not sure what you're seeing.

daniloPenaR:
> > Just noticed that on IE11 the popup doesn't get closed when I press 'Esc' key. I don't know if this could be a BrowserStack issue or an IE11's one, is that relevant though?
> 
> @daniloPenaR I can't replicate this when I run it on IE11, so I'm not sure what you're seeing.

@mikesoltow Weird, going to check it again...

Dhalton:
@daniloPenaR Any update? I wasn't able to reproduce the issue, but I'm not sure if I should put the ready to merge label on this yet or not

daniloPenaR:
@Dhalton Nevermind, go for it.

PR #616, opened 2021-07-01
CACTUS-630: fix props tables for various components

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-630

Also fixed a few other problems with bad links, grammar, outdated info, etc., though I'm sure I didn't get everything.

The doc generator seems so random... A few components have their filename from `dist` instead of `src`; some components have an automatically generated `displayName`, some don't, and some don't even seem to work when they have an _explicit_ `displayName`. Grr.


Dhalton (approved):

mikesoltow (approved):

daniloPenaR (approved):
(I didn't noticed that this PR has 2 approvals, nevertheless I'm going to approve).

PR #620, opened 2021-07-02
CACTUS-568: reduce min width of Select component

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-568

I picked the new min width using the shortest reasonable placeholder I could think of (`...`). Technically you can set the placeholder to the empty string, but I think it's still better to have at least a little space to work with.

I also fixed the placeholder styles so they have the same overflow behavior as selected value text, and set a default background color (noticed it didn't have one in the rules form, where the prototype called for putting a Select over a lightContrast background).

BREAKING CHANGE: reduced min-width could break layouts in theory, though very unlikely in practice


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -208,10 +207,9 @@ const ValueSpan = styled.span`
   text-overflow: ellipsis;
 `
 
-const Placeholder = styled.span`
+const Placeholder = styled(ValueSpan)`
```
mikesoltow, line 210: This actually caused the placeholder to shift upwards a bit.  I don't think that's an issue on its own, as it appears to be the same alignment as the values themselves.  I just wonder if we should fix the `ValueSpan` itself so it's actually centered. `vertical-align: middle` seems to work, but it looks like the icon button needs some alignment too (maybe an extra px of `top`).

We can do that separately if you want...but it seems like an easy fix.

mikesoltow (commented):
Looks pretty good.  I have one question about the placeholder alignment.

Also, it looks like word wrapping on the dropdown items was already broken in IE.  Since we're potentially making the dropdowns even skinnier, would you mind fixing that really quickly?  I think all you have to do is add `word-wrap: break-word;` to `StyledOption` (I didn't get to test, though, cause my BrowserStack IE11 stopped working).

wilysword:
I figured the alignment change would be too subtle for the diff engine to catch, so I just regenerated all the storyshots.

mikesoltow (approved):
Changes look good...thanks!

Dhalton (approved):
LGTM

PR #625, opened 2021-07-13
Remove flakey UI Config tests

mikesoltow:
These tests are constantly killing our builds.  We'll need to address these tests down the road, but for now I think we'll be better served by not running them.


wilysword (approved):
Now that's a well-named file.

stevemom-repay (approved):

PR #628, opened 2021-07-16
CACTUS-665 :: Require Media Queries

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-665

### Testing
1. Pull down this branch and run `yarn cleanup && yarn build && yarn web start`
2. Inspect some components that use breakpoints/media queries at different screen sizes and make sure they're behaving as they should at each respective screen size. Some good components to test with are:
  - Layout
  - BrandBar
  - Footer
  - Table
  - DataGrid
  - Basically any component that adjusts its display/layout depending on screen size.

### DIFF
--- a/modules/cactus-web/src/Select/Select.tsx
+++ b/modules/cactus-web/src/Select/Select.tsx
@@ -877,41 +877,39 @@ class List extends React.Component<ListProps, ListState> {
           {options.length === 0 && !this.props.canCreateOption ? (
             <NoMatch>{matchNotFoundText}</NoMatch>
           ) : (
-            options.map(
-              (opt): React.ReactElement => {
-                const optId = opt.id
-                const isSelected = getSelected(opt)
-                let ariaSelected: boolean | 'true' | 'false' | undefined = isSelected || undefined
-                const isCreateNewOption = comboBox && optId === `create-${this.state.searchValue}`
-                // multiselectable should have aria-selected on all options
-                if (multiple) {
-                  ariaSelected = isSelected ? 'true' : 'false'
-                }
-                return (
-                  <InternalOption
-                    key={optId}
-                    option={opt}
-                    className={activeDescendant === optId ? 'highlighted-option' : undefined}
-                    data-role={isCreateNewOption ? 'create' : 'option'}
-                    aria-selected={ariaSelected}
-                    onMouseEnter={this.handleOptionMouseEnter}
-                  >
-                    {isCreateNewOption ? (
-                      <ActionsAdd mr={2} mb={2} />
-                    ) : multiple ? (
-                      <CheckBox
-                        id={`multiselect-option-check-${optId}`}
-                        aria-hidden="true"
-                        checked={isSelected}
-                        readOnly
-                        mr={2}
-                      />
-                    ) : null}
-                    {opt.label}
-                  </InternalOption>
-                )
+            options.map((opt): React.ReactElement => {
+              const optId = opt.id
+              const isSelected = getSelected(opt)
+              let ariaSelected: boolean | 'true' | 'false' | undefined = isSelected || undefined
+              const isCreateNewOption = comboBox && optId === `create-${this.state.searchValue}`
+              // multiselectable should have aria-selected on all options
+              if (multiple) {
+                ariaSelected = isSelected ? 'true' : 'false'
               }
-            )
+              return (
+                <InternalOption
+                  key={optId}
+                  option={opt}
+                  className={activeDescendant === optId ? 'highlighted-option' : undefined}
+                  data-role={isCreateNewOption ? 'create' : 'option'}
+                  aria-selected={ariaSelected}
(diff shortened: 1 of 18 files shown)


mikesoltow:
Hey @Dhalton : just a heads up that I'm going to re-target this to the `cactus-web-v8` branch.  While this doesn't break Cactus Web by itself, it does break the theme.  And since I made a breaking change to Cactus Web in my Flatten Action Colors PR that makes the next major version of the theme a peer dependency, we might as well bundle this up with that.

modules/cactus-web/src/StyleProvider/StyleProvider.tsx
```diff
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import * as styledComponents from 'styled-components'
 
-import { breakpointOrder, breakpoints } from '../helpers/constants'
```
mikesoltow, line 6: I think you can remove these fields from `constants.ts` as well, right?

mikesoltow (commented):
Looks pretty good.  Just one question.

modules/cactus-web/src/StyleProvider/StyleProvider.tsx
Dhalton: Yeah you're right, although now that I'm looking at it, there's an `isResponsiveTouchDevice` boolean in there whose logic depends on `breakpoints.small`. So we could either refactor that into a function where we pass the breakpoints from the theme to it and refactor the places it's used, or we could figure out the largest width for a touch device that you'd see in the real world and use that value. I like the first option better but it may be a little tricky to get a reference to the theme everywhere it's used 🤔

wilysword (commented):
Looks good to me. Do we want to refactor the existing usages (where we have `p.theme.mediaQueries?.small`), save that for a separate ticket, or just do it little by little as we come across it in other tickets?

Dhalton:
> Looks good to me. Do we want to refactor the existing usages (where we have `p.theme.mediaQueries?.small`), save that for a separate ticket, or just do it little by little as we come across it in other tickets?

I guess I just assumed we'd do it little by little since it doesn't hurt to have it written that way. I'm going through and making some changes right now, though, so I might as well just update all of those as well.

wilysword (approved):

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -84,6 +84,7 @@ export interface SelectProps
   onChange?: React.ChangeEventHandler<Target>
   onBlur?: React.FocusEventHandler<Target>
   onFocus?: React.FocusEventHandler<Target>
+  theme: CactusTheme
```
mikesoltow, line 87: Since we probably don't want to expose this prop (and we don't want to generate docs for it), I'd suggest removing this line and defining the props for `SelectBase` as `SelectProps & { theme: CactusTheme }`.  You might even want to store that as a type alias so you can use it in the `displayStatus` function as well.

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -1490,7 +1496,7 @@ class SelectBase extends React.Component<SelectProps, SelectState> {
               id={`${id}-input`}
               name={name}
               autoComplete="off"
-              autoFocus={isResponsiveTouchDevice ? true : false}
+              autoFocus={!this.isTouchDevice ? true : false}
```
mikesoltow, line 1499: Is this boolean backwards?

mikesoltow (commented):
I like the changes :+1: Just a couple more tiny things.

modules/cactus-web/src/Select/Select.tsx
Dhalton: 🤦‍♂️  Yep nice catch

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -1566,6 +1576,12 @@ export const Select = styled(SelectBase)`
   }
 `
 
+type SelectType = typeof StyledSelect & { Option: typeof SelectOption }
+
+const Select = StyledSelect as SelectType
+
+export { Select }
+
```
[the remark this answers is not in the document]
Dhalton: 🤔  Hmm, I'll have to think about this a bit more. The only problem with what you're suggesting is that then you get type errors when you try to render `Select.Option`, since `index.ts` references the named export.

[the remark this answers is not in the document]
mikesoltow: You can always change that line to `export { default as Select } from './Select/Select'`.  I know we do that in a couple of other places.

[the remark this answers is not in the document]
Dhalton: Oh okay, I didn't know that was an option. Yeah I'll just do that then

mikesoltow (approved):
Thanks!

PR #632, opened 2021-07-23
CACTUS-675 :: Button Font

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-675

So the reason these were left out was because, for some reason, they cause the integration tests for `mock-ebpp` to fail. I ended up removing the `global` flag from `StyleProvider` in the example app to get them to pass. It's more of a workaround than a solution, and the app looks pretty gross without them, but I'm not sure what else to do. I tried overriding the button styles in `mock-ebpp` with a global style, but that didn't seem to work. This app is only used for integration tests so I think it should be okay to do that...right? 🤷‍♂️ 

To test, just bring up the storybooks and inspect a `button` element to make sure it has the correct font and text styles.


daniloPenaR (commented):
yeah, the app looks pretty weird without the `global` prop. I was wondering if that could happen with other tests in another app

Dhalton:
> I was wondering if that could happen with other tests in another app

The only tests that I found were failing were the ones using the `mock-ebpp` app. The tests were hanging when trying to select an action in a `SplitButton` and that's not used anywhere in the other two example apps.

Dhalton:
Thanks for figuring this one out Mike!

mikesoltow (approved):
So maybe my approval is kind of cheating, since I contributed to this.  But I checked out the main changes (the font), and all of that looked good.  So I'm going for it.

PR #634, opened 2021-07-26
CACTUS-657 Footer Redesign

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-657

First commit is kind of unrelated (read commit message for more info), but I wanted it so that people could restyle the footer easily, if they wanted, by doing something like this:
```
const NewFooter = styled(Footer)`
  .footer-logo {
    background-color: chartreuse;
  }
`
```
That's also why I switched to the `.attrs({ as })` styling strategy. (It's a pain that I keep learning new things and changing my mind about what the "best" way is...)

For the most part I strictly followed the style guide here. The only sort-of exception is the placement of the text. The style guide specifies a dynamic gap between the logo and text, but doesn't indicate how that should be accomplished. After playing around a bit, I thought the way that looks best is to center the text, with a max-width of 1200px so there'll still be a larger gap on extraLarge screens.


mikesoltow:
@Hplata I'm tagging you to make sure you're okay with the following.  I took a look, and it's fine with me, but I trust your opinion more than mine!
>For the most part I strictly followed the style guide here. The only sort-of exception is the placement of the text. The style guide specifies a dynamic gap between the logo and text, but doesn't indicate how that should be accomplished. After playing around a bit, I thought the way that looks best is to center the text, with a max-width of 1200px so there'll still be a larger gap on extraLarge screens.

Here's what it looks like on a large screen:
Screen Shot 2021-07-28 at 1 11 21 PM

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -1,260 +1,146 @@
...
-    [key, setLinks]
-  )
-  return null
-}
-
-const StyledFooter = styled.footer.attrs({ role: 'contentinfo' as string })<{ isGrid: boolean }>`
-  ${(p) => textStyle(p.theme, 'small')};
-  word-wrap: break-word;
-  overflow-wrap: anywhere;
+export const Footer = styled(FooterBase).attrs({ as: FooterBase, role: 'contentinfo' as string })`
```
mikesoltow, line 104: You'll have to remind me what this actually does again.  I'm sure it's legit, but it just reads so strangely as "styled A as A."

mikesoltow (commented):
I'm essentially good with this.  I have a suggestion on the tests, but it's not a blocker.  The only other thing that would keep me from approving now is some questions I have about inconsistencies in the Style Guide with regards to link color.  I've pinged Hector and Juan about them, so we'll see what they say.

modules/cactus-web/src/Footer/Footer.tsx
wilysword: When creating an element, `styled-components` does something like this (simplified):
```
const component = attrs.as || props.as || target // where target comes from `styled(target)`
return React.createElement(component, propsPlusAttrs)
```
In this, case, the `as: FooterBase` is the important point, and I could have done `styled.footer` or even `styled(() => null)` and it would have worked the same because the `attrs` overrides everything else. The only reason I did `styled(FooterBase)` is because that tells Typescript what props I'm supposed to have available in the CSS definitions, and it's slightly shorter than
```
styled.footer.attrs(...)<FooterProps>`
  ...etc.
`  as React.FC<FooterProps>
```
It's a pretty small difference, though, so if you (or anyone else) think it would look better or be more readable like that, I don't mind.

mikesoltow: Got it.  And remind me again why it's important to have `FooterBase` in the `as` prop instead of the target?  Does that have something to do with the "folding" you mentioned in the ToggleCards PR?

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -1,260 +1,146 @@
...
-  for (let i = 0, linkIndex = 0; i < maxCols && linkIndex < links.length; i++) {
-    const start = linkIndex
-    linkIndex += numPerCol
-    divided[i] = links.slice(start, linkIndex)
-    if (i < extra) {
-      divided[i].push(links[linkIndex++])
+const lightLink = css`
+  a {
+    :link {
+      color: ${(p) => p.theme.colors.base};
```
mikesoltow, line 75: The updated color is now `callToAction`

modules/cactus-web/src/Footer/Footer.tsx
```diff
@@ -1,260 +1,146 @@
...
+    }
+    :visited {
+      color: ${(p) => p.theme.colors.darkContrast};
```
mikesoltow, line 78: The updated color is now `darkestContrast`

mikesoltow (changes requested):
Unfortunately there was some confusion over the link color in the Style Guide.  We got it sorted out, but there are two small changes to make.

modules/cactus-web/src/Footer/Footer.tsx
wilysword: It prevents people from breaking the component (or at least getting really weird results) by passing an `as` prop. Honestly it's kind of overkill, since changing the type to `FC<FooterProps>` would make Typescript complain about an `as` prop, and beyond that if someone passes an undocumented prop and breaks their app it's basically their own fault. It makes me feel better, though, to be just that little bit more proof against fools.

modules/cactus-web/src/Footer/Footer.tsx
wilysword: Just to be sure: the style guide now _says_ darkest contrast, but the color shown is still `darkContrast`. I want to be extra sure here, because it's unusual to have the `:visited` color be the same as regular non-link text. (Although it is for the `dark` variant...)

modules/cactus-web/src/Footer/Footer.tsx
mikesoltow: Fair enough!  Thanks for the explanation.

modules/cactus-web/src/Footer/Footer.tsx
mikesoltow: Okay, I think we've settled on the colors after digging through them even more 🤞 :
* `:visited` for `lightLink` should remain `darkContrast`
* `:focus` for `lightLink` _and_ the `dark` variant should be `{ color: darkContrast, backgroundColor: lightCallToAction }`

modules/cactus-web/src/Footer/Footer.test.tsx
```diff
@@ -12,24 +12,20 @@ describe('component: Footer', () => {
...
-        </Footer>
+        <Footer logo={REPAY_LOGO}>Custom Content</Footer>
       </StyleProvider>
     )
 
     expect(getByText('Custom Content')).toBeInTheDocument()
   })
 
-  test('should be able to pass links using Footer.Link', () => {
+  test('should be able to pass links using a', () => {
```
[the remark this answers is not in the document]
wilysword: I think the styles are well-covered by the storyshots, so I'll remove it.

mikesoltow (approved):
Thanks, Glen!

daniloPenaR (approved):

wilysword:
I already started to fix the conflicts on https://github.com/repaygithub/cactus/pull/635, so I'm just going to wait till that's merged to fix all the conflicts at once.

PR #638, opened 2021-08-02
CACTUS-694 :: DateInput Bug

mikesoltow:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-694

I also fixed a bug where the calendar popup wasn't closing on blur if you clicked inside of the modal.  We're using `@reach/dialog` under the hood, and they call `stopPropagation` on click, so our `document.body` click handler, which was handling this before, wasn't getting called.  The new behavior I added mimics what we do with `usePopup` (I didn't actually switch to `usePopup` because that would be a whole thing.  If you think we should do that, I can create a tech debt ticket for that...just let me know.).

The bug resulted because the previous alignment settings were inheritable fields, so EVERYTHING that was a child of a Dimmer was getting centered (see the updated "modals with alert" screenshot...previously the text in the Alert was getting centered).  I changed to using flex settings to control the Dimmer alignment.  These fields aren't inheritable, so they won't mess up any ancestors...while still allowing this children of a Dimmer to be centered.

Also, I added some new screenshot tests to cover both this bug and to ensure that we're testing the activated state of the Dimmer (which was affected by this change).  I generated the updated Dimmer screenshot before implementing the fix, so I verified that the fix didn't mess with the alignment.


modules/cactus-web/src/Dimmer/Dimmer.story.tsx
```diff
@@ -34,3 +37,9 @@ export const ActivePageDimmer = (): React.ReactElement => {
     </>
   )
 }
+
+ActivePageDimmer.parameters = {
+  beforeScreenshot: async (page: Page) => {
+    await page.click('[id="dimmer-btn"]')
```
wilysword, line 43: This is fine (equivalent) so you don't have to change it, but in the interest of spreading CSS knowledge: the ID selector is `#dimmer-btn`.

wilysword (approved):

modules/cactus-web/src/Dimmer/Dimmer.story.tsx
mikesoltow: D'oh!  I swear I knew that...I just brain-farted

Dhalton (approved):
LGTM!

PR #640, opened 2021-08-02
CACTUS-664 Theme Style Helpers

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-664

I added all the helpers from channels + cactus-web, although the API is more like what I did in channels. I even added a couple of new ones, so we pretty much cover every theme attribute.

If anyone can think of a better name for the media query helpers than `mediaGTE` and `mediaLT`, that'd be great: I really don't like those names, but couldn't think of anything better that wasn't several words long.

Last thing to be aware of is that I changed some things about the build/Typescript config. It worked within the repo, but I don't know much about JS packaging, so hopefully it'll still work after it's deployed. Make sure you do `yarn cleanup` before testing, though, to get rid of the old build code.

Finally, if you want to see the new helpers in action, I made a branch where I applied most of them at least once: https://github.com/repaygithub/cactus/compare/cactus-664-theme-helpers...verify-664?w=1

### DIFF
--- a/modules/cactus-theme/src/converters.ts
+++ b/modules/cactus-theme/src/converters.ts
@@ -8,7 +8,7 @@ export function rgbToHsl(red: number, green: number, blue: number): [number, num
 
   const max = Math.max(red, green, blue)
   const min = Math.min(red, green, blue)
-  let hue: number
+  let hue = 0
   let saturation: number
   let lightness = (max + min) / 2
 
@@ -45,19 +45,22 @@ export function rgbToHsl(red: number, green: number, blue: number): [number, num
 export function hexToRgb(hex: string): [number, number, number] {
   hex = hex?.trim?.()
   if (/^#?([a-f\d]{3}){1,2}$/i.test(hex)) {
-    let result
+    let result: string[] | null = null
     if (hex.length < 6) {
       result = /^#?([a-f\d]{1})([a-f\d]{1})([a-f\d]{1})$/i.exec(hex)
-      result[1] += result[1]
-      result[2] += result[2]
-      result[3] += result[3]
+      if (result) {
+        result[1] += result[1]
+        result[2] += result[2]
+        result[3] += result[3]
+      }
     } else {
       result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex)
     }
-    return [parseInt(result[1], 16), parseInt(result[2], 16), parseInt(result[3], 16)]
-  } else {
-    return [255, 255, 255]
+    if (result) {
+      return [parseInt(result[1], 16), parseInt(result[2], 16), parseInt(result[3], 16)]
+    }
   }
+  return [255, 255, 255]
 }
 
 // Calculating YIQ for easy contrast comparisons based on https://24ways.org/2010/calculating-color-contrast
--- a/modules/cactus-theme/src/helpers/base.ts
+++ b/modules/cactus-theme/src/helpers/base.ts
@@ -1,0 +1,40 @@
+import { CactusTheme } from '../theme'
+
+export type ThemeProps = { theme: CactusTheme }
+
+type ThemeFunc<A extends any[], R> = (props: ThemeProps, ...args: A) => R
+
+export type StyleFunc<R> = (props: ThemeProps) => R
+
+export interface HelperFunc<A extends any[], R> {
+  (...args: A): StyleFunc<R>
+  (pt: CactusTheme | ThemeProps, ...args: A): R
+}
+
+export const wrap = <A extends any[], R>(func: ThemeFunc<A, R>): HelperFunc<A, R> => {
+  const helper: any = (first: any, ...rest: any[]) => {
+    if (typeof first === 'object') {
+      if ('theme' in first) {
+        return (func as any)(first, ...rest)
+      } else if ('space' in first && 'mediaQueries' in first) {
+        // Picked two relatively uncommon names to represent the theme shape.
+        return (func as any)({ theme: first }, ...rest)
+      }
+    } else if (rest.length === 0 && first in helper) {
+      return helper[first]
+    }
+    return (p: ThemeProps) => (func as any)(p, first, ...rest)
+  }
+  return helper
+}
+
+export const memo = <A extends any[], R>(
+  func: ThemeFunc<A, R>,
+  ...toCache: A[0][]
+): HelperFunc<A, R> => {
+  const helper: any = wrap(func)
+  for (const arg of toCache) {
+    helper[arg] = (p: ThemeProps) => (func as any)(p, arg)
+  }
+  return helper
+}
--- a/modules/cactus-theme/src/helpers/base.ts
+++ b/modules/cactus-theme/src/helpers/base.ts
@@ -12,18 +12,20 @@ export interface HelperFunc<A extends any[], R> {
 }
 
 export const wrap = <A extends any[], R>(func: ThemeFunc<A, R>): HelperFunc<A, R> => {
-  const helper: any = (first: any, ...rest: any[]) => {
-    if (typeof first === 'object') {
-      if ('theme' in first) {
-        return (func as any)(first, ...rest)
-      } else if ('space' in first && 'mediaQueries' in first) {
+  const helper: any = (...args: A) => {
+    if (typeof args[0] === 'object') {
+      if ('theme' in args[0]) {
+        const first = args.shift() as ThemeProps
+        return func(first, ...args)
+      } else if ('space' in args[0] && 'mediaQueries' in args[0]) {
         // Picked two relatively uncommon names to represent the theme shape.
-        return (func as any)({ theme: first }, ...rest)
+        const theme = args.shift() as CactusTheme
+        return func({ theme }, ...args)
       }
-    } else if (rest.length === 0 && first in helper) {
-      return helper[first]
+    } else if (args.length === 1 && args[0] in helper) {
+      return helper[args[0]]
     }
-    return (p: ThemeProps) => (func as any)(p, first, ...rest)
+    return (p: ThemeProps) => func(p, ...args)
   }
   return helper
 }
@@ -34,7 +36,7 @@ export const memo = <A extends any[], R>(
 ): HelperFunc<A, R> => {
   const helper: any = wrap(func)
   for (const arg of toCache) {
-    helper[arg] = (p: ThemeProps) => (func as any)(p, arg)
+    helper[arg] = helper(arg)
   }
   return helper
 }
--- a/modules/cactus-theme/src/helpers/borders.ts
+++ b/modules/cactus-theme/src/helpers/borders.ts
@@ -1,0 +1,66 @@
+import { BorderSize, CactusColor, Shape } from '../theme'
+import { memo, ThemeProps, wrap } from './base'
+
+type BorderSizeOpts = { [K in BorderSize]: unknown }
+const DEFAULT_BORDER: BorderSizeOpts = { thin: '1px', thick: '2px' }
+const radiusFactors: { [K in Shape]: number | null } = { square: 0, intermediate: null, round: 1 }
+
+/**
+ * Returns a selected value depending on the `theme.border` setting.
+ * By default returns the border thickness, but can select any kind of value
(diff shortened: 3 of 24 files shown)


modules/cactus-theme/src/helpers/breakpoints.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/** Returns a min-width media query of the type used by `styled-system`. */
+const _mediaGTE = (p: ThemeProps, breakpoint: Breakpoint) => p.theme.mediaQueries[breakpoint]
+export const mediaGTE = memo(_mediaGTE, 'small', 'medium')
+
+/** Returns a media query that's the inverse of the equivalent `mediaGTE(X)` query. */
+const _mediaLT = (p: ThemeProps, breakpoint: Breakpoint) =>
+  // Assumes query starts with `@media screen and (min-width:`
+  `@media screen and (not ${p.theme.mediaQueries[breakpoint].slice(18)})`
+export const mediaLT = memo(_mediaLT, 'small', 'medium')
```
mikesoltow, line 20: I'm testing this in mock-ebpp and can't get it to work.  Is there a reason you didn't use max-width?  I tried that, and it seemed to work.

modules/cactus-theme/src/helpers/shadows.ts
```diff
@@ -0,0 +1,49 @@
...
+  '0px 12px 24px',
+  '0px 30px 42px',
+  '0px 45px 48px',
+]
+const isCssValue = (x: unknown): x is string => typeof x === 'string' && !x.includes(':')
+
+/** Returns a `box-shadow` property with the given standard shadow type (from the style guide). */
+const _boxShadow = (p: ThemeProps, shadowType: number, colorName = 'lightCallToAction') => {
+  const color_ = p.theme.colors[colorName as CactusColor] || colorName
+  return `box-shadow: ${shadowTypes[shadowType]} ${color_}`
```
mikesoltow, line 20: We should probably tack on a `;` here (and in the `shadow` function)

modules/cactus-theme/src/helpers/borders.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Returns a border CSS value with the given color and thickness according to `theme.border`.
+ * The color can be any color name or string, though `theme.colors` is checked first.
+ * The thickness can be customized with an options object like `borderSize` accepts.
+ */
+const _border = (p: ThemeProps, colorName: string, opts?: Partial<BorderSizeOpts>) => {
+  const thickness = _borderSize(p, { ...DEFAULT_BORDER, ...opts })
+  const color_ = p.theme.colors[colorName as CactusColor] || colorName
+  return `${thickness} solid ${color_}`
```
mikesoltow, line 24: Should we prefix this string with `border:`?  That would be a bit more consistent with some of the other helpers.

modules/cactus-theme/src/helpers/borders.ts
```diff
@@ -0,0 +1,66 @@
...
+    vOffset = `-${thickness}`
+  } else if (direction === 'left') {
+    hOffset = thickness
+  } else if (direction === 'right') {
+    hOffset = `-${thickness}`
+  } else {
+    spread = thickness
+  }
+  const color_ = p.theme.colors[colorName as CactusColor] || colorName
+  return `box-shadow: inset ${hOffset} ${vOffset} 0 ${spread} ${color_}`
```
mikesoltow, line 64: We should probably tack on a `;`

mikesoltow (commented):
This looks great!  I just have a few minor comments, but I tested a bunch of this in mock-ebpp, and they worked well.  The currying works beautifully, so thanks for implementing that.

modules/cactus-theme/src/helpers/base.ts
```diff
@@ -0,0 +1,40 @@
...
+        // Picked two relatively uncommon names to represent the theme shape.
+        return (func as any)({ theme: first }, ...rest)
+      }
+    } else if (rest.length === 0 && first in helper) {
+      return helper[first]
+    }
+    return (p: ThemeProps) => (func as any)(p, first, ...rest)
+  }
+  return helper
+}
```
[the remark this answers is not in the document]
wilysword: To me, `...(args as X)` is at least as ugly. I could try some other things, but I'm not promising anything, nor will I spend a lot of time on this.

modules/cactus-theme/src/helpers/breakpoints.ts
wilysword: It worked for me on the `verify-664` branch (Header -> With Description story). Technically `not (min-width)` is not the same as `max-width`: there's 1px of overlap with the latter, so I'd prefer not to use it if possible. Could you check the Header story? If that's not working it might be a browser difference, although I though I tested this on at least Safari and IE.

modules/cactus-theme/src/helpers/shadows.ts
wilysword: I admittedly don't have a _good_ reason for this, but I do have a reason for it: I think it looks better like this:
```
  ${boxShadow(5, 'gray')};
```
than it does without the semicolon. It also is more consistent with the majority of the other helpers: I think the only ones that include their own semicolons are `fontSize` and `textStyle`, and that's because those can return blocks instead of individual properties.

Having said that, if you really want a semicolon on all helpers that return full properties, I won't fight you on it.

modules/cactus-theme/src/helpers/borders.ts
wilysword: No, because this isn't just applicable to `border`. It can also be used with `outline`, and the individual border sides e.g. `border-top`.

modules/cactus-theme/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "src"
+  },
+  "include": ["src/**/*.ts"]
```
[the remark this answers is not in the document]
wilysword: `cactus-web` includes all the test/story files in the component directories alongside the regular code. `cactus-theme` only has source code in the `src` directory, so you don't need to exclude anything.

modules/cactus-theme/src/helpers/base.ts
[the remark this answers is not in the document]
mikesoltow: Fair enough.  This isn't a blocker.  It's more of an issue of me feeling like `any` is cheating than an ugliness thing, FWIW (as you are correct that `...(args as X)` is ugly.

modules/cactus-theme/src/helpers/shadows.ts
mikesoltow: I brought it up because, when I was testing in mock-ebpp, I kept doing things like:

```
const X = styled(Y)`
  ${boxShadow(5, 'gray')}
`
```

and getting errors. For some reason it seems more intuitive for the semicolon to be included...but that could just be me.  Feel free to wait for the next reviewer to weigh in on this.

modules/cactus-theme/src/helpers/breakpoints.ts
mikesoltow: Here's what I did if you want to try to reproduce.  I added the following to `mock-ebpp/containers/accounts.tsx`.  I then changed `DataGrid` to `StyledDataGrid`.

```typescript
const StyledDataGrid = styled(DataGrid)`
  background-color: ${color('success')};
  ${mediaLT('extraLarge')} {
    background-color: ${color('error')};
  }
`
```

wilysword: So apparently none of the browsers except Firefox understand the concept of of operator precedence, and will only work if the `not` is at the beginning of the query. Never mind that with something like `@media not screen and (min-width: X)` the `not` _should_ apply to either just the `screen` part, or the whole thing, and therefore should only match things that aren't screens: somehow not only do browsers recognize that the `not` skips over the `screen`, but most browsers _require_ it to be written like that. \</rant>

modules/cactus-theme/src/helpers/base.ts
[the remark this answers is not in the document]
wilysword: Changed this up a bit, see if you like it better.

modules/cactus-theme/src/converters.ts
```diff
@@ -8,7 +8,7 @@ export function rgbToHsl(red: number, green: number, blue: number): [number, num
 
   const max = Math.max(red, green, blue)
   const min = Math.min(red, green, blue)
-  let hue: number
+  let hue = 0
```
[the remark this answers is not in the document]
wilysword: When I modified the Typescript config, it started complaining that `hue` was being referenced before assignment. I assume Typescript wasn't actually covering this file before: all the changes in this file are Typescript fixes.

modules/cactus-theme/src/helpers/base.ts
```diff
@@ -0,0 +1,42 @@
...
+export const memo = <A extends any[], R>(
+  func: ThemeFunc<A, R>,
+  ...toCache: A[0][]
+): HelperFunc<A, R> => {
+  const helper: any = wrap(func)
+  for (const arg of toCache) {
+    helper[arg] = helper(arg)
+  }
+  return helper
+}
```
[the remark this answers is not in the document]
wilysword: `wrap` is what make the helpers' multiple call modes possible. As it says in the docs, you call the helpers in several ways:
```
helper({ theme }, ...args)
helper(theme, ...args)
helper(...args)({ theme })
```
All the helpers are actually defined as `helper({ theme}, ...args)`, so `wrap` checks the arguments passed in and chooses the correct call to make: can it call the helper directly, or does it need to return a curried function.

`memo`  is meant to reduce memory usage: there are some helpers that come up dozens of times, like `color('callToAction')` or `space(3)`. Rather than create dozens of copies of the same function, `memo` makes it so that every call with a common argument returns the same function.

modules/cactus-theme/src/helpers/base.ts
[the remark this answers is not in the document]
mikesoltow: Yup, that looks great...thanks!

wilysword:
Sorry for adding something at the 11th hour, but I just remembered a helper that I missed: `lineHeight`. The `line-height` property is mostly only used via the `textStyle` helper, but several times I've come across a situation where I wanted to make an icon or other block element have the same height as a line of text: this helper to the rescue!

modules/cactus-theme/tests/size-helpers.test.ts
```diff
@@ -1,5 +1,8 @@
-import { fontSize, iconSize, space, textStyle } from '../src/helpers/sizes'
-import { expectCurry, expectMemo } from './helpers'
+/**
+ * @jest-environment jsdom
+ */
```
mikesoltow, line 3: Should we just set `"testEnvironment": "jsdom"` inside `package.json` instead?  I know a few of the other packages have had to do that.

mikesoltow (approved):
Changes look good to me.  Thanks!

wilysword:
Sorry times 2, but I've been agonizing all morning over the best interface for the `lineHeight` helper, and as I was eating lunch I decided that there really wasn't much use for the simple value version that couldn't easily be covered in the render func version; as such I simplified the interface a bit and removed the media watch.

modules/cactus-theme/src/helpers/borders.ts
daniloPenaR: Is it a CSS property, isn't it? should it return a `;` at the end

wilysword: @daniloPenaR To avoid confusion, I'm marking this conversation as resolved. If you want to vote on the issue of including semicolons, do it on the other comment thread: https://github.com/repaygithub/cactus/pull/640#discussion_r682014764

Sorry for the inconvenience.

daniloPenaR:
My vote in the discussion of the `;` is a Yes.

mikesoltow (approved):
:+1: I think the interface update was a nice change.  LGTM

daniloPenaR (approved):
This looks pretty awesome and very useful. Thanks.

wilysword:
Due to popular demand, I added semicolons to all helpers that output full properties.

mikesoltow (approved):
I just double-checked the semi-colons, and everything works great.  You have earned the rare triple-approval from me.  Maybe even rarer than a triple play?

PR #642, opened 2021-08-05
CACTUS-674 :: BrandBar Tests

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-674

As I was testing, I noticed that the `BrandBar.UserMenu` didn't support Home/End keyboard functionality, so I added that in real quick, since it was pretty easy. For some reason, when I put that logic in a `switch` statement, it broke the arrow functionality. It appeared that the inputs were being applied twice. For instance, one `ArrowDown` would result in the focus moving down two indices. That's why I kept it in an if-else-if chain.

Anyway, I think these tests should be more robust now, but if you notice anything I missed, let me know.

### Testing
1. Open the "BrandBar > Basic Usage" story
2. Open the UserMenu and use the Home/End keys. They should move the focus to the beginning/end of the list, respectively.


modules/cactus-web/src/BrandBar/BrandBar.test.tsx
```diff
@@ -20,16 +20,30 @@ describe('component: BrandBar', () => {
               </BrandBar.UserMenuItem>
             </BrandBar.UserMenu>
           </BrandBar>
+          <div>Click me</div>
         </StyleProvider>
       )
 
-      userEvent.click(getByText('Test name'))
+      const trigger = getByText('Test name').parentElement as HTMLElement
```
mikesoltow, line 27: I've had good luck with changing these types of queries to `getByRole`.  In this case, you could do `getByRole('button, { name: 'Test name' })`.  I think that's a little less tied to the specific DOM structure since roles are tied to usability/accessibility.

Or you could try using a test ID for this element, since we're supposed to start doing that anyway (I keep forgetting 😬 ).

modules/cactus-web/src/BrandBar/BrandBar.test.tsx
```diff
@@ -20,16 +20,30 @@ describe('component: BrandBar', () => {
...
+      expect(settings).not.toBeVisible()
+
+      userEvent.click(trigger)
+      userEvent.click(settings)
       await animationRender()
-      userEvent.click(getByText('Settings'))
+
       expect(actionOne).toHaveBeenCalled()
+      expect(getByText('Logout')).not.toBeVisible()
+      expect(document.activeElement).toBe(trigger)
```
mikesoltow, line 43: Not a blocker, but I just realized that jest-dom has a `.toHaveFocus()` assertion.  It essentially does the same thing, but it uses `element.ownerDocument.activeElement` instead of `document.activeElement`.  No need to change unless you want to...just sharing.

mikesoltow (approved):
LGTM.  Just two optional comments.

mikesoltow:
@Dhalton Actually, I have one request while you're in here: can you add `beforeScreenshot` logic to the "user menu link" and "org dropdown" stories?  It really only makes sense to have screenshots for these stories if the dropdown is visible.

Dhalton:
> @Dhalton Actually, I have one request while you're in here: can you add `beforeScreenshot` logic to the "user menu link" and "org dropdown" stories? It really only makes sense to have screenshots for these stories if the dropdown is visible.

I updated the tests & opened the org dropdown before the storyshot. However, I'm not sure which story you mean by "user menu link". I don't see any stories by that name, but the basic usage story is opening the user menu before its storyshots.

mikesoltow:
> I updated the tests & opened the org dropdown before the storyshot. However, I'm not sure which story you mean by "user menu link". I don't see any stories by that name, but the basic usage story is opening the user menu before its storyshots.

D'oh!  Looks like screenshots aren't automatically removed when a story is removed, so I was looking at a screenshot without a parent.  (Perhaps we should look into that.). Thanks for fixing that one test, though!

modules/cactus-web/src/BrandBar/BrandBar.test.tsx
```diff
@@ -20,16 +20,30 @@ describe('component: BrandBar', () => {
...
+      userEvent.click(getByText('Click me'))
+
+      expect(settings).not.toBeVisible()
+
+      userEvent.click(trigger)
+      userEvent.click(settings)
       await animationRender()
-      userEvent.click(getByText('Settings'))
+
       expect(actionOne).toHaveBeenCalled()
```
daniloPenaR, line 41: what is that `animationRender()` on this part of the test? ( I know this is not part of your changes but I want to be sure I am understanding the main purpose of that helper`

daniloPenaR (commented):
This looks nice. You've added good testing cases for this component, just a minor question I had when I was looking into this...

modules/cactus-web/src/BrandBar/BrandBar.test.tsx
Dhalton: Sometimes after taking an action that changes something on screen "visually", you need to wait for the component to re-render so that whatever you're expecting to be there will be there. This one is waiting for the popover to close and for the focus to shift back to the trigger before proceeding.

daniloPenaR (approved):
Sorry about the delay. LGTM.

PR #644, opened 2021-08-06
Test/update docs

stevemom-repay:


wilysword:
@mikesoltow Should we be reviewing this PR, or close it?

mikesoltow:
@wilysword Yeah, I'd like to review it.  I've just been putting it off.  But I'll try to get to it by next week.  Feel free to take a look and review with your thoughts if you'd like.

wilysword (commented):
The information in these docs seems generally useful; there are missing examples, of course, and it's rather incongruous that the examples are all in Python considering this is a JavaScript repo. Before merging, someone would likely need to go through and fix the existing examples, and add missing ones.

I'm also not clear on why he removed the how-to on End-to-End Testing, since that seems to cover significantly different information.

luisniebla (approved):
LGTM

PR #645, opened 2021-08-10
CACTUS-664: change shadow fallback to outline

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-664

I was working on Calendar when I noticed that even when I set the `border-radius` to the same values for the outer and inner divs, they didn't line up when I turned box shadows off. Turns out `border-radius` is the outer radius, which means the inner radius changes when you add a border. As such, outline is a closer analogue of box-shadow, because it does not affect the dimensions of the container in any way.

This is just my personal opinion, though, so feel free to reject the PR if you disagree.

### DIFF
--- a/docs/Theme/Style Helpers.md
+++ b/docs/Theme/Style Helpers.md
@@ -159,9 +159,9 @@ If `theme.boxShadows` is false, the fallback is returned instead:
 - Else it's returned as-is.
 
 ```
-shadow(0, 'lightContrast') // = 'border: 1px solid [theme.colors.lightContrast]'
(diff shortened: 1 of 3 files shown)


mikesoltow (commented):
Don't outlines ignore the border radius though?  I'm not sure I totally understand your problem, so maybe that's what you're looking for.  I'm just not sure it's totally analogous to a box shadow if the outline is always rectangular (since I think box shadows respect the border radius).

wilysword:
Apparently this is (yet another) difference between Firefox and other browsers. Why the hell would the outline _not_ have the radius? You give me hope Firefox, only for other browsers to take it away.

PR #648, opened 2021-08-11
CACTUS-629 :: OrgSelect Dropdown

Dhalton:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-629

I was going to look into a virtualized list to help with performance, but I tested with 1000 orgs and performance didn't seem to be an issue. Even when you search, I didn't notice a performance hit. If you guys notice any issues with that story as you're testing, let me know and I can dig into the possibility of a virtualized list.

Also, I added width props to `Divider` since it was useful for me in this scenario. I thought that component already supported width props, because in the past, when I've rendered just `<Divider />` nothing shows up until I pass `width="100%"`. So hopefully you guys are okay with that.


Dhalton:
This is being moved to `ui-private` so I'm closing this PR.

PR #653, opened 2021-08-31
CACTUS-702 :: Select & DateInput Heights

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-702

Turns out that you actually can't easily configure the text styles for these elements, so there was no need to adjust line height/padding based on font size.

I spent some time writing some "style" tests for these changes, only to realize that we have the storyshots for that 🤦 although, if you guys want me to add them back in to the unit tests, that'd be easy enough.

### Testing
Just open up the `Select` and `DateInput` stories and inspect the elements to make sure they have the correct height (35px including borders). Make sure to check in other browsers as well.


daniloPenaR (approved):
Tested on Chrome, IE11, and Firefox, and seems to work well. 

About the "style" tests you mention, I guess it is not necessary to include 'em having snapshot tests. But I am just curious about how they look. Maybe could you share that in a different `commit` or `branch`?

Dhalton:
> About the "style" tests you mention, I guess it is not necessary to include 'em having snapshot tests. But I am just curious about how they look. Maybe could you share that in a different `commit` or `branch`?

I don't have the commit anymore, but they were pretty much just rendering the fields and checking the heights/borders/line heights with `expect(element).toHaveStyle(...)`

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -270,7 +271,7 @@ const SelectTrigger = styled.button`
   ${NavigationChevronDown} {
     position: absolute;
     right: 14px; // 14 + 2px from border
-    top: 11px;
+    top: 14px;
```
wilysword, line 274: You added 3 to the height, so to keep this centered it should be `12.5px`. (If you don't like fractional pixels, I think 12 looks best, though 13 is also fine.)

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -233,7 +233,8 @@ const SelectTrigger = styled.button`
   box-sizing: border-box;
   min-width: 60px;
   width: 100%;
-  height: 32px;
+  height: 35px;
+  ${(p) => textStyle(p.theme, 'body')}
   padding: 0 28px 0 16px;
```
wilysword, line 238: To center the text span, I believe `padding-top` should be 3px (line height (27px) + borders (2px) + y padding (6px) = 35px).

wilysword (commented):
Basically looks good, but I think the centering is a bit off for Select.

modules/cactus-web/src/Select/Select.tsx
Dhalton: I'm guessing you meant `padding-top` and `padding-bottom` should both be 3px, right? Based on your explanation. I pushed up a change, but let me know if that's not what you meant.

wilysword: Technically yeah, but as long as `height` is set to 35px, padding-bottom doesn't actually affect the placement of the text so I didn't mention it.

wilysword (approved):

PR #657, opened 2021-09-03
CACTUS-697: fix test to account for outside dates

wilysword:
Fixes the test that broke when the month changed.


mikesoltow (approved):
I'll get out ahead of the build and just approve this.

Dhalton (approved):

PR #658, opened 2021-09-07
CACTUS-699: make Layout more flexible, export useLayout

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-699

Now that Layout can support multiple fixed components or columns, the "layout props" that were previously used in positioning calculations are no longer accurate. This led to a number of other changes:

- Removed `addLayoutStyle`; it was previously only used for action bar panels, which should now be positioned using `positionPopup`.
- ~`ActionBar.Panel` now has a default `positionPopup` called `positionPanel`.~
- EDIT: I'm going to be moving this to `ActionBar.PanelPopup` instead.
- `ActionBar.Panel` is now always positioned absolutely instead of sometimes being fixed. This simplified the styles a bit, but also required adding a `z-index` to `SidebarDiv` (because `z-index` is weird like that).
- Modified `shouldForwardProp` from `StyledPopup` to forward width so I can use it in `positionPanel` to check if I need to set a default width in the styles.
- Moved the `useAction` logic for `BrandBarUserMenu`; because of how `useAction` works it was causing `positionPanel` to run before the element was actually visible. Because hooks can't be called conditionally, `useAction` now accepts a null element so that conditional logic can be used outside the hook to turn it into a noop. EDIT: The reason behind this change is gone, but I think it's a good change so I'm going to keep it.
- `BrandBar` now calls the `useLayout` hook. It uses two different layout roles, because the user menu can appear on the Sidebar while the rest of the brandbar is still visible (and role must be unique).
- `Layout.Content` (Main) also now calls the `useLayout` hook.
- When appearing on the bottom bar, `MenuBar` shifts its button to the left using absolute positioning. It previously did that by shifting the entire nav, but that screwed up the new `positionPanel`, so I moved the positioning styles to the button instead.

Snapshots: I have no idea why the BrandBar storyshot was getting skewed: the styles should have been almost identical. Adding an empty `Layout.Content` fixed the issue. The other change was for the Layout: Short Content story, changing the ActionBar knob default to true. (It was changed to false as a regression test for a bug that doesn't apply to the new Layout styles.)


Dhalton (commented):
Looks good for the most part. I especially like the `CustomGrid` story you added. I did notice one quirk while testing on IE, though.

If you set the height of the main content to some value that would cause you to have to scroll, the footer stays the same height but the main content grows down below it and you get these blanks spaces on the left & right. See screenshot:

Screen Shot 2021-09-09 at 12 06 48 PM

wilysword:
@Dhalton That wasn't really a problem with the grid, or even the story as such; IE just can't handle any conflicts at all between the height style and the height in `grid-template-rows`.

wilysword:
FYI the build failure is apparently a bug in Node. It's already been fixed in their code, supposedly, but I'm not sure how long it'll be till the fix is released.

Dhalton (approved):

mikesoltow (commented):
Didn't get to do nearly as much review as I would like, but what I've seen so far looks great.

One thing: I would love to tighten up test coverage a bit for the `useLayout` hook...potentially in another ticket depending on what we want to do.  The custom grid story obviously has a screenshot, but you can't really verify the behavior of the various grid items when the content height is larger than the viewport, etc.  We may actually want to write some normal tests for that hook, in that case.  But it's enough of a lift that I'd be fine following up with a tech debt ticket.

What do you think?

wilysword:
I don't see how you could test layout behavior in unit tests, since jest doesn't actually do visual layout. The most you could do is test the resulting styles; that might be okay for the fixed elements, but envisioning what a grid looks like from the styles would be tricky. I suppose it might still be useful to have tests covering the specific behavior of the algorithm, i.e. when height/width is ignored, when row/col starting values are auto-incremented, etc. Still, I think the story covers that at least reasonably well, and we could possibly add more knobs to cover things you're concerned about. The height knob already simulates overflow (sort of), besides what the first story has, so I'm not sure what else you think is missing.

mikesoltow:
Yeah, I meant more of testing out the algorithm...with the assumption that we'd make assertions on the class returned from the hook as well as the resulting styles stored in the context.  I know we can get some of that from the stories, but the automated screenshots only tell you so much, while the knobs require manual testing.  I'd just love to add an additional layer of automated testing to add some confidence that, for example, when you add item A above item B, then item B has a position of X and a height of Y...something like that.  It doesn't have to be every single style.  🤷 

As I said, that's another ticket, so I'm happy to discuss more as a team.

mikesoltow (changes requested):
So the rest of my review went very well...until I tested on Safari.

If you open up the Custom Grid story and set the height to 5000px, the content of the fixed bottom elements disappears.  I did some more testing by adding a fixed bottom component to the Basic Usage story, and it seems that Safari just makes all children of the fixed rows disappear if the content of the main area requires you to scroll.

I, unfortunately, have no idea what is causing this.  I'd be happy to dig into it some more if you'd like, but nothing was jumping out at me immediately.

Also, there was an oddity with the Custom Grid story on the default height where some white space was showing up to the right of the main content (see attached screenshot).  Not sure if that's something that just needs to be fixed by tweaking the `GridItem` props or something we want to fix in the layout code.
Screen Shot 2021-09-15 at 4 26 45 PM

wilysword:
The first issue you mentioned was the only one that was actually a problem. Safari inexplicably can't handle nested fixed items, so I had to change the Layout wrapper back to absolute. (That's what it was before so it shouldn't break anything.)

The other issue was weird: I couldn't find anything about it while Googling, even after I found what the problem was. For some reason beyond the comprehension of we mere mortal programmers, applying the `writing-mode: vertical-rl` style shifted the entire element 8px to the right. Since that's clearly a bug in Safari and not a problem with our code I thought about just leaving it, but it was easy enough to fix so I went ahead and did so.

mikesoltow (commented):
Thanks for fixing that.  Safari looks good now.  But I did notice one more issues that can be reproduced as follows.

1. Add the following component to `Layout.story.tsx`.

```
const StoryFixedBottom = () => {
  const layoutClass = useLayout('test', { fixed: 'bottom', size: 56 })
  return (
    <div className={layoutClass}>
      <Button>CLICK ME</Button>
    </div>
  )
}
```

2. Render that component _below_ `StoryFooter` in the Basic Usage story.
3. Open up the story on a desktop view.  Everything should look good.
4. Shrink the browser window to a mobile size.  Expand it back to a desktop size.
5. The MenuBar will now be rendered below the footer.

wilysword:
The problem was that JavaScript's sort algorithm assumes a homogenous array (which is a good assumption, normally), so it doesn't have to compare every item to every other item: if the order went grid, fixed, grid, then both grid items might be considered "equal to" the fixed item and even if the first grid item is actually "greater than" the second, they'll be left in their original position.

My solution is to just insert new items into the correct sorted order, using a stable sort algorithm that only compares items of the same type; thus fixed items won't affect the ordering of grid items and vice-versa.

mikesoltow (commented):
Sorry for the delay only to come back with one more thing:

1. Add that fixed bottom layout component with a button (see my previous comment) to the Layout Basic Usage story.
2. Click the Button.  Notice that the focus outline is clipped by the Sidebar.  Something is up with the z-index 😞 

Screen Shot 2021-09-21 at 12 36 34 PM

wilysword:
I mentioned in the description why the Sidebar has a z-index. Fixing the display of the layout item contents should be up to the devs, in my opinion: if someone actually added a banner like that they'd most likely have padding and/or add a z-index themselves.

mikesoltow:
> I mentioned in the description why the Sidebar has a z-index. Fixing the display of the layout item contents should be up to the devs, in my opinion: if someone actually added a banner like that they'd most likely have padding and/or add a z-index themselves.

I'll defer to you on the necessity of a z-index on the Sidebar.  However, I would think that it's a bit unintuitive to have to add a z-index manually to content in a fixed bottom bar like that.  I can't think of a reason why a developer _wouldn't_ want content in that bar to be rendered above other content, so I feel like that should be the default.  Could we maybe add a z-index ourselves to fixed position bars like that?

wilysword:
Okay, I added the `zIndex` option for fixed layout elements to make setting the z-index easier. In the end I decided _not_ to have a default, because then it would be difficult to remove the z-index if they didn't want it.

mikesoltow (approved):
This looks great.  Thanks, Glen!

PR #663, opened 2021-09-27
CACTUS-672 Replace Storybook knobs with controls

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-672

First commit is `cactus-icons`, second is `cactus-web`. Besides replacing `knobs` with `controls`, I also replaced `actions` with `console.log` functions with more targeted & useful information. Finally, I had to add another custom addon to work around some of the weirdness that is Storybook. All the stories should have at least as much control as they did before, but some of them should have even more (because I didn't want to go through disabling all the controls Storybook added automatically, unless necessary for a particular story).

Storyshot changes:
- Accordion: added padding so the outline variant wouldn't cut its edges off.
- IconButton: no idea what's causing this one, but doesn't seem like a big deal.
- Notification (tiny screens): this is actually more correct, as the default positioning on tiny screens is center, not right.
- TextInput (with status): renamed because I fixed a variable name (displayName -> storyName).
- Deleted three stories that didn't add anything, visually or functionally.

Another change you might notice is all the `as Meta` changed to `as const`. Typescript barely checks anything on an `as` type conversion, so we're not really losing anything there, and `as const` satisfies eslint as well. If we really want to type-check that value, I'd suggest adding a function like `storyMeta<C>(component: C, meta: Meta<C> = {})`.

**You can stop reading here**, the rest of this is me complaining about Storybook because this ticket took twice as long as it should have. At first, reading the docs I was like "this is pretty cool," and it was. The idea is great, and pulling docs/controls from the Typescript definitions is awesome. (Aside: it might be even better if PropTypes gave a way to recover their original parameters, like if you have `x = PropTypes.oneOf(options)`, then you could do something like `x.type == 'oneOf'; x.options`.) But then the problems started. Docs say you can hide a control by adding `table: { disable: true }`; I wasn't going to type that out a thousand times, so I made a constant. But inexplicably, some "hidden" controls would still show up in the table. It seemed completely random which ones, and when I used console.log to inspect the argTypes they appeared to be correctly disabled, they just weren't rendering that way. Then, I tried to do an array and, despite having a default value, it was rendering in the controls table as undefined. Then when I clicked the "Set Object" button, it defaulted to `{}` which broke the story because it was expecting an array, not an empty object. When I had two argTypes in the same file using the same `options` array, the second one would render as if the array was empty; but when I had a constant arg type (ICON_ARG) that I used in several files, that worked just fine...unless I used it twice in the same file. I'm not sure if they just have terrible state management or are extremely careless with mutability/side effects, but it's _very annoying_.

Eventually I decided the best thing to do was to write my own addon to control the side effects. Unfortunately, the docs on writing addons kind of suck (they assume writing for a standalone package, e.g. that would be distributed via npm), and the entire addon system is pretty weird: you _have_ to write two files, one of which basically just says where the other is; why can't you just point to the second file directly? (Later I realized it was because the first file is loaded in NodeJS on the backend, and the other is packaged with the front-end code; but I'm 99% sure they could omit or simplify that step with a little configuration in `.storybook/main.js`.) Then when I finally got it to recognize my addon, I had to hunt through their code for how to modify the argTypes, because that's basically undocumented. And then when I finally got that working, `argTypes` was empty on all the stories; which as it turns out is because they have them in two places, `context.argTypes` and `context.parameters.argTypes`, with (you guessed it) no documentation on why there are two or what the difference is.

\</rant>


Dhalton (approved):
Sounds like this was a huge pain but everything looks great and works well as far as the controls go.

I did find a bug with the BrandBar but it's definitely not related to your PR, so you definitely don't _need_ to fix it here. If you do want to sneak it in, it's a super easy fix. We just need to add `$isProfilePage={isProfilePage}` on [this](https://github.com/repaygithub/cactus/blob/master/modules/cactus-web/src/BrandBar/BrandBar.tsx#L288) line. If you're against fixing it here, let me know and I'll whip up a ticket for it real quick.

wilysword:
@Dhalton Do you think that's the fix needed for https://repayonline.atlassian.net/browse/CACTUS-725?

Dhalton:
> @Dhalton Do you think that's the fix needed for https://repayonline.atlassian.net/browse/CACTUS-725?

Yes, that's definitely the fix for that ticket. Didn't know that ticket existed. Whoops!

daniloPenaR (approved):
#Respect

PR #664, opened 2021-09-28
CACTUS-729 ::  Can't Clear DateInput

daniloPenaR:
[CACTUS-729](https://repayonline.atlassian.net/browse/CACTUS-719)
### PR description:
Basically, I put a new logic into the `getDerivedStateFromProps` method for clearing the input when the `value` prop is passed with a falsy value.  Then I wrote a new test case to cover that. 

If you want you could open DateInput.story.tsx file and modify the story `ControlledWithString` with the following 
```typescript
export const ControlledWithString = (): ReactElement => {
  const [value, setValue] = React.useState<Date | string | null>('2019-09-16')
  return (
    <>
      <DateInput
        disabled={boolean('disabled', false)}
        id="date-input-with-string-value"
        name={text('name', 'date-input-with-string-value')}
        value={value}
        format="YYYY-MM-dd"
        onChange={(e) => setValue(e.target.value)}
      />
      <Button ml={3} onClick={() => setValue(null)}>
        Clear Value
      </Button>
    </>
  )
}
```
and make sure that when you click on `Clear Value` the DateInput displays its placeholder. 

### Other steps for testing: 

- [x] `cd modules/cactus-web && yarn test:focus "./modules/cactus-web/src/DateInput/"`  Make sure all the test are in green ✅ .
- [x] Run the Storybook app and play with the different `DateInput` stories. Not only with the `Controlled with the string` one. Especially clearing the values and setting new dates.   
- [x] Test the Stories on IE11 and Firefox too.

### DIFF
--- a/modules/cactus-web/src/DateInput/DateInput.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.tsx
@@ -378,6 +378,7 @@ export interface DateInputProps
 
 interface DateInputState {
   value: PartialDate
+  prevValue: string | Date | null | undefined
   focusMonth: number
   focusYear: number
   locale: string
@@ -406,8 +407,10 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
 
     const initValue = props.value === undefined ? props.defaultValue : props.value
     const value = PartialDate.from(initValue, { format, locale, type })
+    const prevValue = props.value
     this.state = {
       value,
+      prevValue,
       focusMonth: value.getMonth(),
       focusYear: value.getYear(),
       type,
@@ -497,6 +500,10 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
         updates.focusMonth = value.getMonth()
         updates.focusYear = value.getYear()
       }
+    }
+    if (!props.value && props.value !== state.prevValue) {
+      updates = updates || {}
+      updates.value = PartialDate.from(null, { type: state.value.getType() })
     }
 
     return updates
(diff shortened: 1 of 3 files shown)


modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -498,6 +501,10 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
         updates.focusYear = value.getYear()
       }
     }
+    if (!props.value && props.value !== state.prevValue) {
+      updates = updates || {}
+      updates.value = PartialDate.from(null, { type: state.value.getType() })
```
wilysword, line 506: There are a couple of problems here:
1. State should never be changed unless `props.value !== undefined`.
2. You're never updating `state.prevValue`, which means that if the _initial_ value was '' or null (whichever you're using to clear), this doesn't work at all.
3. `PartialDate` has three parameters: type, locale, and format. Here you're losing two of them, so e.g. if you're passing `format="MM/dd/YYYY"`, then you clear the value, after that change events will be raised in the default format of `YYYY-MM-dd`.

wilysword (changes requested):

modules/cactus-web/src/DateInput/DateInput.tsx
daniloPenaR: 3. How is it better to preserve those parameters? I was looking that above we clone the state's value and then we made an assignment 
```typescript
      let value = state.value.clone()
      value = PartialDate.from(props.value, { type: value.getType() })

```

or if is it better to directly define those parameters as in the class definition 
```typescript 
      const locale = props.locale || getLocale()
      const type = props.type || 'date'
      const format = props.format || getDefaultFormat(type)
      updates.value = PartialDate.from(props, { type: state.value.getType(), locale, format })
```

wilysword: The second example you have there seems like it'd work.

PR #670, opened 2021-10-25
CACTUS-751 :: Default Text Element

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-751

The ticket doesn't mention it but there was a `Span` component we were exporting from `cactus-web`. I removed it since it wouldn't make sense to keep it with these changes.

### Testing
Run `yarn web start` and look at the `Text` story. Make sure everything looks as expected.


Dhalton:
Just realized some updates to the storyshots were generated...going to check those out real quick and make sure everything is OK there.

PR #671, opened 2021-10-28
CACTUS-703 :: Table & DataGrid Margins

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-703

### Notes
- The Table was setting `min-width: calc(100% - 32px);` for `fullWidth`, but obviously that won't work for customizable margins (that 32px was to account for the 16px of margin on each side). I don't know if there's a reason it needs to be `min-width`, but I changed to `width: 100%;` here because for some reason `min-width: 100%;` doesn't seem to take margins into account. I don't think that's a problem but if anyone has more insight on why that was setting min width let me know.
- I'm certain that there will be updates to the DataGrid storyshots (I removed the style overrides from the DataGrid basic usage story) but I've been having issues building/running the storyshot tests. I started a [slack thread](https://repay.slack.com/archives/CABGTU4S3/p1635455104077300) about it but as of now I don't know how to resolve it. I wanted to get this pushed up so we could start getting eyes on it while I'm out on PTO though.

### Testing
1. Run the storybooks and bring up the `Table` basic usage story. Play around with the margin prop and inspect the table wrapper to make sure the margins are correctly being applied.
2. Open the `DataGrid` storybook and set the margin using the control. Inspect the DataGrid and make sure the margins are being applied. Set `fullWidth` to true and make sure everything looks good.


wilysword (commented):
What you've done looks good, but I actually think the way fullWidth is currently implemented is wrong. It looks okay in Storybook, but I think that's actually because of the flex container: if you go into dev tools and remove all the styles from the `StyledContainer`, then make it widhth & height 100% instead of vh/vw. Then when you have fullWidth + margin, you can see it overflows. Not sure if you want to try to fix that now or make a new ticket.

Dhalton:
> What you've done looks good, but I actually think the way fullWidth is currently implemented is wrong. It looks okay in Storybook, but I think that's actually because of the flex container: if you go into dev tools and remove all the styles from the `StyledContainer`, then make it widhth & height 100% instead of vh/vw. Then when you have fullWidth + margin, you can see it overflows. Not sure if you want to try to fix that now or make a new ticket.

I'd prefer to fix it here rather than create a new ticket for it. I'll look into it today.

Dhalton:
As per our standup conversation, I won't be fixing this here since we're going to be doing more of a "global" fix. I will say, though, that this may be a tough problem to solve unless anyone has some ideas that I'm not thinking of.

I think we'll still have to use `calc` to subtract the margins from the width, so that leaves us with a couple of options (that I can think of):
1. Extracting any `marginProps` that would affect the horizontal margin before rendering, using them to calculate the offset we'd need on the wrapper (this would be ugly in order to account for `px`, `em`, and number values that correspond to the theme spacing) and then pass the offset to the wrapper.
2. Applying the margin styles to the wrapper, using a ref to get the computed left/right margin styles after it renders, then passing the offset to the wrapper so it can render again with the proper calculation.

Option 2 is probably the way to go I'd think, since we wouldn't have to do manual calculations that styled system is already doing, but I was having trouble implementing it (margin values were behind by one render) but there's probably a way to get it to work.

wilysword (approved):
I think a better option would be to replace `fullWidth` with just `width`, and maybe get rid of the wrapper. That way people can calculate their own widths if 100% isn't right. Or do padding _and_ margin; unless they really need margin collapse for some reason, padding + border-box would be better than margin since it'll automatically do the calculation for you.

Dhalton:
> I think a better option would be to replace `fullWidth` with just `width`, and maybe get rid of the wrapper. That way people can calculate their own widths if 100% isn't right. Or do padding _and_ margin; unless they really need margin collapse for some reason, padding + border-box would be better than margin since it'll automatically do the calculation for you.

I like that idea. Giving full control over the width is better imo.

Dhalton:
So I was able to update the storyshots after pulling from `node:14` in the storyshot Docker files instead of just `node`. That did change more storyshots than I was expecting, but I'm pretty sure it's something to do with the available fonts in that version of node. The only differences I saw were minor and they all just looked like changes to the font.

daniloPenaR (approved):

PR #672, opened 2021-10-29
CACTUS-759 :: Alert timeout Feature

daniloPenaR:
[CACTUS-759](https://repayonline.atlassian.net/browse/CACTUS-759)


modules/cactus-web/src/Notification/Notification.test.tsx
```diff
@@ -93,4 +93,23 @@ describe('component: Notification', () => {
...
+          </Alert>
+        </Notification>
+      </StyleProvider>
+    )
+
+    expect(queryByText('Notification')).toBeInTheDocument()
+
+    await waitFor(() => {
+      expect(setOpen).toHaveBeenCalled()
+    })
```
wilysword, line 113: I think you're supposed to return the expect result, like this (though if it's working, maybe I'm wrong?):
```suggestion
    await waitFor(() => expect(setOpen).toHaveBeenCalled())
```

wilysword (commented):
The Notification story makes sense, but I feel like the unit test should be in `Alert.test.tsx` instead. Functionally it looks good, though.

wilysword (approved):

Dhalton (approved):
LGTM 👍

PR #673, opened 2021-10-29
CACTUS-747 Modal Styles

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-747

Lots of info in the ticket, so be sure to read that for testing. If I made any decisions you guys disagree with, let me know.

BREAKING CHANGE: Modal is no longer flex by default; it also does not have automatic internal scrolling. Instead of default width, it now has default max-width.

Besides all the changes to Modal, I also made a slight change to Dimmer: brought the style to the outer component so it would be easier to override.

The only part I'm not really happy with is the scrolling: overflow and flex really don't work well together. I think I got it working well by default, but it's easy to mess up using the available style props.

The type library upgrade added a definition to the `system` function, that I needed to the `flexFlow` shortcut prop.


modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -1,172 +1,217 @@
...
+  Styled,
+  styledProp,
+  styledUnpoly,
+  styledWithClass,
+} from '../helpers/styled'
 import IconButton from '../IconButton/IconButton'
 
-export type ModalType = 'action' | 'danger' | 'warning' | 'success'
+export const modalType = ['action', 'danger', 'warning', 'success'] as const
+export type ModalType = typeof modalType[number]
```
Dhalton, line 41: Didn't know you could extract a type from an array like this. Pretty cool

Dhalton (approved):
LGTM

modules/cactus-web/src/Modal/Modal.tsx
wilysword: Yeah, the array has to have the `as const` at the end, otherwise the type is just `string[]`.

daniloPenaR (approved):
I am still amazed by that `convertPercentToFixed` solution.

PR #681, opened 2021-11-17
CACTUS-746 Label Props

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-746

I figured since `Text` already did what I wanted `Label` to do, why not just use it? I had to increase the CSS specificity (`&&&`) so that props could override the hard-coded styles correctly. While I was at it I added a couple of additional text styles by using the `typography` helper from `styled-system`. (I mostly did this to have fewer variables floating around, the extra styles were just bonus.)

In `AccessibleField` I mostly just moved the styles around a bit, from a subsection of `AccessibleField` to a separate component. The `disabled` color, though, I moved inside the component so I could be sure it would override any custom color passed in. I made the margins symmetrical when there's no tooltip, which is kind of a breaking change but feels more like a bug fix.

`FileInputField` had duplicated the styles from `AccessibleField`, so I removed those. This led to the second breaking-but-not-really change, replacing the weird relative positioning on the label and the tooltip with a margin. The resulting visual placement is the same, but it makes the wrapper div taller which theoretically has an infinitesimal chance of messing up layouts.


daniloPenaR (commented):
Is it not necessary to include any kind of update in the documentation?

wilysword:
The props table is automatically generated, I double-checked that it successfully pulled in the new props. I didn't think any of the rest of the documentation was affected by the changes, did you have a specific change in mind?

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -208,34 +212,42 @@ function AccessibleFieldBase(props: InnerProps) {
...
   )
 }
+AccessibleFieldBase.displayName = 'AccessibleField'
 
-export const AccessibleField = styled(FieldWrapper)
-  .withConfig(omitProps<AccessibleFieldProps>(width, margin, flexItem))
-  .attrs({ as: AccessibleFieldBase })`
+export const AccessibleField = styledUnpoly(FieldWrapper, AccessibleFieldBase).withConfig(
+  omitProps<AccessibleFieldProps>(width, margin, flexItem)
+)`
```
daniloPenaR, line 219: So this `styledUnpoly` is implemented there just to restrict that AccessibleFieldBase from not being replaced with any other component/tag via `as` prop?

daniloPenaR:
> The props table is automatically generated, I double-checked that it successfully pulled in the new props. I didn't think any of the rest of the documentation was affected by the changes, did you have a specific change in mind?

Oh no Just had that concern related to the props, but nothing more  specific.

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
wilysword: Correct. The code that was there before did the same thing, but I recently added the `styledUnpoly` helper to the repo so I figured as long as I was doing some refactoring I might as well use it.

Dhalton (approved):
LGTM

Dhalton (commented):
I realized after reviewing Mike's PR that the reason the integration tests started hanging again is because those `SplitButton` changes went out with cactus-web v9 and the example apps were still using v8. I bet we don't need to delete the part of that test.

Dhalton (approved):
LGTM

daniloPenaR (approved):
👍

PR #682, opened 2021-11-18
Upgrade web and icons packages on non-published apps

mikesoltow:
https://repayonline.atlassian.net/browse/CACTUS-788


Dhalton (approved):
Ah, so _this_ is why the integration tests started failing in master. Example apps were still pulling in `cactus-web` v8. Sorry I should've done this when I released. Thanks for taking care of it.

wilysword (approved):

PR #683, opened 2021-11-18
CACTUS-721 :: Tab/TabList Flex Styles

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-721

I realized after digging into this that you can already achieve the look that the ticket is describing without these features just by using the `TabList` props, but I don't see any harm in exposing the flex item props in case devs want more control, and `justifyContent="stretch"` is already technically supported, this just adds TypeScript/PropTypes support for it.

Also I cherry-picked Mike's commit for upgrading cactus deps in the example apps since it should fix the CI here.

### Testing
I added the flex item controls to both of the `Tabs` stories so you can set some values for `flexGrow`, `flexShrink`, and `flexBasis` to see how they affect the tabs. Note that these styles will apply to all of the tabs. You can also see that `stretch` is available as a `justifyContent` option too.


modules/cactus-web/src/Tabs/Tabs.story.tsx
```diff
@@ -33,19 +35,30 @@ export default {
   component: TabList,
 } as const
 
-export const BasicUsage: Story<typeof TabList, { tabCount: number }> = ({ tabCount, ...args }) => {
+export const BasicUsage: Story<typeof TabList, FlexItemProps & { tabCount: number }> = ({
+  tabCount,
+  ...args
+}) => {
   const [active, setActive] = React.useState(-1)
+  const tabListProps = split(args, 'justifyContent', 'fillGaps', 'fullWidth')
```
wilysword, line 43: If you're removing the flex props on the next line, I'd think what was left in `args` would be the same as before, so you wouldn't need this line at all.

modules/cactus-web/src/Tabs/Tabs.story.tsx
```diff
@@ -56,35 +69,57 @@ export const BasicUsage: Story<typeof TabList, { tabCount: number }> = ({ tabCou
...
-      <h2>Cloudy Nebula</h2>
-    </TabPanel>
-    <TabPanel id="serial-box" tabId="serial-num" mx={5} my={4}>
-      <p>N3BU14</p>
-    </TabPanel>
-  </TabController>
-)
-WithController.args = { justifyContent: 'space-evenly' }
+export const WithController: Story<typeof TabList, FlexItemProps> = (args) => {
+  const tabListProps = split(args, 'justifyContent', 'fillGaps', 'fullWidth')
```
wilysword, line 89: Same here.

wilysword (approved):
I don't really mind this, I guess, although the primary cause you mentioned (stretching across the screen) could have already been done using `fillGaps`, couldn't it?

Dhalton:
> I don't really mind this, I guess, although the primary cause you mentioned (stretching across the screen) could have already been done using `fillGaps`, couldn't it?

Yeah it could've, which was an oversight on my part. I still think it could be useful to expose this level of control, though.

modules/cactus-web/src/Tabs/Tabs.story.tsx
Dhalton: Ah okay, I had never used this `split` helper before & I didn't notice it updates the original object. I'll update it

daniloPenaR (approved):
LGTM.

PR #684, opened 2021-11-30
CACTUS-722 :: display: block stories

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-772

### Some General Notes
- To see the option to render stories in the old flex container, you'll probably have to run the stories with the `--no-manager-cache` flag.
- The ticket says that the container should have 100% width and height. 100% width is fine, but the storybook root element has no height, so `height: 100%` does nothing. That's why I'm still using `100vh`.
- Any stories that are intended for testing collisions are still rendered inside a flex container so that they're still useful.

I'm splitting my findings into two categories: Things I think we should fix within the components themselves, and things that we don't necessarily need to fix, but that look strange and we may want to consider adjusting how they're rendered in the stories.

### Things We Should Fix
- `Card`: Cards default to `display: block`, stretching them across the screen. Should we consider making them `display: inline-block` to alleviate this?
- `ColorPicker`: It stretches across the whole screen. I _think_ I'd consider this something we should fix, given that it's not really a "field" element. Setting it to `display: inline-flex` would fix it. Side note: should we add a `ColorPickerField` component that wraps it in `AccessibleField`?
- `List`: Again, it takes up the entire width, which is evident in the "with Dividers" story. Setting `display: inline-block` would fix this.
- `Select`: Takes up the whole width. This one actually has `width: 100%` hard-coded in the trigger's CSS. My suggestion would be to remove that; it doesn't fit in with our other (non-field) form elements.
- `Table`: Horizontal margins cause the table to overflow in the `display: block` container when `fullWidth` is set to `true`. This is probably the trickiest to solve…The problem is that `fullWidth` sets the wrapper width to 100%, but that doesn’t account for margins. Before margins were customizable, we were using `calc()` to subtract the hard-coded margins from the width. If we want to do that again, we’ll have to create a way to determine what the horizontal margins will be based on the margin props, which could be difficult. Open to other ideas on this though.
- `Text`: Looks weird with the default margin styles in the story. It’s `display: inline`, but I wonder if we should change it to `display: inline-block` so that margins are applied correctly even if the text wraps to a new line.

### Things That Look Weird But Aren't Bugs

The components listed below all look weird in the `display: block` container for the same reason: they stretch across the whole screen. I only mention these because I wonder if we want to adjust the stories themselves to make them look better. On the one hand, the way they’re currently rendered is their “true” form. On the other hand, because the containing div is so wide, it just looks weird in the story. I could go either way.

- `AccessibleField`
- `CheckBoxGroup`/`RadioGroup`
- `FileInputField`
- `SelectField`
- `TextAreaField`
- `TextInputField`

Please have a look at at least the stories I mention above, if not all of them. Let me know if there's any weird-ness you think I may have missed, and please give me your thoughts on my proposed fixes/whether or not you think they _should_ be fixed.


daniloPenaR (commented):
Most of them are looking OK. 
I agree with that list of components you made. Components like `AccessibleField` won't look good by taking the entire width of the screen on larger screens and this could take'em away from the desired representation we want. (and the purpose of this ticket I guess)

One question that comes to my mind is, on the ticket ACs there is an option to render a story inside a flex container just as before. I didn't see that on the Cactus Theme options tab. Am I missing something?

Dhalton:
> One question that comes to my mind is, on the ticket ACs there is an option to render a story inside a flex container just as before. I didn't see that on the Cactus Theme options tab. Am I missing something?

The answer to this question is under the "General Notes" section in the PR description.

> To see the option to render stories in the old flex container, you'll probably have to run the stories with the `--no-manager-cache` flag.

modules/cactus-web/cactus-addon/index.tsx
```diff
@@ -65,36 +69,45 @@ const ProvideCactusTheme: React.FC<ProvideCactusThemeProps> = ({
...
       setTheme(generateTheme(params))
     }
 
-    const updateBackground = ({ inverse: newInverse }: any): void => {
+    const updateBackground = ({ inverse: newInverse }: any) => {
       setInverse(newInverse)
     }
 
-    const updateBorderBox = ({ borderBox: newBorderBox }: any): void => {
+    const updateBorderBox = ({ borderBox: newBorderBox }: any) => {
```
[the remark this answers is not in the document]
Dhalton: Just cleaning things up. You don't need an explicit return type on a function if it's not exported.

Dhalton:
> Components like `AccessibleField` won't look good by taking the entire width of the screen on larger screens and this could take'em away from the desired representation we want. (and the purpose of this ticket I guess)

I'm confused by this...can you elaborate on what you mean/what course of action you're alluding to here?

daniloPenaR:
> > Components like `AccessibleField` won't look good by taking the entire width of the screen on larger screens and this could take'em away from the desired representation we want. (and the purpose of this ticket I guess)
> 
> I'm confused by this...can you elaborate on what you mean/what course of action you're alluding to here?

What I am trying to say is that I agree that the components you listed in both `Things We Should Fix` and `Things That Look Weird But Aren't Bugs` and with your proposed fixes for them.

daniloPenaR (approved):
LGTM At this point. Let me know if you have any additional changes.

modules/cactus-web/cactus-addon/index.tsx
```diff
@@ -37,20 +38,23 @@ const alignmentMap: { [k in AlignmentTypes]: CSSObject } = {
...
+  ${(p) => (p.inverse ? p.theme.colorStyles.base : p.theme.colorStyles.standard)};
+  ${(p) => p.overrides}
+  ${(p) => p.borderBox && '* { box-sizing: border-box; }'}
+`
 
-  ${(p): string => p.borderBox && '* { box-sizing: border-box; }'}
+const FlexContainer = styled(DefaultContainer)`
+  width: 100vw;
+  display: flex;
+  ${(p) => alignmentMap[p.align]};
```
wilysword, line 54: So if you look at the DateInput/Controlled with string story, it's supposed to have `alignItems: 'start'` in the overrides but this line takes precedence. The easiest way to get around it would probably be to define the overrides like this:
```
&& {
  ${(p) => p.overrides}
}
```
(I don't actually think it looks bad with the center alignment, in fact it's more consistent with the other stories, but there could be other overrides where it matters more.)

modules/cactus-web/src/Table/Table.story.tsx
```diff
@@ -64,7 +64,7 @@ export const Layout: TableStory = ({
   }
 
   return (
-    <Table {...args}>
+    <Table {...args} marginLeft={2} marginRight="25px">
```
wilysword, line 67: I'm not at all clear on why you added this, especially since it prevents the `margin` control from working correctly.

wilysword (approved):
I think pretty much all the stories look good like this, although I'm iffy on the Table story, and a few stories could probably use more refinement but it needn't be done in this ticket.

`Card` should definitely stay as block; if you look at how we usually use it in channels, it's mostly as a column or section. It would pretty much never be used without margins (or padding in an outer container), so I don't see much of an issue there. `List` is probably similar.

I agree with you regarding `ColorPicker` and `Select`, but it's actually complicated by the fact that although the visible part of `TextInput` doesn't stretch, it actually has a block wrapper around it that does. I'd prefer if all our simple inputs were inline elements, but it would be a breaking change so I'm not sure what the best path to standardization would be at this point.

For `Table` we should probably remove the default margin and add an actual `width` prop, instead of `fullWidth`, so the dev can do any needed calculations themselves. I still think we were often too restrictive of what styling we allowed in the early days of the team, and that's come back to bite us quite often (Modal, DataGrid & Table, Label, etc.).

`Text` does look a little weird, but it's mostly a styling component. I'd be open to supporting a `display` prop, but I think most of the time you can easily change that behavior just by using `as="p"`, or `as="div"` for a generic block element without margins; the need for `inline-block` is pretty rare.

For the rest of the field components, it's basically the same as Card and List: yeah they look a little weird when they're too wide, but that's more of a page design issue than a component design issue.

modules/cactus-web/cactus-addon/index.tsx
Dhalton: Ah, thanks for catching that. I'll update it

modules/cactus-web/src/Table/Table.story.tsx
Dhalton: Yeah these are definitely some test props I was using to investigate and forgot to remove. I'll get rid of them

Dhalton:
> For `Table` we should probably remove the default margin and add an actual `width` prop, instead of `fullWidth`, so the dev can do any needed calculations themselves. I still think we were often too restrictive of what styling we allowed in the early days of the team, and that's come back to bite us quite often (Modal, DataGrid & Table, Label, etc.).

Yeah, this thought definitely crossed my mind too. Do you know if the styled-system `width` helper function would know how to handle a calculated width value? Something like: `width="calc(100% - 32px)"`. I assume it would but I haven't tested anything like that myself before.

wilysword:
Yes, all styled-system functions accept arbitrary strings. The only disadvantage in this case is that you'd have to use actual pixel values, where the `margin` props would most likely use indexes to the `space` array. (I wonder if it would be possible to make some kind of styled components `calc` helper...)

PR #685, opened 2021-12-08
CACTUS-790 Cactus Forms

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-790

Pretty straightforward. I copied `cactus-i18n` and then renamed stuff, so hopefully I got everything configured that I need to. Compare to https://github.com/repaygithub/repay-api/blob/master/shared_ui/components/FinalField.js which is where the originals came from.

For the test commit, I strongly recommend you review with ignored whitespace. `rules-fields.tsx` is just the original `rules.tsx` renamed, though it didn't show up that way in git. (I did change/remove the `disableUp`/`disableDown` thing.) While I was writing the tests I realized a bit of future work could be a `FieldArray` component, since I hate the interface of the one in `react-final-form-arrays`.


modules/cactus-form/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@repay/cactus-form",
+  "version": "1.0.0",
```
Dhalton, line 3: Should we start this off at version 0.0.1 or 0.1.0? The only reason I ask is that I'm pretty sure lerna will make me choose from a set of versions that are higher than what's in the package.json, and I wouldn't want the initial release for this package to be 1.0.1

modules/cactus-form/src/Field.ts
```diff
@@ -0,0 +1,194 @@
...
+  processMeta?: PostProcessor
+  getFieldComponent?: ComponentMapper
+}
+
+const DEFAULT_SUB: FieldSubscription = {
+  value: true,
+  error: true,
+  touched: true,
+  submitError: true,
+  modifiedSinceLastSubmit: true as any,
```
Dhalton, line 55: I'm curious why the need for `as any` here?

modules/cactus-form/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Documentation is available in the [root of the repository](https://github.com/repaygithub/cactus/tree/master/docs/Forms/).
+
+## Contributing
+
+Use the example applications for implementation testing and UAT.
+
+### Commands
+
+> All of these commands can be run from the root of the repository using the `i18n` shortcut. (e.g. `yarn i18n test`)
```
Dhalton, line 15: Still some leftover references to the i18n package here.

modules/cactus-form/src/Field.ts
```diff
@@ -0,0 +1,194 @@
...
+      props.type = 'checkbox'
+      return CheckBoxField
+
+    case 'radio':
+      props.type = 'radio'
+      return RadioButtonField
+
+    case 'select_multiple':
+    case 'select-multiple':
+    case 'multiSelect':
```
Dhalton, line 83: Do you think we could/should ever standardize this to just `multiSelect`? I'm pretty sure it's written this way to be compatible with Channels, but I'd certainly like it better if there weren't multiple keys that map to the same object. I'm not entirely sure the refactor work that would entail in Channels, but I think it's something to consider especially since other teams will use these components eventually.

examples/mock-ebpp/src/index.tsx
```diff
@@ -34,6 +35,7 @@ const App = (): React.ReactElement => {
             <UIConfig path="/ui-config" />
             <Faq path="/faq" />
             <Rules path="/rules" />
+            <RulesFields path="/rules-fields" />
```
Dhalton, line 38: Is there a particular reason you left this out of the MenuBar itself? I guess it makes sense not to advertise that there are duplicated pages here, but on the other hand, we're really the only people who use these example apps so I don't think it'd hurt to make it easier to get to the new page. Honestly I'm fine either way, I just wanted to make sure it's not something you forgot.

Dhalton (approved):
Looks good overall. I have a few comments/questions but nothing major so I'm going to approve

examples/mock-ebpp/src/containers/rules.tsx
```diff
@@ -1,361 +1,265 @@
...
 
 import { post } from '../api'
 import FieldsAccordion from '../components/FieldsAccordion'
 
-let ruleKeyIndex = 0
-let conditionKeyIndex = 0
-let actionKeyIndex = 0
-
-type RulesState = {
+interface Field {
```
daniloPenaR, line 13: Is it good to keep the same name of `Field` for this interface even if there's an import with the same Name?

modules/cactus-form/src/Field.ts
wilysword: The types defined in `final-form` inexplicably use `false` instead of `boolean` as the type of that key. The rest are `boolean`, so I have no idea what the author was thinking.

modules/cactus-form/src/Field.ts
wilysword: `select-multiple` is what you get if you render a `<select multiple ref={ref}>` element and do `ref.current.type`. What's wrong with having multiple keys map to the same input type?

examples/mock-ebpp/src/index.tsx
wilysword: Personally I don't see much reason to even have the page, and in the long term I figure we'll probably get rid of it. The integration tests cover both right now, though, so I see little reason to link to two identical pages when one of them allows you to test everything you need.

modules/cactus-form/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "main": "dist/cactus-form.cjs.js",
+  "module": "dist/cactus-form.js",
+  "types": "dist/index.d.ts",
+  "repository": "https://github.com/repaygithub/cactus/tree/master/modules/cactus-form",
+  "license": "MIT",
+  "scripts": {
+    "cleanup": "rm -rf dist coverage",
+    "predev": "yarn run cleanup && yarn build:types",
+    "dev": "repay-scripts dev --lib src/index.ts",
+    "build": "repay-scripts build --lib src/index.ts",
```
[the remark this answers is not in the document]
wilysword: No idea. I just copied this from `cactus-i18n`, and that was written long before I became a member of this team.

examples/mock-ebpp/src/containers/rules.tsx
wilysword: I generally try to avoid it, although clearly neither Typescript nor JavaScript have a problem with it. Since I'm already making a couple of other changes, I guess I'll change this too.

modules/cactus-form/package.json
wilysword: Done. Whoever does the release will need to remember to change the `mock-ebpp/package.json` as well.

modules/cactus-form/src/Field.ts
Dhalton: It just seems like a confusing API to me, but if there's a reason it has to be that way then that's fine. I know this is the current implementation in Channels so I wouldn't want to break it for them

daniloPenaR (approved):

Dhalton (approved):
Updates look good to me

daniloPenaR (approved):

PR #686, opened 2021-12-09
Cactus 715:: Dependency updates

daniloPenaR:
[CACTUS-715](https://repayonline.atlassian.net/browse/CACTUS-715)

Finally. 

## Notes

- `Storybook upgrades`: As we talked a few days ago, it is preferable to not bother on upgrading Storybook yet and _maybe_ wait ’til version 7 is released.
- `@fluent`: Here we have a library that is currently implementing `ESM` in their latest version, and this breaks our code when compiling, more of this discussion [here](https://repay.slack.com/archives/CABGTU4S3/p1637711842044300?thread_ts=1637260186.024400&cid=CABGTU4S3)
- `puppeteer`:  Didn't perform any upgrades due to incompatibilities with our current storybook version. 
- **`Loose warning on babel config:`** Here is a warning `The "loose" option must be the same for @babel/plugin-proposal-class-properties....` so I added the option to the plugin section in every babel config file. Let me know if you still seeing that warning. 

## Testing
After pulling down the branch, run `yarn && yarn cleanup && yarn build`

### Website
Run `yarn build` && `yarn start`. Let me know if there's some inconsistency in the docs page. 
The `Gatsby` migration from v3 to v4 was quite ''simple'' so I hope it'll build successfully. 

### Modules
Run all the tests, and start all the applications and make sure everything is ok. 

### Examples
I've tried running all the tests and starting all the examples with every upgrade.


wilysword (approved):
I applaud your patience, sir. Things seem to be working fine.

wilysword:
I am in fact still seeing that "loose" warning. I noticed that `ui-tools/modules/babel-preset/index.js` (on branch master) is not the same as `cactus/node_modules/@repay/babel-preset/index.js`. Not sure if that means that change just hasn't been released yet, or if we need to upgrade the version of `@repay/babel-preset` in this repo. I'm not sure if we need all the `.babelrc` changes; it might be enough to just upgrade the babel-preset.

Dhalton (commented):
Looking good. I am still getting the warning for the "loose" option. I don't know why, though...I upgraded to the latest `@repay/babel-preset`, which includes `["@babel/plugin-proposal-private-property-in-object", { "loose": true }]` but it didn't help

Looks like some storyshots need to be updated too

wilysword:
I was kind of wondering about that, but I think `["@babel/plugin-proposal-private-property-in-object", { "loose": true }]` is wrong. All the other plugins in that file are like `[require("@babel/plugin-proposal-private-property-in-object").default, { "loose": true }]`. If that's the case, we'd need a new ticket for fixing it in that repo, then a new release, _then_ we could fix it in cactus; I don't think we'd need to hold up this ticket if that's the case, though.

Dhalton (dismissed):
LGTM

Dhalton:
So I had approved and was getting ready to merge this to master, but I wanted to test some things locally to make sure everything is good, but I came across something we'll need to fix before this is is merged. When I tryo run any of the example apps locally with `yarn start`, I get this error:
```
*** Failed to compile. ***

[object Object]


	Review the above errors to fix the build.```
Obviously that's not a very descriptive "error message" though

**Comment by Dhalton** [comment]
Something with the lockfile definitely just got messed up during one of the merge commits. To fix it locally, I just deleted the `yarn.lock`, then re-installed everything

File: modules/cactus-fwk/.babelrc [hunk 1]
```diff
@@ -1,4 +1,5 @@
 {
   "presets": ["@repay/babel-preset"],
-  "sourceMaps": "inline"
+  "sourceMaps": "inline",
+  "plugins": [["@babel/plugin-proposal-private-property-in-object", { "loose": true }]]
```
Dhalton, line 4: Can you remove all of these additions to the babel configs since we'll be getting a fix in for it in the `babel-preset` package itself

daniloPenaR:
I am afraid that copying the file from master maybe cause some issues building the rest of the project

daniloPenaR:
I've followed what you did. Deleting the yarn.lock file and copying from the master, then execute `yarn install` again, but still printing the same error.  One fact that is confusing me, is that the localhost and the applications seems to be compiled and those can be checked in the localhost server where those are running.

Dhalton:
> One fact that is confusing me, is that the localhost and the applications seems to be compiled and those can be checked in the localhost server where those are running.

Yeah that's because it's something to do with a missing webpack dev server dependency. If you upgrade to the latest version of repay scripts it shows the actual error message

daniloPenaR:
So is it Okay if I push up the current changes in the `Yarn.lock` even if the message is printed out?

Dhalton (approved):
Everything seems to be working now as far as I can tell. Thanks for sticking this out Danilo

PR #689, opened 2021-12-15
CACTUS-775: update node version to LTS

wilysword:
[CACTUS-775](https://repayonline.atlassian.net/browse/CACTUS-775)

The ticket was for upgrading to 16. If I'd used `node:16` it would probably be more stable, but I suspect we'd actually _want_ to know when the next LTS version comes out, even if it means we have to fix some stuff when it happens; if I'm wrong about that, though, it's easy enough to fix.

Here's the image information showing that `node:lts` is currently version 16: https://hub.docker.com/layers/node/library/node/lts/images/sha256-534004248435dea5cecf3667232f800fc6bd101768306aecf6b87d78067b0563?context=explore


Dhalton (approved):
I'm a little surprised that this didn't require any fixes; I know it will in at least `ui-tools`, but I guess not here! I'd prefer to use `lts` so I'm good with this 👍

daniloPenaR (approved):

PR #690, opened 2021-12-15
CACTUS-711 :: Backstage Docs POC

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-711

So what I've learned in researching & working on this is that Backstage entities (systems, components, groups, etc) are "separate" from techdocs, but you can't push up techdocs without an entity to associate it with. So although this ticket was really only asking for techdocs, I created a base entity structure for our team and the packages in cactus so that they'd be visible in Backstage. I think we should add entities for our tools in `ui-tools` & `ui-private`, but that config should live in those repos.

You'll see that I moved everything that was previously in the `docs` folder into a new folder called `website-docs`. I had to do this because mkdocs expects your techdocs to be in a `docs` folder at the root of the project.

To run the docs locally, pull down this branch and run `yarn techdocs`. That should spin up a docker container and then you can visit `http://localhost:3000` to view & edit the docs.

We have a new Jenkins job to publish the docs, that can be found [here](https://jenkins.internal.repay.ninja/blue/organizations/jenkins/cactus/activity). It'll run any time it detects changes to master (right now it's actually set up to pull from this branch but I'll change that when this gets merged).

To view the published docs & our entities in Backstage, go [here](https://backstage.internal.repay.ninja/catalog/default/group/ui-team). That link will take you directly to the UI Team's page in Backstage. There, you can navigate to the entities associated with cactus, as well as view the techdocs.

Let me know if you have questions or concerns!


wilysword (approved):
I had to `rm -rf $(find -name node_modules)` before building the docs website would work, but I checked that and the Backstage docs, everything looks good.

Dhalton:
So after discussing with the team, we decided it would be better to house the techdocs in `ui-private`, so I removed everything techdocs-related from this PR and I'll add it in a PR over there. This is now purely the entity configuration for the UI team & cactus components. @wilysword you may want to re-review

Dhalton:
It's also possible that we have a security concern with using the repay cli in the Jenkinsfile here since this is public, so we may end up moving all of that to `ui-private` as well, so this is on hold until we hear back about whether or not it's alright to keep that in this repo.

Dhalton:
I removed the Jenkins pipeline from this PR since we'll be doing the publishing from `ui-private`

wilysword (approved):

daniloPenaR (approved):

.backstage/catalog-info.yaml
```diff
@@ -0,0 +1,234 @@
...
+    email: gphelps@repay.com
+  memberOf: [ui-team]
+---
+apiVersion: backstage.io/v1alpha1
+kind: User
+metadata:
+  name: dpena
+spec:
+  profile:
+    displayName: Danilo Peña
```
daniloPenaR, line 191: Thanks for adding the ñ character 😄

PR #694, opened 2021-12-21
CACTUS-799 Story Cleanup

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-799

For most of them I just followed what the ticket says and restricted the max width on the stories (which is why very few of the iPhone storyshots were updated). For Card, I also added a `maxWidth` style and used that in the story (because (a) it already supported `width` so I saw no harm in supporting `maxWidth`, and (b) it was the first one I did and I hadn't thought up the max-width override I used on all the other stories yet). Finally, I did a little style refactoring on some of the components, switching to newer helpers and adding a `&&` specificity increase around the prop styles to make sure they override the default styles. (That's something I'm probably going to start doing more often, since it makes the styles more predictable.)

I'd suggest reviewing one commit at a time, and each commit message also has some more detailed information about my reasoning.


Dhalton (approved):
LGTM

daniloPenaR (approved):

PR #696, opened 2021-12-24
Cactus 795 :: Sonarqube

daniloPenaR:
[CACTUS-795](https://repayonline.atlassian.net/browse/CACTUS-795)

There's a Bug on the `Accordion` Component: 
```typescript 
onEnter: function (this: { height: number }, node: HTMLElement) {
    this.height = node.getBoundingClientRect().height
    node.style.height = '0'
    node.scrollTop // Force reflow.
 },
```
As far as I know, `node.scrollTop` returns the number of pixels scrolled. So, Sonarqube reports this: 

> Expected an assignment or function call and instead saw an expression.

I wonder if that logic is intentional inside the onEnter event.


wilysword (approved):
The call to `scrollTop` is indeed intentional, that's why there's a comment there explaining what it's doing (forcing a reflow). I copied that from the `react-transition-group` library.

Dhalton (approved):

PR #698, opened 2021-12-29
CACTUS-782 Input refs

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-782

Individual commit messages have a bit more info, but the short story is that I did four main things:
- Added ref support where it didn't already exist.
- Added tests to prove that the refs are assigned to the correct element and that Typescript has the correct type.
- Refactored the styles with the theme helpers.
  - On TextInput and TextArea, I changed how the width/height props are applied so that the resulting element is actually the correct dimension, not just visually so.
  - On CheckBox and RadioButton I also moved the disabled styles so we wouldn't have to pass that prop to the visible psuedo input.
- Made `id` optional on CheckBox and RadioButton, and `name` optional on Checkbox, since there's no reason for it to be required. (Not related to the ticket at all, of course, I just got annoyed when writing the tests and Typescript told me I had to add an `id`...)

### DIFF
--- a/modules/cactus-web/src/CheckBox/CheckBox.story.tsx
+++ b/modules/cactus-web/src/CheckBox/CheckBox.story.tsx
@@ -1,16 +1,16 @@
 import React from 'react'
 
 import { CheckBox } from '../'
-import { actions, HIDE_CONTROL, Story } from '../helpers/storybook'
+import { actions, Story, STRING } from '../helpers/storybook'
 
 export default {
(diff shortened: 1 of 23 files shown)


daniloPenaR (approved):
I don't have major comments about this. Looks good to me. I left my approval now cause I'm going to get some vacation.

Gracias Glen, Happy new year.

Dhalton (approved):
LGTM

PR #700, opened 2022-01-05
CACTUS-745 :: StatusMessage Updates

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-745

I did a bit of refactoring so you may want to review commits individually. Mainly the refactoring was to clean up the component and use some of the newer style helpers, but there's also some refactoring I had to do to ignore the `info` status for other components.

We were previously importing the `Status` type from the `StatusMessage` file for components that also support those props, but since those components don't support the new `info` status, I moved the type definition to `helpers/status` and then extended from that type to include `info` for `StatusMessage`.

### Testing
1. Open up the new `StatusMessage` story and check out the info variant to make sure it looks right.
2. Inspect the page and set `color: white` div containing the `StatusMessage`s. Make sure the text inside the messages don't reflect that white foreground color.


modules/cactus-web/src/StatusMessage/StatusMessage.tsx
```diff
@@ -1,73 +1,70 @@
...
+  success: StatusCheck,
+  warning: NotificationAlert,
+  error: NotificationError,
+  info: NotificationInfo,
 }
 
 const Noop = (): null => null
 
-const StatusMessageBase: React.FC<StatusMessageProps> = ({
+const StatusMessage: React.FC<StatusMessageProps> = ({
```
wilysword, line 43: What was the reasoning for this change? I admit I've done similar changes in the past, but that's mostly in situations where the internal logic is significant enough that the `as` prop has no value; in this case it looks like the only internal logic is adding a role and an icon, so I'm not sure that's the case. Still, I probably wouldn't have even mentioned this, except it affected the styles in several other places...

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
```diff
@@ -230,7 +231,7 @@ export const AccessibleField = styledUnpoly(FieldWrapper, AccessibleFieldBase).w
     padding: 0 ${space(3)};
   }
 
-  .field-status-row ${StatusMessage} {
+  .field-status-row div {
```
wilysword, line 234: This is the one that gives me greatest pause: the `children` passed to `StatusMessage` can be any node, which could potentially include another div.

modules/cactus-web/src/StatusMessage/StatusMessage.tsx
```diff
@@ -1,73 +1,70 @@
...
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
+import { margin, MarginProps } from 'styled-system'
 
-import { textStyle } from '../helpers/theme'
+import { Status as BaseStatus } from '../helpers/status'
 
-export type Status = 'success' | 'warning' | 'error'
+type Status = BaseStatus | 'info'
+type ColorKey = 'successLight' | 'warningLight' | 'errorLight' | 'lightContrast'
```
wilysword, line 16: I'd suggest using `CactusColor` from `@repay/cactus-theme` instead.

modules/cactus-web/src/StatusMessage/StatusMessage.tsx
```diff
@@ -1,73 +1,70 @@
...
+const StyledStatusMessage = styled.div<{ $status: Status }>`
   padding: 2px 4px;
   position: relative;
   box-sizing: border-box;
   overflow-wrap: break-word;
   display: inline-block;
-  ${(p) => textStyle(p.theme, 'small')};
-  ${(p) => statusMap[p.status]}
+  ${textStyle('small')};
+  ${(p) => colorStyle(statusMap[p.$status])}
```
wilysword, line 64: This works, but involves an extra function call. I'd suggest calling it in the second form instead:
```suggestion
  ${(p) => colorStyle(p, statusMap[p.$status])}
```

modules/cactus-web/src/StatusMessage/StatusMessage.tsx
```diff
@@ -1,73 +1,70 @@
...
+  ${margin}
 
-  ${NotificationError}, ${NotificationAlert}, ${StatusCheck} {
+  ${NotificationError}, ${NotificationAlert}, ${StatusCheck}, ${NotificationInfo} {
     margin-right: 4px;
     vertical-align: -2px;
```
wilysword, line 69: I'm not sure if it would be better or not, but icons support both margin and vertical align props; do you think it would be better/cleaner to pass them as props instead?

modules/cactus-web/src/StatusMessage/StatusMessage.tsx
```diff
@@ -1,73 +1,70 @@
...
+const StatusPropType = PropTypes.oneOf<Status>(['success', 'warning', 'error', 'info'])
 
-interface StatusMessageProps extends React.HTMLAttributes<HTMLDivElement> {
+interface StatusMessageProps extends React.HTMLAttributes<HTMLDivElement>, MarginProps {
   status: Status
 }
 
-type StatusMap = { [K in Status]: ReturnType<typeof css> }
+type StatusMap = { [K in Status]: ColorKey }
+type IconMap = { [K in Status]: React.ElementType<any> }
```
wilysword, line 25: If you do decide to pass props to the icon, I'd say change this from `Reac.ElementType<any>` to `React.ElementType<IconProps>` or just `typeof StatusCheck`.

wilysword (commented):
The styling looks good, but I have some concerns.

modules/cactus-web/src/StatusMessage/StatusMessage.tsx
Dhalton: Yeah I mean I mainly just like this pattern better & wanted to use it here, although it's not necessary. I see your concerns with the styling in other places, though. So I'll revert it

modules/cactus-web/src/AccessibleField/AccessibleField.tsx
Dhalton: Good point

modules/cactus-web/src/StatusMessage/StatusMessage.tsx
Dhalton: Yeah I'd say either that or just changing this selector to `svg` to make it more general. But I'll go with your suggestion since that's less prone to unintended side effects.

wilysword (approved):

PR #701, opened 2022-01-06
CACTUS-727: utilize `patch-package` to fix react-final-form

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-727

This is now framed as an optional (but recommended) step, because technically everything in `cactus-form` does work just fine without it, it's only some interactions with specific `cactus-web` components that suffer. Even so, the amount of extra setup is no greater than patch-package itself, so I think it's about as good as we can get.

I tested it using yalc & a newly generated bare-bones app, but I'm not 100% sure it would work using the normal install process. I also don't know how publishing packages works; do any changes need to be made to the release script to ensure the patch file(s) are part of the package? Can we include anything other than the `dist` directory in the release so we don't need to copy the patch files there?


Dhalton (approved):
This looks good to me. As far as releasing goes, I don't think we should need to make any changes to the release script. It just calls `lerna publish` which is going to look at the `dist` folder. AFAIK you can't tell it to look at multiple folders like ` dist` _and_ `patches`, but I think this should work. I think we won't know for sure until we publish this and try to pull it in

modules/cactus-form/patches/react-final-form+6.5.7.patch
```diff
@@ -0,0 +1,108 @@
+diff --git a/node_modules/react-final-form/dist/react-final-form.cjs.js b/node_modules/react-final-form/dist/react-final-form.cjs.js
```
daniloPenaR, line 1: is this an auto-generated file?

wilysword: Yes, it's generated when you run `yarn patch-package` without any other arguments.

daniloPenaR (approved):

PR #702, opened 2022-01-12
CACTUS-726 Input Status Styles

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-726

As is becoming a habit, individual commit messages may have a few extra details.

Incidentally, although I'm relatively sure that DateInput and FileInput don't need `id`/`name` to be required any more than ColorPicker, they're significantly more complex and I didn't want to make that change and then be proven wrong (nor did I want to comb through the code to verify a change that has nothing to do with this ticket).

### DIFF
--- a/modules/cactus-web/src/helpers/status.ts
+++ b/modules/cactus-web/src/helpers/status.ts
@@ -1,22 +1,23 @@
+import { CactusTheme, color } from '@repay/cactus-theme'
 import PropTypes from 'prop-types'
-import { css } from 'styled-components'
+import { CSSObject } from 'styled-components'
 
-export type Status = 'success' | 'warning' | 'error'
-export const StatusPropType = PropTypes.oneOf<Status>(['success', 'warning', 'error'])
+const statuses = ['success', 'warning', 'error'] as const
+export type Status = typeof statuses[number]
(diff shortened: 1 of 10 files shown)


modules/cactus-web/src/helpers/status.ts
```diff
@@ -1,22 +1,23 @@
...
+const statuses = ['success', 'warning', 'error'] as const
+export type Status = typeof statuses[number]
+type StatusBackground = 'successLight' | 'warningLight' | 'errorLight'
+export const StatusPropType = PropTypes.oneOf<Status>(statuses)
 
-type StatusMap = { [K in Status]: ReturnType<typeof css> }
+interface StatusProps {
+  theme: CactusTheme
+  status?: Status | null
+  disabled?: boolean
```
daniloPenaR, line 13: I understand the using of `status` and `disabled` props... But I don't see where you are using `theme` prop in the `getStatusStyles` function... is there mandatory? or how does it works?

wilysword: `theme` must be part of props where we call `color(p, p.status)` down below. Remember that the `color` helper function is equivalent to `p.theme.colors[p.status]`.

daniloPenaR (approved):

Dhalton (approved):
LGTM

PR #707, opened 2022-01-19
CACTUS-816 :: borderSize types

Dhalton:
https://repayonline.atlassian.net/browse/CACTUS-816

To be honest, I don't know if the `(p: ThemeProps) => Interpolation<ThemeProps>` is right. The ticket says it'd be complex to include that, and obviously what I'm doing isn't complex, so if it's incorrect let me know.

### DIFF
--- a/modules/cactus-theme/src/helpers/borders.ts
+++ b/modules/cactus-theme/src/helpers/borders.ts
@@ -1,15 +1,12 @@
-import { DefaultTheme, Interpolation, ThemedCssFunction } from 'styled-components'
+import { CSSObject } from 'styled-components'
 
 import { BorderSize, CactusColor, Shape } from '../theme'
 import { memo, ThemeProps, wrap } from './base'
 
-type BorderSizeOpts = {
-  [K in BorderSize]:
-    | string
-    | ThemedCssFunction<DefaultTheme>
-    | ((p: ThemeProps) => Interpolation<ThemeProps>)
+type BorderSizeOpts<T> = {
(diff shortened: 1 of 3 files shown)


wilysword (changes requested):
`Interpolation` _is_ simpler than I thought it was, but the problem with `(p: ThemeProps)` is that it limits what you can do. Imagine you have a component like this:
```
const MyComp = styled(BaseComp)<{ cool: number }>`
  some-property: ${borderSize({ thin: (p) => p.cool + 'px', thick: (p) => p.cool * 2 + 'px' })};
`
```
That won't work because your interpolation signature has "lost" the rest of the component's props.

As such, I'm going to go ahead and say I was wrong to even want that in the type signature. Especially because it'd be easy to write the above like this instead:
```
const MyComp = styled(BaseComp)<{ cool: number }>`
  some-property: ${(p) => borderSize(p, { thin: p.cool + 'px', thick: p.cool * 2 + 'px' })};
`
```
I also noticed just now that it either needs to only support strings, or you need two variants, because in the context of the `border` helper it can't support interpolations at all since that one returns a string. So maybe you could have signatures like
```
import { CSSObject } from 'styled-components'
borderSize(opts: BorderSizeOpts<string | CSSObject>)
border(color: string, opts: Partial<BorderSizeOpts<string>>)
```
Sorry I gave bad advice before.

wilysword (approved):

daniloPenaR (approved):

PR #708, opened 2022-01-24
CACTUS-821 :: Saturation Multiplier 

daniloPenaR:
[CACTUS-821](https://repayonline.atlassian.net/browse/CACTUS-821)

This is my current approach on this:
I created the `saturationMultiplier` variable and added to the theme object. Set 1 as default and multiplied that factor with every saturation value of the `hsl` representations of the following colors: 
 
- Base
- CTA
- CTA light
- Contrast light
- Contrast Medium
- Contrast Dark
- Contrast Darkest

I added one range control for that Saturation Multiplier into the `Cactus Theme` addon in the Storybook app of the `cactus-web` module. So you can run the storybook app and test that. 

Don't forget to add the current version of the `theme` package using `yalc`.

### DIFF
--- a/docs/Theme/README.md
+++ b/docs/Theme/README.md
@@ -53,14 +53,15 @@ If only a `primary` color is provided or the secondary color is white, the theme
 
 Whether you generate a theme using a primary hue or with primary and secondary hex values, there are several other optional keys you can set in the options object:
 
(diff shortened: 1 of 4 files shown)


wilysword (approved):

Dhalton (approved):
LGTM, I'll merge after you update the logic Glen mentioned

PR #714, opened 2022-02-02
CACTUS-797 :: Docs Versioning POC

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-797

So this is where this is at right now:

- Only pulling in docs from cactus-web v8 at the moment
- The v8 docs are not showing up in the sidebar menu just yet -- but you can still view them by adding `cactus-web/v8` in the url before the `components/{{componentName}}` part. For example, to see the v8 docs for the `Text` component, you'd go to this url: `https://localhost:8001/cactus-web/v8/components/text/`
- The interactive examples are pulling in the newest versions of these components which is not ideal.

Where do we go from here?
- We'll need to figure out what we want to do about the interactive examples. From what I've seen, it doesn't look like we can swap out the imports on the fly, but I could be wrong. Another option may be to just get rid of the interactive examples for any old docs. Even that could be tricky though.
- For a finished product, I'd like to modify the sidebar to allow us to have deeper nesting levels. For instance, I think it'd work best if users could select "Cactus Web" on the sidebar, then have a child dropdown menu where they could select which version they want to see docs for.
- Eventually we'll need to start pulling older versions. Should be easy to set that up though.

Let me know what you guys think

### DIFF
--- a/website/gatsby-config.js
+++ b/website/gatsby-config.js
@@ -37,6 +37,15 @@ module.exports = {
         ignore: ['**/*.ts', '**/*.tsx', '**/*.json', '**/*.snap'],
       },
     },
+    {
+      resolve: 'gatsby-source-git',
+      options: {
+        name: 'cactus-web/v8',
+        remote: 'https://github.com/repaygithub/cactus.git',
+        branch: '@repay/cactus-web@8.1.0',
+        patterns: 'modules/cactus-web/src/**/*.mdx',
+      },
+    },
     'gatsby-transformer-sharp',
     'gatsby-plugin-sharp',
     {
--- a/website/gatsby-node.js
+++ b/website/gatsby-node.js
@@ -99,8 +99,27 @@ function getTitle({ node, getNode }) {
 exports.onCreateNode = ({ node, actions, getNode }) => {
   const { createNodeField } = actions
 
-  if (node.internal.type === 'Mdx') {
-    const isComponent = getNode(node.parent).sourceInstanceName === 'components'
+  const parentNode = getNode(node.parent)
+  const sourceInstanceName = parentNode?.sourceInstanceName
+
+  if (sourceInstanceName && sourceInstanceName.match(/cactus-web\/v[0-9]+/g)) {
+    createNodeField({
+      name: 'slug',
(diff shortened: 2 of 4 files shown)


wilysword (commented):
Great work.

As far as how to select different docs version, I like the way a lot of Python projects do it, with a little floating menu (in the bottom right): https://docs.celeryproject.org/en/stable/. I guess it's trickier in this case because we have multiple libraries, not just one.

For the live examples, not sure what to say. Maybe it'd be worth it to look into other JS doc generating libraries, given that this isn't the first time we've had trouble with gatsby (especially performance). If there were some way to get it to compile the _entire_ site in advance, including the cactus-web code that it imports locally, that might fix the issue.

Dhalton:
> As far as how to select different docs version, I like the way a lot of Python projects do it, with a little floating menu (in the bottom right): https://docs.celeryproject.org/en/stable/. I guess it's trickier in this case because we have multiple libraries, not just one.

Yeah I like that way of doing it and I'm down to shoot for that, but it would definitely be tricky. Not only for the reason you mentioned, but also because as a user I'd expect that if I choose a version and navigate to another page, that same version I chose would still be selected. That's the part that I'm more concerned about in terms of what's possible with Gatsby.

> For the live examples, not sure what to say. Maybe it'd be worth it to look into other JS doc generating libraries, given that this isn't the first time we've had trouble with gatsby (especially performance). If there were some way to get it to compile the _entire_ site in advance, including the cactus-web code that it imports locally, that might fix the issue.

I'm definitely not opposed to looking into other libraries. I'm not a big fan of Gatsby and maybe there's something else out there with better support for something like this.

daniloPenaR (commented):
Following Glen's opinion. It seems that there are many ways to handle the versioning from the UI perspective and the trickier part is to handle the multiple libraries we support into that Gatsby app...
I like the idea of starting to look for an alternative to Gatsby in order to solve how that versioning must be handled (I like that idea of a floating menu. Or a menu fixed next to the component's or library's main Title, like in the Jest docs. 
Screen Shot 2022-02-03 at 11 48 44 AM

PR #717, opened 2022-02-10
CACTUS-832: add Range & RangeField components

wilysword:
[CACTUS-832](https://repayonline.atlassian.net/browse/CACTUS-832)

I didn't even notice until right before I made the PR, but the style guide actually has the right half of the bar as "light gray" instead of "light call to action" in some of the panels. I didn't implement it that way for a few reasons:
- It doesn't say what color the left part of the bar should be in that case.
- Seems like a pretty drastic color change for a hover style.
- I think it looks good the way it is.

If anyone disagrees, though, I wouldn't mind revisiting. I'd probably make the colors a focus style rather than a hover style, though.

The only other part where I strayed from the style guide is the addition of a `:focus-visible` style. It's not the most attractive thing, but then neither are most browsers' built-in `:focus-visible` styles. It does its job without interfering with other styles, which I think is the most important thing there.


Dhalton (approved):
Code looks good, works well. I like the styles the way they are. Looks clean

daniloPenaR (approved):
Amazing.

PR #721, opened 2022-02-15
CACTUS-850 Sonarqube

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-850

I didn't really see anything to do in `ui-tools` or `ui-private`, so this is it for the ticket. I started before I noticed the report was out of date, so the first two are things Danilo already fixed (or ignored), but I would rather have fixed in a different way.

I didn't bother with all the code duplication, just the two biggest contributors.


Dhalton (approved):
LGTM

daniloPenaR (approved):

PR #723, opened 2022-02-18
CACTUS-738: change Tag close icon to button

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-738

I kind of recycled the `closeOption` prop for setting it to not be a button, since the name seemed to fit, and I could add an option to it while still being backwards compatible. Other changes:

- Fixed some prop types, including removing a completely redundant one from `IconButton`.
- Removed the `data-role="close"` prop, which seems to have been basically a hack to make implementing the `Select` component easier.
- In its place, added the `aria-controls` prop; fulfills basically the same purpose in terms of CSS queries, but also gives an easier reference to the tag, making it easier to share a single handler for multiple tags.

IE sucks as usual.


Dhalton (approved):
LGTM

daniloPenaR (approved):
👍

PR #726, opened 2022-03-02
CACTUS-783 :: Icon Alignment

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-783

Like I said in standup, I didn't find any regressions to fix after updating the icons. If you find any or already know of any off the top of your head, fill me in and I'll get them updated.

Looking at the storyshot diffs is probably the best way to review this; there are a lot of changes to the svg files that are more or less pointless to review in the diff.

I also fixed a little issue with the side bar in the docs website. There's been some whitespace there forever that's always bugged me, but it was a super easy fix.

### Testing
- Run the `cactus-icons` storybook and check out all of the icons. Make sure nothing sticks out.
- Run the `cactus-web` storybook and check out some components that use icons. Some good ones to check would probably be Accordion with Outline, DateInput, Select, SplitButton, MenuButton, ColorPicker, IconButton and so on.


wilysword (approved):
Seems okay. I know you didn't make the SVGs, but the only thing I can say is that the "i" in the `NotificationInfo` icon looks a lot stranger than the old one.

Dhalton:
>  the "i" in the `NotificationInfo` icon looks a lot stranger than the old one.

See I thought the same thing. I might as well just double check with Hector to make sure that's the right icon.

daniloPenaR (approved):
Looks good to me, let me know if it needed to be re-reviewed after double-checking with Hector.

Dhalton:
@wilysword @daniloPenaR Turns out there was a problem with the svg export so thanks for calling that out, Glen. It's been updated. Probably doesn't warrant a full re-review but if you guys want to have another look, go for it. Lmk if you're good with me merging

PR #728, opened 2022-03-03
Test Cleanup

Dhalton:
Had some extra time and I've been thinking about doing this for a while so I figured I'd put up a rough draft and see what you guys think.

Basically I want to replace all of the `render` usage in our tests with this `renderWithTheme` function I wrote. I just started with `DateInput` tests since that's the test file I had open. It's pretty simple, but it removes a lot of repetitive lines that aren't specific to any tests. I figure if we do this, we could also just do some general cleanup where it makes sense, too. For instance, here I got rid of all of the unnecessary return value types and imported `animationRender` instead of re-defining it in this file.

So do you guys think this is worth our time? Do you think there are any use cases I may have missed? We're light on work right now so I think if we're gonna do it now's the time.

Oh yeah, and I'd recommend looking at the diff without whitespace changes.

### DIFF
--- a/modules/cactus-web/src/DateInput/DateInput.test.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.test.tsx
@@ -1,30 +1,20 @@
-import { generateTheme } from '@repay/cactus-theme'
-import { act, fireEvent, render } from '@testing-library/react'
+import { act, fireEvent } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import * as React from 'react'
 
+import animationRender from '../../tests/helpers/animationRender'
+import renderWithTheme from '../../tests/helpers/renderWithTheme'
 import { PartialDate } from '../helpers/dates'
 import KeyCodes from '../helpers/keyCodes'
 import Modal from '../Modal/Modal'
-import { StyleProvider } from '../StyleProvider/StyleProvider'
 import DateInput from './DateInput'
 
-function animationRender(): Promise<void> {
-  return new Promise((resolve): void => {
-    setTimeout((): void => {
-      window.requestAnimationFrame((): void => resolve())
-    }, 0)
-  })
-}
-
-describe('component: DateInput', (): void => {
-  describe('can be controlled', (): void => {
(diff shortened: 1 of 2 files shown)


wilysword (approved):
I like it well enough. I'm somewhat torn whether we should make ticket(s) specifically for this, or just make a "clean as you go" rule like we did with linting channels at first.

Dhalton:
My idea was to make ticket(s) for it, but I can see an argument for cleaning things up as we go. Ultimately this is just to make our lives ever-so-slightly easier, mainly when writing new tests, so as long as we just make sure to refactor any time we touch a test file, I'd be fine with that. Then maybe down the line when there aren't many files left to clean up we can make a ticket with a lesser scope.

daniloPenaR (approved):
Nice, I like the "clean as you go" strategy. 
Another strategy is to create the first ticket with a group of components (like all the `Field` components, for example) and implement the `renderWithTheme` helper there and see what happens. 
Just another option, just thinking which one would be more convenient to us.

Dhalton:
Since you guys are on board with the concept, I'll resolve some conflicts and turn this into a PR that I can merge so we'll have the helper there.

After thinking about it some more, I kind of want to argue to make a ticket and maybe just time box it to 3 days and see how much we can get done in that time span, then do the same for subsequent tickets if necessary. My reasoning is that I'd rather dedicate time to it up front and knock it out quickly than live with a code base that's inconsistent like that for who knows how long. Also, with the linting in Channels, you didn't really have to dedicate any time to it outside of your normal workflow. All you had to do was run the linter and auto-fix, but here we'd have to dedicate actual time to refactoring, causing me to lean more toward making ticket(s). What do you guys think?

wilysword:
That's fair. This change is also both smaller and more regular than the linting changes tended to be.

Dhalton:
@daniloPenaR @wilysword I'm going to merge this today so that the helper is in master. I'm also going to make a ticket that we can go over for grooming tomorrow. If you guys are _really_ against the ticket idea then we can talk about it more then

PR #731, opened 2022-03-31
CACTUS-872: Test Cleanup

daniloPenaR:
[CACTUS-872](https://repayonline.atlassian.net/browse/CACTUS-872)

Nothing too much to add. 
There's a _special_ case at `Box` component.

### Testing
Run `yarn web test`, every test of every component should pass.

### DIFF
--- a/modules/cactus-web/src/Box/Box.test.tsx
+++ b/modules/cactus-web/src/Box/Box.test.tsx
@@ -2,37 +2,36 @@ import { generateTheme } from '@repay/cactus-theme'
 import { render } from '@testing-library/react'
 import * as React from 'react'
 
+import renderWithTheme from '../../tests/helpers/renderWithTheme'
 import { StyleProvider } from '../StyleProvider/StyleProvider'
 import Box from './Box'
 
-describe('component: Box', (): void => {
-  test('should accept built-in props', (): void => {
-    const { getByText } = render(
-      <StyleProvider>
-        <Box
-          position="relative"
-          display="block"
-          top="-1"
-          right="auto"
-          bottom="auto"
-          left="-1"
-          margin={2}
-          padding={2}
-          width="120px"
-          height="120px"
-          backgroundColor="darkestContrast"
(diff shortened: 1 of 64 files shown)


modules/cactus-web/src/Box/Box.test.tsx
```diff
@@ -42,12 +41,11 @@ describe('component: Box', (): void => {
     expect(style.borderRadius).toBe('20px')
   })
 
+  // THos should Be solved first !!!!
```
Dhalton, line 44: What?

wilysword (commented):
This must have been pretty boring; I applaud you for getting through it.

modules/cactus-web/src/Select/Select.test.tsx
```diff
@@ -22,18 +21,16 @@ describe('component: Select', () => {
...
     )
 
     expect(getByText('Who?')).not.toBeNull()
   })
 
   test('can receive options as children, with altText', () => {
-    const { queryByText } = render(
-      <StyleProvider>
+    const { queryByText } = renderWithTheme(
+      <>
```
Dhalton, line 33: I don't think this fragment is needed

Dhalton (approved):
I'm not sure what that comment means and I think there's a `Fragment` in there that we don't need, but overall it looks good.

It looks like this probably takes care of all of the tests, is that correct?

Dhalton:
> Also, here in that test, there's a `rerender` function that is redefining the theme object, a special case that is not present in the other tests

Yeah, the `renderWithTheme` function kind of assumes that you'd rerender with the same theme as the initial render. If you want, I bet it'd be possible to wrap the `rerender` function in another function that accepts an optional theme config argument. Then if that extra argument exists, we could generate another theme, otherwise use the first one. But this seems to be a one-off case, and I'm not super fussed about it, so I'm fine with leaving it. My approval still stands.

wilysword (approved):

PR #732, opened 2022-03-31
CACTUS-868 :: Docs Versioning

Dhalton:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-868

I'm still not sure why, but the `DataGrid` page (for current docs) was crapping out because `component` was undefined in `PropsTable`. So to get around that, I imported the component directly. I also added the `PropsTable` for `BottomSection` because it was missing before.


daniloPenaR (approved):

wilysword (approved):
I didn't check everything, obviously, but most of it seemed to work. `DataGrid` wasn't the only one missing props, though; `AccessibleField` was on older versions, as well. And on versions 7 & 8, `DataGrid` had a fatal error when I tried to load the page. Seems good enough as a POC, though, I'm still a bit wary of gatsby in general but it'd probably be a lot of work to switch, especially since we can't really update the old versions to match.

Dhalton:
> And on versions 7 & 8, `DataGrid` had a fatal error when I tried to load the page.

It's so weird, I'm like 95% sure I tested that page on those versions and didn't have an issue, but sure enough, today I try it and I get the error. What do you think about the change I just pushed up? It prevents the entire page from crashing in the event that `component` is `undefined` in `PropsTable`. At least that way, users would still be able to see docs we write explicitly.

Dhalton:
@wilysword Are you good with the change I made to `PropsTable`? It prevents the page from crashing entirely, at least.

wilysword:
Still ✅. (Weird they don't have an :approval: emoji or something that's the same as their approval symbol.)

PR #734, opened 2022-04-01
CACTUS-883: add NotificationProvider and accompanying hook

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-883

I put the code in with Notification, partly because I'm lazy but mostly because I doubt anyone will ever directly use the Notification component again after this is released, so I wanted it on the same docs page.

Not mentioned on the ticket, but added here, is the `clearNotification` function.


modules/cactus-web/src/Notification/Provider.tsx
```diff
@@ -0,0 +1,145 @@
...
+export interface ElementArgs extends NotificationArgs {
+  element: React.ReactElement
+}
+
+const hasElement = (args: any): args is ElementArgs => !!args.element
+
+interface SetNotification {
+  (note: AlertArgs): React.Key
+  (note: ElementArgs): React.Key
+}
```
daniloPenaR, line 25: Is that a kind of _hybrid_ type?

daniloPenaR (approved):
Tested it and works perfectly. It'd be nice to implement it inside of the `create-repay-ui` templates. Anyway, I left some questions and my approval.

modules/cactus-web/src/Notification/Provider.tsx
wilysword: Same as the other comment, this is overloading. But when using arrow functions, you can't do the multiple definitions thing that you do with the `function` keyword, so instead you have to make an interface with multiple call signatures. It's functionally the same as what you see in the test you commented on, though.

modules/cactus-web/src/Notification/Provider.tsx
```diff
@@ -0,0 +1,145 @@
...
+      }
+      element = React.createElement(Alert, alertProps, message)
+    }
+    this.dispatch({ element, key, vertical, horizontal })
+    return key
+  }
+}
+
+// Separate component so that changes to the notifications don't rerender the entire tree.
+const NotificationRenderer = ({ store }: { store: NotificationStore }) => {
```
Dhalton, line 111: Clever to separate the rendering from the provider itself. I wouldn't have thought to do that.

Dhalton (commented):
Looks awesome. Do you think the alerts in each notification should be rendered with some top/bottom margin to space them out a bit though?

wilysword:
> Do you think the alerts in each notification should be rendered with some top/bottom margin to space them out a bit though?

Personally I think it looks fine the way it is, and if the dev wants space they can always add it when setting the notification.

Dhalton (approved):

PR #738, opened 2022-04-05
CACTUS-880 `@repay/scripts` upgrade

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-880

I found that regenerating the lockfile, and even upgrading `@repay/scripts` without touching anything else, caused the same hanging tests issues that Danilo saw in the recent upgrade ticket. I assumed that there would be only one or two libraries actually at fault for that so I ran `yarn-deduplicate` to see what changed, then split that list into chunks and de-duped them one at a time (that's the first commit, originally several commits which I squashed when I was done). Eventually I found that it was the switch to `core-js` that caused the problem, and that anything v3.16 or higher didn't work. `resolutions` in the `package.json` is supposed to be able to specify a particular parent package for the pinned dependency, but I couldn't get that to work so I just pinned it globally.

With the recent rollup config fix, I also removed the config overrides used by several of the modules. I left the one in `theme-components` because I have no idea what it's doing or why.


daniloPenaR (approved):
This leads me to a few questions:
- Why are we using rollup and webpack (specially rollup)?
- Can you give me more details on the process you followed doin' this:
> then split that list into chunks and de-duped them one at a time (that's the first commit, originally several commits which I squashed when I was done)

I mean, you took the entire lockfile and start looking by parts? 
(maybe this could be useful to me for the next time I need to deal with that kind of tickets 😅 )

Anyway, seems good to me, here's my approval.

wilysword:
> Why are we using rollup and webpack (specially rollup)?

Don't really know about rollup; I just know that we recently fixed a config discrepancy that meant we could remove the related config overrides. I think webpack is kind of like a compiler, it takes your code and processes it into a bundle that can be sent to browsers to be executed.

> Can you give me more details on the process you followed doin' this

1. Ran `yarn deduplicate`, then saved the results of `git diff` to a file.
2. Used vim magic (I'm sure other editors can do it too) to prune the diff file till I had just the list of packages affected by the dedup; one of which should be causing the tests to hang.
3. Used `yarn deduplicate --packages` to deduplicate a few packages at a time, following this pattern (I split the list of packages into several files based on first letter):
    1. ``yarn deduplicate --packages `cat a-c` ``
    2. ``rm -rf `find -name node_modules` && yarn install``
    3. `yarn w standard test:ci`
    4. If the tests run, I would commit the changes and move on to the next chunk of packages.
    5. If the tests hang, I would further split the current chunk until I'd narrowed it down to just one that was causing the issue.

That ended up being `core-js`, or more specifically, the version of `core-js` used by `@repay/scripts`.

PR #740, opened 2022-04-06
CACTUS-753: fix styled-system `propTypes`

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-753

`Flex`, `Spinner`, and `TextArea` were the only ones I found that currently had this issue. There were quite a few places, though, where we _could_ be using `styled-system` to implement responsive styles, but are not; I'll probably go through them one of these days and make some tickets.

`justifyContent` is a special case, unfortunately, as long as we have to support IE. I did try to make a responsive version of the fix, but unlike regular styles, the `::before` and `::after` pseudo-elements don't just disappear as they're overridden.

I added a couple of other minor Typescript fixes and removed some propTypes for the `as` prop, which I think were probably incorrect and almost certainly pointless.

### DIFF
--- a/modules/cactus-web/src/Alert/Alert.tsx
+++ b/modules/cactus-web/src/Alert/Alert.tsx
@@ -147,10 +147,9 @@ export const Alert = styled(AlertBase).withConfig(
   }
 `
 
-// @ts-ignore
 Alert.propTypes = {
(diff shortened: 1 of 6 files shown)


Dhalton (approved):
LGTM

PR #741, opened 2022-04-08
CACTUS-885 :: Upgrade to Yarn Berry

Dhalton:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-885

The more notable changes are changes in syntax:
- No more artbitrary `pre` and `post` scripts -- `preinstall` and `posinstall` still work but any others must be explicitly called now.
- `--frozen-lockfile` has been changed to `--immutable`
- `yarn workspaces run` has been changed to `yarn workspaces foreach -i -v --exclude cactus run`.
  - `-i` (interlaced) prints workspace output at run time instead of truncating it.
  - `-v` (verbose) prints workspace names next to output.
  - `--exclude cactus` is needed because it runs the command for the _root_ workspace, too, which causes an infinite loop in cases where both the root workspace and child workspaces contain a script with the same key, so we skip the root workspace.

One other quirk was that, when running in the CI pipeline, workspaces didn't have access to `jest`, so I had to add it to the root. I had a github thread related to that, but I can't seem to find it anymore....


wilysword (approved):

Dhalton:
I just noticed that Testcafe is only running some of the integration tests and skipping a lot of them, so I'll need to figure that out before we merge this.

PR #744, opened 2022-04-18
CACTUS-887 :: Fix Tooltip Scrolling/Remove @reach/tooltip

Dhalton:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-887

The `usePopup` usage is a little weird since most of the props that hook returns aren't useful for the tooltip. Instead of adding a case to the `usePopup` function to account for tooltips, I just took the props it returns and used only the ones I needed. If you guys want me to add a tooltip case to `usePopup`, let me know and I'll refactor.

~The default position function is wrapped in another function that takes the theme as a prop. I had to do this because I found that border radius overrides were persisting when scrolling and causing collisions in different places, so we have to set the base border radius each time the position function is called. We need the theme to get the border radius values, though, so that's where the wrapper comes in.~ This was removed

You may be wondering why I changed the padding to margins in `AccessibleField`. This was because the padding was throwing thing positioning off, but when using margins instead of padding, the coordinates included the extra space, fixing that problem.

### DIFF
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
@@ -89,7 +89,7 @@ describe('component: AccessibleField', () => {
 
     const tooltipText = getByText('JSX Tooltip')
     expect(tooltipText.tagName).toBe('STRONG')
-    expect(tooltipText.parentElement).toHaveAttribute('id', 'aftest-tip-popup')
+    expect(tooltipText.parentElement).toHaveAttribute('id', 'aftest-tip')
     expect(tooltipText.parentElement).toHaveAttribute('role', 'tooltip')
 
     const errorText = getByText('Error Paragraph')
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
@@ -32,14 +32,14 @@ describe('component: AccessibleField', () => {
   })
 
   test('provides an accessible tooltip', () => {
-    const { getByLabelText, getAllByText } = renderWithTheme(
+    const { getByLabelText, getByText } = renderWithTheme(
       <AccessibleField label="Accessible Label" name="text_field" tooltip="woot tooltips!">
         <input name="text_field" data-is="accessible" />
       </AccessibleField>
     )
 
     expect(getByLabelText('Accessible Label').getAttribute('aria-describedby')).toBe(
-      getAllByText('woot tooltips!')[1].closest('[id]')?.id
+      getByText('woot tooltips!').closest('[id]')?.id
     )
   })
 
@@ -66,7 +66,7 @@ describe('component: AccessibleField', () => {
   })
 
   test('alternate prop types', () => {
-    const { container, getByText, getAllByText } = renderWithTheme(
+    const { container, getByText } = renderWithTheme(
       <AccessibleField
         id="aftest"
         label={<em>Accessible Label</em>}
@@ -87,9 +87,9 @@ describe('component: AccessibleField', () => {
     const input = container.querySelector(`#${label?.getAttribute('for')}`)
     expect(input).toHaveAttribute('name', 'text_field')
 
-    const tooltipText = getAllByText('JSX Tooltip')[1]
+    const tooltipText = getByText('JSX Tooltip')
     expect(tooltipText.tagName).toBe('STRONG')
-    expect(tooltipText.parentElement).toHaveAttribute('id', 'aftest-tip')
+    expect(tooltipText.parentElement).toHaveAttribute('id', 'aftest-tip-popup')
     expect(tooltipText.parentElement).toHaveAttribute('role', 'tooltip')
 
     const errorText = getByText('Error Paragraph')
--- a/modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
+++ b/modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
@@ -32,14 +32,14 @@ describe('component: AccessibleField', () => {
   })
 
   test('provides an accessible tooltip', () => {
-    const { getByLabelText, getByText } = renderWithTheme(
+    const { getByLabelText, getAllByText } = renderWithTheme(
       <AccessibleField label="Accessible Label" name="text_field" tooltip="woot tooltips!">
         <input name="text_field" data-is="accessible" />
       </AccessibleField>
     )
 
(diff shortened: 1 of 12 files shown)


modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
```diff
@@ -32,14 +32,14 @@ describe('component: AccessibleField', () => {
...
-    const { getByLabelText, getByText } = renderWithTheme(
+    const { getByLabelText, getAllByText } = renderWithTheme(
       <AccessibleField label="Accessible Label" name="text_field" tooltip="woot tooltips!">
         <input name="text_field" data-is="accessible" />
       </AccessibleField>
     )
 
     expect(getByLabelText('Accessible Label').getAttribute('aria-describedby')).toBe(
-      getByText('woot tooltips!').closest('[id]')?.id
+      getAllByText('woot tooltips!')[1].closest('[id]')?.id
```
wilysword, line 42: This makes me think there's something off with your implementation. You really only need one copy of the tooltip; I know the default `usePopup` implementation uses `aria-hidden`, but in this case I'd suggest it should always be visible to screen readers, and the toggle should switch between visually hidden and fully visible. (I see what you mean now about `usePopup` not applying very well to this situation.)

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -26,57 +27,86 @@ export interface TooltipHandle {
...
-    return {}
-  }
-  const scrollX = getScrollX()
-  const scrollY = getScrollY()
-  const styles: ReturnType<Position> = {
-    left: triggerRect.left + scrollX,
-    top: triggerRect.top - tooltipRect.height + scrollY,
-  }
+const cactusPosition: (theme: CactusTheme) => PositionCallback = (theme) => {
+  const initialBR = radius(8)({ theme })
```
wilysword, line 42: Not sure I understand why this is here. Can't you just put this in the component styles, and then only override the corner you need? (Probably would also need to set the three you're not overriding, e.g. `tt.style.borderBottomLeftRadius = ''`.)

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -115,68 +145,97 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
...
-            isVisible={isVisible}
-            label={label}
-            ariaLabel={ariaLabel}
-            position={position || defaultPosition}
+            id={popupProps.id}
+            ref={tooltipRef}
+            aria-label={ariaLabel}
+            aria-hidden={!visible}
             style={{ maxWidth }}
+            data-tooltip-popup="true"
```
wilysword, line 207: Do we really need this? If it's just for selectors and you take my advice with toggling visually hidden, you could use `[role="tooltip"]` instead.

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -115,68 +145,97 @@ const TooltipBase = (props: TooltipProps): React.ReactElement => {
...
+  const delayHovering = (value: boolean) => {
+    setTimeout(() => {
+      if (!hoveringPopup.current) toggle(value)
+    }, 500)
   }
 
-  const isVisible = stayOpen || hovering || forceVisible || tooltip.isVisible
   return (
-    <>
+    <div onMouseEnter={() => delayHovering(true)} onMouseLeave={() => delayHovering(false)}>
```
wilysword, line 198: For some reason this div bothers me. Since the popup is `fixed` and won't affect it's container, what do you think about putting it inside the `span` instead? (It's a pretty vague feeling, so if you don't like my suggestion or it doesn't work, don't worry about it too much.)

wilysword (commented):
Functionality seems solid, just a few things I think we could improve over the Reach implementation.

modules/cactus-web/src/AccessibleField/AccessibleField.test.tsx
Dhalton: Yeah, I guess with reach we had the invisible text for screen readers since reach removed it from the DOM, but I guess you're right; here we could just use the tooltip popup for accessibility since it always exists, but just isn't always visible. Thanks

modules/cactus-web/src/Tooltip/Tooltip.tsx
Dhalton: Hmm, I thought I'd have to set the border radii to the intended pixel value but you might be right, I may just be able to remove the override that was there previously. I'll give it a try, thanks

modules/cactus-web/src/Tooltip/Tooltip.tsx
Dhalton: Will do

modules/cactus-web/src/Tooltip/Tooltip.tsx
Dhalton: I'll give it a shot and if I run into too much trouble I'll just keep it like this

Dhalton:
@wilysword This is updated. The one quirk that I to get around was with the id for the tooltip. It needs to use the id returned from `usePopup`, otherwise it's not positioned correctly the first time it's displayed, but we also need fields that have a tooltip to know what that id is in order for them to be accessible. I just adjusted `AccessibleField` so that it appends the `-popup` to the tooltip id, since that's what `usePopup` does. I don't particularly love that, so if you have any ideas for how I could do that better, let me know.

wilysword:
`usePopup` accepts a `popupId` argument, can't you just use that to set the ID to whatever you need it to be?

Dhalton:
> `usePopup` accepts a `popupId` argument, can't you just use that to set the ID to whatever you need it to be?

Yep, thanks for letting me know

daniloPenaR (approved):

modules/cactus-web/src/Tooltip/Tooltip.tsx
```diff
@@ -246,11 +258,7 @@ export const TooltipPopup = styled.div`
   border: ${border('callToAction')};
   border-radius: ${radius(8)};
 
-  display: block;
-  &[aria-hidden='true'] {
-    display: none;
-  }
-
+  display: ${(p) => (p.$visible ? 'block' : 'none')};
```
wilysword, line 261: As far as I know, `display: none` hides from screen readers just like `aria-hidden` does. You'll need a different way of hiding it, like `{ scale: '0', opacity: '0' }`.

Dhalton: Oh okay I didn't know that, I'll update it

wilysword (approved):

PR #746, opened 2022-04-21
CACTUS-900 :: Configurable Error Messages

Dhalton:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-900

### DIFF
--- a/docs/Forms/API Documentation.md
+++ b/docs/Forms/API Documentation.md
@@ -126,7 +126,7 @@ You can also define custom required error messages with the `requiredMsg` prop. 
 <Field required requiredMsg={<div>Required Field</div>} {...fieldProps} />
 ```
 
-To use `required` without the validator you'd need to pass an explicit `validate` function, like a noop (or use `configureDefaults`/`withDefaults` to set a default validator).
+To use `required` without the validator you'd need to pass an explicit `validate` function, like a noop (or use `configureDefaults`/`withDefaults` to set a default validator or default `requiredMsg`).
 
 ### Refs & Extensions
 
--- a/docs/Forms/API Documentation.md
+++ b/docs/Forms/API Documentation.md
@@ -116,6 +116,16 @@ This validator is also available separately:
 import { validateRequired } from '@repay/cactus-form'
 ```
 
+You can also define custom required error messages with the `requiredMsg` prop. Both strings and React nodes are acceptable:
+
+```
+<Field required requiredMsg="Required Field" {...fieldProps} />
+```
+
+```
+<Field required requiredMsg={<div>Required Field</div>} {...fieldProps} />
+```
+
 To use `required` without the validator you'd need to pass an explicit `validate` function, like a noop (or use `configureDefaults`/`withDefaults` to set a default validator).
 
 ### Refs & Extensions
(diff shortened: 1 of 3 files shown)


daniloPenaR (approved):
LGTM.

docs/Forms/API Documentation.md
```diff
@@ -116,6 +116,16 @@ This validator is also available separately:
...
+
+```
+<Field required requiredMsg="Required Field" {...fieldProps} />
+```
+
+```
+<Field required requiredMsg={<div>Required Field</div>} {...fieldProps} />
+```
+
 To use `required` without the validator you'd need to pass an explicit `validate` function, like a noop (or use `configureDefaults`/`withDefaults` to set a default validator).
```
wilysword, line 129:
```suggestion
To use `required` without the validator you'd need to pass an explicit `validate` function, like a noop (or use `configureDefaults`/`withDefaults` to set a default validator or default `requiredMsg`).
```

modules/cactus-form/src/Field.ts
```diff
@@ -169,6 +171,15 @@ export default makeConfigurableComponent(Field, {
   processMeta: addError,
 })
 
+const makeRequiredValidator = (requiredMsg: React.ReactNode) => {
+  return (value: any): React.ReactNode | undefined => {
```
wilysword, line 175: I'd suggest making the implementation here be the current `validateRequired` function, and then down below you could have
```
export const validateRequired = makeRequiredValidator('Missing required field')
```

wilysword (commented):
This looks good. I have an alternate suggestion, but only if you think it would be better; I'm suggesting it because it would be _slightly_ more performant, but not enough for that to be a real consideration. I just like thinking of such things.

Anyway, the suggestion is instead of `requiredMsg`, you have a `validateRequired` prop: that way you could make the validator separately and it wouldn't have to be re-created on every render. Because there'd be no point to passing `required validateRequired={x}` when you could just pass `validate={x}`, this would _only_ be used via default props.

Like I said, the performance consideration is practically negligible, and I have no problem with the current implementation, so you can just reject my suggestion and I'll approve the PR.

modules/cactus-form/src/Field.ts
maxrcollins: It looks like Dhalton was trying to have 'Missing required field' be the back up error message in the current implementation. If so I'd say just do this:
```
const makeRequiredValidator = (requiredMsg?: React.ReactNode) => {
  return (value: any): React.ReactNode | undefined => {
    if ((!value && value !== 0 && value !== false) || value?.length === 0) {
      return requiredMsg || 'Missing required field'
    }
  }
}

// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
export const validateRequired = makeRequiredValidator()
```

Also line 153 could then be simple again: `fieldConfig.validate = makeRequiredValidator(requiredMsg)`

Dhalton:
@wilysword Normally I would implement your suggestion, but in this case, having come back from Reactathon with the rules form migration ticket hanging over my head, I'm more concerned about getting that done as soon as possible, so in the interest of time I think I'll just leave this as-is, so long as everyone is OK with that.

wilysword (dismissed):

PR #747, opened 2022-04-25
CACTUS-780:: Pagination Max Width

daniloPenaR:
[CACTUS-780](https://repayonline.atlassian.net/browse/CACTUS-780)

I've written some hooks for calculating the width of the component and based on that width the Component will compute the number of items that will be rendered. 
- `width` and `maxWidth` props are now exposed. 
- Changed `lowerBreak` and `upperBreak` to `leftBreak` and `rightBreak` since I feel those new names are pretty much accurate. 
- The logic has been modified at the moment of getting the items list, there are new variables that calculate the pages that will be shown with the rest of the buttons `['first', 'prev', ...pages, 'next', 'last']` and a new function for rendering those buttons/items. 
- I introduced a new `itemsMaxAmmount`, this prop controls the maximum of items to be rendered by the component.

## Testing: 
Go to the Pagination story, change the size of the preview to tiny values, and check the number of buttons calculated each time. Check the controls for `width` and `maxWidth` props, play around with different values and check that the number of items rendered is changing when you change the width of the component. The minimum amount of items rendered cannot be less than `5`.


modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -146,6 +108,141 @@ const PageLinkBase: React.FC<PageLinkProps> = (props: PageLinkProps): ReactEleme
 
 const noop = () => undefined
 
+const useGetWidth = (ref: React.MutableRefObject<HTMLElement | null>) => {
+  const [width, setWidth] = useState<number>(1)
+  useEffect(() => {
+    const { current } = ref
+    if (current && current.parentElement) {
+      setWidth(current.parentElement.offsetWidth)
+      window.addEventListener('resize', () =>
```
Dhalton, line 117: I think you could define the function you're passing to `addEventListener` separately so that you don't have to re-write the same thing on line 122-123

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -38,6 +38,7 @@ export interface PaginationProps extends MarginProps {
   nextPageLabel?: string
   lastPageLabel?: string
   makeLinkLabel?: (page: number) => string
+  itemsMaxAmmount?: number
```
Dhalton, line 41: "Amount" only has one "m".

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -146,6 +108,141 @@ const PageLinkBase: React.FC<PageLinkProps> = (props: PageLinkProps): ReactEleme
...
+
+  return width
+}
+
+const useGetListLimit = (navigationWidth: number, itemsMaxAmmount = 13) => {
+  const MIN_AMOUNT = 5
+  const [listLimit, setListLimit] = useState(5)
+
+  useEffect(() => {
+    if (Math.floor(navigationWidth / 32) < MIN_AMOUNT) {
```
Dhalton, line 136: What is 32? Is that the width of each item? If so, I think we should define it as a constant with a name so it's easier to understand what this arithmetic is doing.

Dhalton (commented):
Haven't tested anything yet but I have a couple of comments

modules/cactus-web/src/Pagination/Pagination.tsx
daniloPenaR: Yeah, it is, the average width of each item, when the page is > 9, the width increases a bit.

modules/cactus-web/src/Pagination/Pagination.tsx
Dhalton: I was thinking about this a bit more and I think it may be better to just rename this to `maxItems`. I think this name is a little long-winded.

modules/cactus-web/src/Pagination/Pagination.test.tsx
```diff
@@ -142,4 +126,11 @@ describe('component: Pagination', () => {
     }
     expect(func).toHaveBeenCalledTimes(1)
   })
+  test('Pagination: `width` prop', async () => {
+    const { getByLabelText } = renderWithTheme(
+      <ManagedPagination size={12} start={1} width="400px" />
+    )
+
+    expect(getByLabelText('Pages').getAttribute('width')).toBe('400px')
```
wilysword, line 134: I feel like it would make more sense to assert the styles rather than the attribute.

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -158,8 +256,13 @@ export const Pagination: React.FC<PaginationProps> = ({
   currentPageLabel,
   prevPageLabel,
   nextPageLabel,
+  maxItems,
   ...props
 }) => {
+  const navigation = createRef<HTMLElement>()
```
wilysword, line 262: This should be `useRef`, unless you have a reason for wanting a new ref object on every render?

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -181,23 +284,17 @@ export const Pagination: React.FC<PaginationProps> = ({
...
-        </PageButton>
-        <PageButton {...commonProps} page={pageCount} aria-label={lastPageLabel}>
-          <NavigationLast />
-        </PageButton>
+        {itemsList.map((page) => {
+          return itemRenderer(
+            page,
+            commonProps,
+            pageCount,
+            [prevPageLabel, nextPageLabel, lastPageLabel],
```
wilysword, line 294: Using an array here makes it confusing in the body of the function. If you changed it to an (or just passed them as separate args) the references inside `itemRenderer` would be clearer.

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -146,6 +110,140 @@ const PageLinkBase: React.FC<PageLinkProps> = (props: PageLinkProps): ReactEleme
 
 const noop = () => undefined
 
+const useGetWidth = (ref: React.MutableRefObject<HTMLElement | null>) => {
+  const [width, setWidth] = useState<number>(1)
+  useEffect(() => {
+    const { current } = ref
+    const widthCallBack = () =>
+      setWidth((value) => ref.current?.parentElement?.offsetWidth || value)
```
wilysword, line 118: Why are you looking at the width of the parent element instead of the `nav`?

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -146,6 +110,140 @@ const PageLinkBase: React.FC<PageLinkProps> = (props: PageLinkProps): ReactEleme
...
+
+  useEffect(() => {
+    if (Math.floor(navigationWidth / ITEM_WIDTH) < MIN_AMOUNT) {
+      setListLimit(5)
+    } else if (Math.floor(navigationWidth / ITEM_WIDTH) > maxItems) {
+      setListLimit(maxItems)
+    } else {
+      setListLimit(Math.floor(navigationWidth / ITEM_WIDTH))
+    }
+  }, [maxItems, navigationWidth])
```
wilysword, line 144: This really doesn't need a `useEffect`+`useState`, you could just do the calculation inline.

modules/cactus-web/src/Pagination/Pagination.story.tsx
```diff
@@ -28,7 +28,7 @@ export const BasicUsage: PStory = ({ onPageChange, ...args }) => {
   const [current, setCurrent] = React.useState(1)
   return <Pagination {...args} currentPage={current} onPageChange={onPageChange.wrap(setCurrent)} />
 }
-BasicUsage.args = { pageCount: 10 }
+BasicUsage.args = { pageCount: 10, width: 'inherit' }
```
wilysword, line 31: Wouldn't `auto` make more sense as a default than `inherit`? Or even just not have a default at all.

wilysword (dismissed):
When I go into the story and set width to `250px`, the component still renders wider than that. (Technically the `nav` element is 250px, but its children overflow.)

I may have some comments on the page break algorithm, but I'll wait to look more closely when all the other stuff is working correctly.

modules/cactus-web/src/Pagination/Pagination.tsx
daniloPenaR: It was initially conceived to look at the width of the current element, but when Pagination is on DataGrid component it took the minimum width possible and always returns 5 as the minimum amount of pages allowed, rendering only one page.

daniloPenaR: Now I realize that this does not solve the main problem cause it's not taking the nav as you pointed out in the main comment of your review, so my guess Is that I have to change it for the `nav` element again and think any other way to deal with this when the pagination is inside of a `flex` container

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -146,6 +110,140 @@ const PageLinkBase: React.FC<PageLinkProps> = (props: PageLinkProps): ReactEleme
...
+    )
+  }
+  if (page === 'prev') {
+    return (
+      <PageButton
+        {...props}
+        key={page}
+        rel="prev"
+        page={Math.max(1, props.currentPage - 1)}
+        aria-label={navigationLabels[0] as string}
```
maxrcollins, line 210: I don't think you need to cast any of these as strings. They will never *not* be strings because of how they're being set here:
image

modules/cactus-web/src/Pagination/Pagination.tsx
daniloPenaR: you're right, my bad.

modules/cactus-web/src/Pagination/Pagination.tsx
```diff
@@ -181,23 +284,17 @@ export const Pagination: React.FC<PaginationProps> = ({
   lastPageLabel = `${lastPageLabel}, ${pageCount}`
```
maxrcollins, line 284: I know this isn't you but we should never be assigning a new value to a prop variable like this. That's part of the reason you were getting type errors in a few places. Please make these new variables with different names, and then get rid of all the `as string` you have going on in the itemRenderer function.

modules/cactus-web/src/Pagination/Pagination.tsx
daniloPenaR: Did you mean... getting rid of useEffect and useState? and perform those cases in a single calculation?

wilysword: Yes. Normally you use `useEffect` if (1) you need to wait for the DOM to actually render the page (like when you get the element width); or (2) if you need to use a callback that changes another component's state. Neither case applies here.

daniloPenaR: gotcha,  thx

Dhalton (approved):
Code looks good to me, works well too

wilysword:
@daniloPenaR Not sure if you noticed https://github.com/repaygithub/cactus/pull/753

maxrcollins:
Looks good to me

maxrcollins (approved):

PR #749, opened 2022-05-03
Cactus 910 set prop type noOptionsText Select

maxrcollins:


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -108,7 +108,7 @@ function willTruncateBlockShow(
 
 const ValueSwitch = (props: {
   selected: ExtendedOptionType[]
-  placeholder: string | undefined
+  placeholder: React.ReactNode | undefined
```
wilysword, line 111: It's not originally part of the ticket, but this is a good idea: you should also change the type in the `SelectProps` interface and the proptypes. Looks like `matchNotFoundText` is another one that could accept nodes.

In general I think it's best to accept nodes everywhere we can, and only restrict it to strings when the prop is going to be passed as an attribute (i.e. `aria-label`). We haven't really made a concerted effort to update old references though, more just fixing them as they come up.

wilysword (commented):
Is the ticket number on this one wrong? 910 is the README update.

Also, we generally put a link to the ticket in the PR description so it's easier for reviewers to compare the ticket ACs to the implementation.

modules/cactus-web/src/Select/Select.tsx
maxrcollins: Yep. Those are all good calls. I'm going to make a new PR when I update the branch name (whoops... still in a bit of a fever haze), and I'll make sure the interface and proptypes are updated.

PR #750, opened 2022-05-03
cactus-898:: set prop type nooptionstext select

maxrcollins:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-898

I updated the prop type for noOptionsText as well as `placeholder` (They get used in the same place so they should be the same type), and as per a conversation with glen, `matchNotFoundText` because it should be a node too for the same reason as `noOptionsText`.

### DIFF
--- a/modules/cactus-web/src/Select/Select.tsx
+++ b/modules/cactus-web/src/Select/Select.tsx
@@ -111,7 +111,7 @@ function willTruncateBlockShow(
 
 const ValueSwitch = (props: {
   selected: ExtendedOptionType[]
-  placeholder: React.ReactNode | undefined
+  placeholder?: React.ReactNode
   extraLabel: string
   multiple?: boolean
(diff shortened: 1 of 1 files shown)


modules/cactus-web/src/Select/Select.tsx
```diff
@@ -108,7 +111,7 @@ function willTruncateBlockShow(
 
 const ValueSwitch = (props: {
   selected: ExtendedOptionType[]
-  placeholder: string | undefined
+  placeholder: React.ReactNode | undefined
```
maxrcollins, line 114: I almost changed this to `placeholder?: React.ReactNode` to make it uniform with other places, but I couldn't tell if this was done differently for a semantic reason

wilysword (commented):
Looks like you missed the `noOptionsText` prop type.

wilysword (approved):

daniloPenaR (approved):

PR #751, opened 2022-05-03
CACTUS-454: add swipe functionality to Calendar

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-454

Not sure why I had to change the hook in Grid to get the DateInput's focus-on-open to work. The render order should be the same, but I guess _something_ changed somehow. 🤷

IE isn't supported, but since I assume swiping is mostly for mobile I doubt that's a big deal. The other thing that requires attention is the "residual", where it keeps sliding after you let go if you move fast. I played around with numbers and functions, but didn't really get it to where it feels completely natural to me; if any of you have ideas or suggestions, I'd be happy to hear them. It might just be because I was using a mouse and not an actual touchscreen.

### DIFF
--- a/modules/cactus-web/src/Calendar/Grid.tsx
+++ b/modules/cactus-web/src/Calendar/Grid.tsx
@@ -286,7 +286,7 @@ const CalendarGridBase = ({
   }, [state.overflow])
   // Second effect: if the grid is not currently focused,
   // make sure there's exactly one gridcell with tabIndex == 0.
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     const gridElement = gridRef.current
     const active = document.activeElement as HTMLElement | null
     // If the grid is currently focused, we don't need to do anything.
@@ -376,12 +376,14 @@ const CalendarGridBase = ({
 
 // TODO Figure out the right style for selected `.outside-date`
 const OUTLINE = { thin: '2px' }
+export const GRID_WIDTH = 300
 export const CalendarGrid = styled(CalendarGridBase)
   .withConfig(omitProps<CalendarGridProps>(margin))
   .attrs({ as: CalendarGridBase })`
   box-sizing: border-box;
   display: inline-block;
-  width: 300px;
+  width: ${GRID_WIDTH}px;
+  flex-shrink: 0;
   padding: 0 10px;
   ${margin}
   ${textStyle('small')}
@@ -394,6 +396,7 @@ export const CalendarGrid = styled(CalendarGridBase)
   }
 
   span {
+    user-select: none;
     box-sizing: border-box;
     display: inline-block;
     padding: 8px 0;
--- a/modules/cactus-web/src/Calendar/Slider.tsx
+++ b/modules/cactus-web/src/Calendar/Slider.tsx
@@ -1,68 +1,475 @@
 import { color } from '@repay/cactus-theme'
+import { noop } from 'lodash'
 import React from 'react'
-import { CSSTransition, TransitionGroup } from 'react-transition-group'
+import { CSSTransition } from 'react-transition-group'
 import styled from 'styled-components'
 
 import { isIE } from '../helpers/constants'
+import { useBox, useRenderTrigger } from '../helpers/react'
+import { GRID_WIDTH } from './Grid'
 
 export type SlideDirection = 'left' | 'right'
+type Transition = SlideDirection | 'recenter'
 
-export interface SliderProps {
+export interface TransitionState {
   transition?: SlideDirection
   transitionKey?: string
 }
 
+type GridRenderer = (date: Date, key: React.Key) => React.ReactElement
+interface SliderProps
+  extends TransitionState,
+    Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange'> {
+  month: number
+  year: number
+  render: GridRenderer
+  onChange: (date: Date, e: React.SyntheticEvent) => void
+  value: unknown
+}
+
+const GRID_HALF = GRID_WIDTH >> 1
 const TIMEOUT = isIE ? 400 : 300
+const GRAVITY = 0.007
+const CSS_CLASS = 'slider'
 
-const Slider: React.FC<SliderProps> = ({ transition, transitionKey, children }) => (
-  <TransitionGroup component={SlideWrapper}>
-    <CSSTransition key={transitionKey} timeout={TIMEOUT} classNames={transition}>
-      {children}
-    </CSSTransition>
-  </TransitionGroup>
-)
+interface Position {
+  x: number
+  time: number
+}
 
-export default Slider
+interface EventData {
+  left: number
+  right: number
+  month: number
+  offset?: number
+  state?: RenderState
+}
 
-// Using `-enter ~ *` works better than `-exit` when you alternate directions.
-const SlideWrapper = styled.div`
-  position: relative;
-  background-color: ${color('lightContrast')};
+interface TransitionOpts {
+  left?: number
+  right?: number
+  duration?: string
+}
 
-  > * {
-    position: absolute;
-    :first-child {
-      position: relative;
+type RenderState = 'idle' | 'preswipe' | 'swiping' | 'animation'
+
+class SliderState {
+  public eventData?: EventData
+  private positions: Position[] = []
+  private timeoutId: any = null
+  private transition: Transition | undefined = undefined
+  private transitionKey: string | number | undefined = undefined
+  private state: RenderState = 'idle'
+  private left = 0
+  private right = 0
+  private year = 0
+  private month = 0
+  private origin = NaN
+  private triggerRender: () => void
+  private setDate: (d: Date, e: React.SyntheticEvent) => void = noop
+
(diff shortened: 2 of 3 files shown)


modules/cactus-web/src/Calendar/Slider.tsx
```diff
@@ -1,68 +1,475 @@
...
+  private transition: Transition | undefined = undefined
+  private transitionKey: string | number | undefined = undefined
+  private state: RenderState = 'idle'
+  private left = 0
+  private right = 0
+  private year = 0
+  private month = 0
+  private origin = NaN
+  private triggerRender: () => void
+  private setDate: (d: Date, e: React.SyntheticEvent) => void = noop
```
maxrcollins, line 69: I'm not going to be able to make a suggestion until I look at this a while longer because this is a bit complicated, but my initial thoughts are that this file should just be the slider logic, and all of the date logic should be in the 'Calendar' file somehow. Keeping things here like `year`, `month`, and `setDate` kind of violate the separation of concerns that I think the separate files were originally intended to uphold.

modules/cactus-web/src/Calendar/Grid.tsx
```diff
@@ -286,7 +286,7 @@ const CalendarGridBase = ({
   }, [state.overflow])
   // Second effect: if the grid is not currently focused,
   // make sure there's exactly one gridcell with tabIndex == 0.
```
maxrcollins, line 288: I know this isn't part of this task but why do we want this? I'm noticing as I tab through this component that it isn't accessible to someone who only uses a keyboard. There's likely something I'm missing but it's weird that as I'm going through the days, only the 20th is tabbable for any month.

modules/cactus-web/src/Calendar/Slider.tsx
```diff
@@ -1,68 +1,475 @@
...
+
+        // We only track primary pointers, so this means there's multiple pointer devices.
+        case 'swiping':
+        case 'preswipe':
+          break
+      }
+      this.endEvent(e)
+    }
+    // TODO What should happen with multiple touches? Just ignore them?
+    // Or stop responding to pointermove events?
```
maxrcollins, line 301: We should probably ignore all pointer events while there are more than one. Someone could be using two fingers to try and enlarge the calendar/page and we don't want the calendar doing anything.

modules/cactus-web/src/Calendar/Slider.tsx
maxrcollins: I'm not arguing for a complete separation of concerns. I actually think that the Slider kind of has to be tailored to this specific component. But for readability's/future-proof's sake, I do think that there is something to be said about at least keeping the data organized and the roles a little more streamlined.

I think I see what you mean with the key issue. I'm not sure how to get away from 'months' here.

I want to see what other people think. It's just a lot to read through, especially because of how custom and controlled it is. I was thinking of ways to make this component a little cleaner overall. I stand by most of my comments, but I also recognize that this repo and a lot these code paradigms are new to me, so there's a good chance it's just me not fully understanding this.

modules/cactus-web/src/Calendar/Slider.tsx
```diff
@@ -1,68 +1,475 @@
...
+    if (transitionKey && transition) {
+      state.enterAnimation(transition, transitionKey)
+    } else {
+      state.clearAnimation()
+    }
+  }
+
+  const grids = state.getGrids(render)
+
+  delete props.value
```
maxrcollins, line 431: It's kind of a react anti-pattern for a component to edit the props it receives. I'm actually kind of surprised you can do this (https://stackoverflow.com/questions/37358135/can-i-delete-props-in-react#:~:text=No%2C%20you%20can't.,copy%20of%20the%20prop%20data). But it feels like it's on the parent component to not pass something that would be actively bad for the child component to have.

wilysword: The props object passed to the component is immutable, but if you unpack it using `{ propName, ...rest }` syntax the `rest` variable is just a normal object that can be edited in any way desired. In this case I should have included `value` in the destructuring up above, but didn't think to do so. (I'll fix that as I address any other PR suggestions.) As to why I passed the value when it isn't needed, it's a simple matter of efficiency: passing `{...this.state}` instead of having to do something like
```
const { value, ...rest } = this.state
...
<Slider {...rest} />
```

modules/cactus-web/src/Calendar/Slider.tsx
```diff
@@ -1,68 +1,475 @@
...
+  month,
+  year,
+  onChange: setDate,
+  render,
+  transition,
+  transitionKey,
+  ...props
+}: SliderProps) => {
+  const triggerRender = useRenderTrigger()
+  const state: SliderState = useBox({ setDate, year, month }, initState, triggerRender)
```
maxrcollins, line 417: useBox is pretty clever. But I may have named this something more reflective of what it does. useMutableRef? idk. Mostly talking to myself.

wilysword: I don't entirely remember my thought process, but I'm pretty sure I pulled the name from Java or C# terminology: when you pass a primitive to a variable that expects an Object, it's "boxed" into the standard Object framework. When using a boxed variable you can change the value but the reference stays the same, which seemed pretty applicable here.

modules/cactus-web/src/Calendar/Slider.tsx
maxrcollins: Ah I should've looked up again. I didn't realize that you were assigning some of the props to `props`.  Could you rename that to `rest` as well to make it a little more semantic at a glance?

Wait will your linter not yell at you for breaking out a variable that doesn't get used anywhere?

wilysword: The linter will complain about unused variables, but only if you don't have a `...rest` variable. Presumably they count an unused variable in that case as equivalent to `delete rest.whatever`.

I actually think I'm going to just remove the rest object entirely; I originally had it because the styled component was structured differently so I was forwarding `className`, but now I'm not actually forwarding any props at all so I'll just get rid of it.

modules/cactus-web/src/Calendar/Slider.tsx
```diff
@@ -1,68 +1,475 @@
...
+          const currentDeltaX = e.clientX - this.origin
+          this.exitSwipe(e.pointerId)
+          if (points.length && e.clientX === points[points.length - 1].x) {
+            points[points.length - 1].time = now
+          } else {
+            points.push({ x: e.clientX, time: now })
+          }
+          const velocity = calculateResidualVelocity(points)
+          const timeToZero = Math.min(1500, Math.abs(velocity / GRAVITY))
+          const accel = -velocity / timeToZero
```
maxrcollins, line 352: Huh. I must've missed the physics class where they went over these specific formulas 🤣

wilysword: Simplified and rearranged, but it's the standard constant acceleration formulas:
```
f(x) = a/2 * t^2 + st + d
f(v) = at + s
```
where `s` is initial velocity and `d` is initial position. I'm actually wondering if it might be more natural-feeling to use a variable acceleration formula.

(It doesn't come up much, but my bachelor's degree is actually in mechanical engineering.)

modules/cactus-web/src/Calendar/Slider.tsx
```diff
@@ -1,68 +1,475 @@
...
+  year: number
+  render: GridRenderer
+  onChange: (date: Date, e: React.SyntheticEvent) => void
+  value: unknown
+}
+
+const GRID_HALF = GRID_WIDTH >> 1
 const TIMEOUT = isIE ? 400 : 300
+const GRAVITY = 0.007
+const CSS_CLASS = 'slider'
```
daniloPenaR, line 33: Maybe my only comment would be related to this, not sure that we have any kind of rules for sorting our file's structure... I feel this would be put at the very beginning right after the imports, then all the interfaces/type declaration, then state class (as it is) and `...rest`

modules/cactus-web/src/Calendar/Slider.tsx
Dhalton: I understand where you're coming from, Max, in terms of readability/maintainability, but I think Glen makes a good point here that these components aren't necessarily intended to adhere to a pattern revolving around separating concerns, and that it's unlikely that we'd need to generalize this solution for other use-cases. So I think some coupling between the Calendar component and its Slider is acceptable.

The additional point about the keys is what solidifies my opinion on this; in order to accommodate the suggestion, we would need a significant refactor which, for the purposes of this ticket, I would say is out of scope and, in my opinion, not worth the time or effort.

wilysword:
I couldn't figure out how to test the multi-pointer behavior, because for some reason when using a mouse it doesn't trigger the event if you're holding down another button. Still, I believe I've made all the recommended changes except the major refactor Max suggested.

daniloPenaR (approved):
Changes look good and the functionality seems to be good too. Thanks

maxrcollins (approved):

PR #754, opened 2022-05-09
CACTUS-915: fix visual tests, rename Docker image tags

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-915

As far as I can tell, the visual tests were still working in GitHub because the Docker container didn't mess with yarn at all: instead it just set up the container OS and used the entire repo as a volume (I think; it's not explained clearly in the GitHub Actions docs). So now we do the same.

While I was at it I reduced the duplication by splitting it into two images, and renamed the images to `cactus/tests` so they're less likely to conflict with any official or custom local image tags.


daniloPenaR (approved):

Dockerfile
```diff
@@ -1,60 +1,4 @@
...
-  libxss1 \
-  libxtst6 \
-  ca-certificates \
-  fonts-liberation \
-  libappindicator1 \
-  libnss3 \
-  lsb-release \
-  xdg-utils \
-  wget
+FROM cactus/tests:base
```
Dhalton, line 1: This is probably just my lack of knowledge about docker, but I'm curious where this `cactus/tests:base` is/comes from. I don't see another dockerfile that does all of the stuff this file was doing...how does this work?

wilysword: The `yarn test:build` command now builds two images, `:base` and `:local`; `:base` is built from the same Dockerfile used by the "storyshot" GitHub action.

Dhalton: Right, okay, I forgot that Dockerfile existed. Makes much more sense now, thanks

Dhalton:
I keep getting this error when I run `yarn test:build` and `yarn test:docker`: `Could not find browser revision 818858. Run "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox yarn install" to download a supported Firefox browser binary.` I assume this is just on my machine though? @daniloPenaR the tests passed for you?

daniloPenaR (commented):
Dismiss my last approval. I tried this again and indeed it's not working, today I haven't test it again because for some reason Docker desktop it's not starting and I got a
```shell 
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.```

**Comment by Dhalton** [comment]
@daniloPenaR If you're having trouble getting docker desktop to start then it makes sense that it wouldn't run

**Comment by Dhalton** [comment]
@wilysword So I've been playing around with this today and so far I've been able to get it to work by installing chrome in the storyshots docker container:
```
RUN wget -q https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
RUN dpkg -i google-chrome-stable_current_amd64.deb
ENV PUPPETEER_EXECUTABLE_PATH='/usr/bin/google-chrome'
```
I still get 34 failures but I'm thinking that's probably because the storyshots were generated with firefox...I haven't been able to figure out how to install firefox in the container, but if you know how, we could do that. If nothing else, we can just re-generate the snapshots using chrome and go from there. 🤷

**Comment by wilysword** [comment]
That really doesn't make any sense, though, I only changed things related to yarn, which shouldn't effect browsers. Are you sure it's not something that changed on mac? I'd suggest two other tests:
1. Go back to a commit on master before we switched to yarn berry, and try to build/run the tests.
2. Try adding this to the Dockerfile (may have to play around with it a bit to get it to work):
```
COPY . /code
RUN yarn install
```
If it's the second one, then yarn must somehow be installing an entire browser or something, which seems crazy.

Dhalton:
It could be that something changed with Mac. I did upgrade to Mojave somewhat recently -- a few weeks ago or so. I'll try out your suggestions

Dhalton:
So I reverted to a commit before we switched to yarn berry and the tests built and ran fine for me there. I'm about to log off for the day so I'll have to get back to your tomorrow about 2

Dhalton:
Copying the project into the docker container and installing didn't fix anything either; still getting the same errors there

wilysword:
Did you switch back to `-v \"/$(pwd)/modules/cactus-web/__image_snapshots__:/code/modules/cactus-web/__image_snapshots__\"` in the `test:docker` command? (I forgot about that when I made my suggestion.) Even if you install in the container, if you include the entire repo as a volume it will just overwrite that.

To be honest, I'm not sure how much I can contribute to fixing this, given that I can't reproduce the issue with my hardware. Even if I were to make/suggest fixes, the feedback loop would be rather long. If you can pinpoint the problem, I can try to find a solution; but chances are that by the time you find the problem you'll _have_ a solution.

Dhalton:
Ok mounting the snapshots volume did fix it for me. So when copying everything and installing, the tests pass on my end. I've been looking in the `@storybook/addon-storyshots-puppeteer` and the `puppeteer` code trying to figure out why that is, but I haven't quite been able to figure it out yet. Before I go down a rabbit hole, is there any reason why we can't just copy the project into the container and install the dependencies?

wilysword:
No, that's fine; I only took it out because I didn't realize that `yarn install` would behave differently on different OSs (I figured JS would all be OS-agnostic). It was split up before because of how Docker layers images: by copying just the `package.json` files at first, the `yarn install` step would only need to be run if any of the `package.json` files changed, instead of if _any_ file changed.

Unfortunately, with how yarn berry works, copying everything piecemeal will be a _lot_ of instructions/image layers; but on the other hand if we just `COPY . /code`, then we'll have to rebuild after every change and the rebuild process will be longer. I'm not sure what the best balance is between space/readability and performance.

Dhalton:
Ok, thanks for looking into it, let me know when you're ready for me to review again

Dhalton:
This doesn't make any sense; I pulled your latest changes and the tests run, but I get 28 failures. I would think no big deal let's just update the snapshots, but they didn't fail on github actions and they must not have failed for you either, so I'm assuming that if we updated the snapshots to pass for me, they'd just start failing on those two fronts. I don't understand why, though. All you did differently from the way I had it set up before was add only the files pertaining to yarn and package dependencies, rather than everything as a whole.

maxrcollins:
It's working properly for me so I'm going to drop an approval. I'll test again if the issues Dhalton is having is an issue with the PR and not an issue with his setup

wilysword:
Made another branch with an alternate way of installing [here](https://github.com/repaygithub/cactus/commit/ca484636c30ddde80e44c7c351a7d8a375ffafef). I didn't push it directly to this branch because depending on how we have to fix Dhalton's issue it may not be compatible, but if we _can_ use it it'll be quite a bit cleaner.

Dhalton:
@wilysword I still get 28 failing tests on this branch, but the other commit you pushed up seems to be working fine for me if you want to bring it over to this branch

daniloPenaR (approved):
Tests are all passing, merge this baby in.

PR #756, opened 2022-05-27
fix(cactus-web): fix event handling on touch screens

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-454

Apparently touch screens cancel pointer events after a short delay for some reason beyond the ken of mere mortals like me. Fortunately the spec writers weren't complete morons and provided a simple way to restore logical functionality.

### UAT
Chrome-based browser devtools have a touch-screen simulator that I used to reproduce & fix the issue (not sure how to get dev code running on an actual mobile device). Firefox supposedly has something similar, but it didn't reproduce the bug so it can't be used to validate the fix.

### DIFF
--- a/modules/cactus-web/src/Calendar/Slider.tsx
+++ b/modules/cactus-web/src/Calendar/Slider.tsx
@@ -456,6 +456,7 @@ const Slider = ({
 const StyledSlider = styled.div<{ $offset: number }>`
   overflow: hidden;
   background-color: ${color('lightContrast')};
+  touch-action: none;
 
   .slider {
     display: flex;


Dhalton (approved):
Was able to test on my phone, works great 👍

maxrcollins:
This is working perfectly on my android now, but it is still getting stuck on my iPhone.

https://caniuse.com/css-touch-action

`iOS Safari only supports auto and manipulation.`

maxrcollins (changes requested):
Left in a separate comment that `touch-action: none;` isn't supported on Safari on IOS devices so this didn't fix iPhones

Dhalton:
> Left in a separate comment that `touch-action: none;` isn't supported on Safari on IOS devices so this didn't fix iPhones

Weird, I tested on my iPhone and I didn't notice any issues. Maybe I just didn't know how to cause the problem? I was just swiping left & right, I thought that would be enough to reproduce it

wilysword:
@Dhalton caniuse said that the Safari version is tied to the OS version, so I guess his OS version is three years out of date?

maxrcollins:
@Dhalton @wilysword My iPhone is the work phone that Repay sent me a month ago. 

Also it looks like they may have fixed that tbh. And I'm now realizing that it does work better than it did before that style. Regardless, I just took this screenshot after your fix:
image

I'm going to try and figure out what exactly I did.

PR #757, opened 2022-05-31
CACTUS-780 :: MIN-ITEMS fix

daniloPenaR:
[CACTUS-780](https://repayonline.atlassian.net/browse/CACTUS-780)

Thread of the [issue](https://repay.slack.com/archives/CABGTU4S3/p1653493399955329)


maxrcollins (approved):
> Also, if I click through the pages for long enough, eventually a 5th button appears. 

I haven't been able to replicate this but it seems like it's not really an issue as long as there are at least 5 buttons and everything fits so it looks good to me 👍

PR #759, opened 2022-06-02
CACTUS-899: use snyk cli to monitor repo

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-899

I tested the action and it works, but we won't know if the "on merge" condition works till we merge this PR. For now I just set the `SNYK_TOKEN` to my personal token, we can switch that out at any time.

Here's the successful run: https://github.com/repaygithub/cactus/actions/runs/2430326793

### DIFF
--- a/.github/workflows/snyk-scan.yml
+++ b/.github/workflows/snyk-scan.yml
@@ -1,0 +1,17 @@
+name: Snyk Scan
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


Dhalton (approved):
Cool that they wrote an action that we can use. Let's merge this baby and see if the on merge functionality works 👍

PR #762, opened 2022-06-09
CACTUS-922 :: Footer Hidden in layout 

daniloPenaR:
[CACTUS-922](https://repayonline.atlassian.net/browse/CACTUS-922)


wilysword (approved):

modules/cactus-web/src/Layout/grid.ts
```diff
@@ -314,7 +314,7 @@ const generateGridStyles = (components: ComponentLayout[]): StyleList => {
   `)
   // Some CSS engines are too dumb to figure out height/width from the fixed position offsets.
   fixed.width = `calc(100vw - ${fixed.left + fixed.right}px)`
```
wilysword, line 316: For consistency maybe we should change this to % as well, unless it breaks something.

PR #763, opened 2022-06-09
CACTUS-851 Snyk issues

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-851

Pretty much all the issues are transitive dependencies of Storybook or Gatsby, so there's not much we can do to fix them. Deleting and regenerating the lockfile got rid of about two thirds of them. Probably could have gotten rid of more, but I had to pin Storybook because they changed a lot of things about how controls work so upgrading that is probably a separate ticket's worth of work.


Dhalton (approved):
LGTM, any idea why the snyk scans are failing on this PR though?

wilysword:
Same reason they were before, because of yarn berry. Now that we have the console/github action scan working, not sure if we want to keep those auto-scans. The console scan only checks `package.json` files, while the auto-scan checks the `Dockerfile`s as well, but I don't know if we actually care about those.

maxrcollins (approved):

PR #765, opened 2022-06-24
CACTUS-941: allow any value type in the `Select` component

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-941

Wow this was so much more involved than I thought it would be. And now I want to refactor this component even more: this change should not have touched so many different places in the code. I ended up changing a lot of things that weren't very related just because they couldn't handle some aspect of the new value type.

A change that comes up quite a bit is to `altText`. Before, if we couldn't come up with an appropriate string, it would fall back to `value`. We still do that if possible, but if `value` is an object or something that obviously won't work. This lead to some internal changes:

- Practically rewrote `ValueSwitch`. At first it was because it detected changes using the `label`/`altText`, which wasn't guaranteed (especially on the first render). I changed that to directly compare the selected values, but somehow that broke the width calculation algorithm even though they seem unrelated. That was way more complicated than it needed to be, so I rewrote it as well. Finally I made a change to ensure we always had the best `altText` possible (see next bullet).
- Changed how the `label` -> `altText` fallback works. Before there was a specific component (`InternalOption`) with a `useEffect` hook that set the `altText` if needed. Nothing wrong with that, but as I looked into `ValueSwitch`, I realized that unless you could guarantee a re-render _after_ the effect hook ran, the new `altText` might not show up. (It seems that re-render _was_ happening, but the cause was unrelated and therefore not guaranteed.) Unfortunately the values and the options are in completely different parts of the render tree, so although I played around with several ways of connecting them, they all ended up feeling rather spaghetti. What I ended up with was a temporary ref callback: it could easily be attached to the `InternalOption`, and easily detected in the `ValueSwitch` to trigger a re-render without affecting the other logic in that component. It's still a bit spaghetti, and it wouldn't reflect "internal" changes in a label (i.e. translations). This is definitely an area I'd like to see further improved in the future.
- Changed how the `id` is assigned to options. Before it always used the `value` as part of the `id`; now that's not always possible, so I added a numeric fallback as well as a mechanism to persist the `id` across renders.
- Rewrote the value calculation in `getDerivedStateFromProps`. Since undefined can no longer be used as a blanket indication of the controlled/uncontrolled status, and the empty string can no longer be taken as a definite "clear" flag, the new logic ended up being quite a bit more complicated. Part of that was just because I made it more bullet-proof: if `props.multiple`, the value is always an array; invalid values are always dropped; stuff like that.
- Changed a few places with the assumption "value is array == props.multiple must be true", which is not a valid assumption now that arrays can _be_ values.


Dhalton (approved):
Yeah this component is an absolute mess. But the code looks good to me and everything worked in my testing as far as I can tell.

maxrcollins:
@_@

maxrcollins:
I'm unable to test in IE11 but this otherwise looks good to me

PR #766, opened 2022-06-29
CACTUS-942 :: Update Cactus Table component to allow for sticky right Column

daniloPenaR:
[CACTUS-942](https://repayonline.atlassian.net/browse/CACTUS-942)

Spent some time yesterday figuring out how to set the scrolling to emulate a sticky column in IE11, without success. So I leave it as it is for now


wilysword (commented):
This looks great. I think we may need to change the `dividers` styles, though: if you turn on dividers & sticky column, you can't see the last divider until you scroll all the way over. In fact, I kind of think the last column needs the dividers even if the `dividers` flag is false: otherwise it's not always obvious that it's sticky. (Want a second opinion on that, though.)

I also don't like the name `stickyRightColumn`. It's perfectly descriptive, but it's also long and limiting: if we ever want to support making other columns sticky, it would be awkward to add. I'd suggest something broader, like `sticky="right"` or `stickyCol="last"`, something along those lines, which would be easier to extend if we ever need to.

Dhalton (commented):
Yeah this looks solid. I agree with Glen that we should include the divider on the sticky column even if `dividers` is false; it's not obvious otherwise.

I also agree that the API could be a bit more flexible in case we need to add on to this functionality later on

Dhalton:
Also you'll probably want to merge #767 into this branch

modules/cactus-web/src/DataGrid/DataGrid.story.tsx
```diff
@@ -394,7 +395,7 @@ BasicUsage.argTypes = {
   nextText: { name: 'PrevNext: nextText', ...STRING },
   disableNext: { name: 'PrevNext: disableNext', control: 'boolean' },
   sortableCols: { name: 'include sortable columns' },
-  stickyRightColumn: { name: 'Sticky right column' },
+  sticky: { name: 'Sticky column' },
```
wilysword, line 398: Could you add the options here?

wilysword (commented):
Still need to address the dividers issue, but the I like the interface better.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -373,13 +385,17 @@ const table = css<TableProps>`
     }
     :last-child {
       border-right: ${(p): string => border(p.theme, 'lightContrast')};
-      ${(p) => (p.sticky === 'right' ? 'position: sticky; right: 0;' : '')};
+      ${(p) =>
+        p.sticky === 'right' ? `position: sticky; right: 0; ${boxShadow(p.theme, 1)}` : ''};
```
wilysword, line 389: It seems like this is the only one you actually need; when I removed the context changes it still worked.

Also, personally I think shadow 0 looks better than shadow 1 here, but you can get other opinions on that.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -20,6 +20,7 @@ type CellAlignment = 'center' | 'right' | 'left'
 type CellType = 'th' | 'td'
 type BorderCorner = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right'
 export type TableVariant = 'table' | 'card' | 'mini'
+export type stickyColAlignment = 'right' | 'none'
```
wilysword, line 23: You should capitalize this to make it consistent with other types.
```suggestion
export type StickyColAlignment = 'right' | 'none'
```

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -28,6 +29,7 @@ interface TableContextProps {
   cellIndex: number
   variant: TableVariant
   dividers?: boolean
+  sticky?: stickyColAlignment
```
wilysword, line 32: Remove this.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -334,6 +346,9 @@ const StyledHeader = styled.thead<TableHeaderProps>`
     text-transform: uppercase;
     border: ${(p): string => border(p.theme, 'base')};
     border-right: ${(p): string => (p.dividers ? border(p.theme, 'mediumContrast') : '')};
+    :last-child {
+      ${(p) => (p.sticky === 'right' ? boxShadow(p.theme, 1) : '')};
+    }
```
wilysword, line 351: Remove this.

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -370,12 +385,17 @@ const table = css<TableProps>`
...
+        p.sticky === 'right' ? `position: sticky; right: 0; ${boxShadow(p.theme, 1)}` : ''};
     }
   }
   tr:nth-of-type(even) {
     td,
     th {
       background-color: ${(p): string => p.theme.colors.lightContrast};
+      :last-child {
+        ${(p) => (p.sticky === 'right' ? boxShadow(p.theme, 1) : '')};
+      }
```
wilysword, line 398: Remove this.

wilysword (approved):

Dhalton (approved):
LGTM

maxrcollins:
Looks good on IE too!

PR #768, opened 2022-07-08
CACTUS-931 Fix prop forwarding

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-931

One question I wanted to ask regarding the first commit: should we name `CactusIcon` differently? `AbstractIcon` might be a better name since it's not actually an icon, or maybe `StyledIcon`. (Or something else if you guys can come up with a better name.)

Most of the changes were changing the prop forwarding but I did remove some default props, mostly booleans that defaulted to false but don't differentiate from `undefined` which is also falsy.

Check individual commit messages for more info, but this is the most important part:

This is the culmination of my understanding of `styled-components` types, which has been improving over the years: and I've been trying to make better use of _our_ types so they work as expected, especially when the polymorphic `as` prop is in play. Here's what I've learned:

- When you do `X = styled(C)<StyleProps>()`, the props of X combine _two_ types: `React.ComponentProps<typeof C> & StyleProps`.
- When you use the `as` prop, it swaps out _only_ the first type: `<X as={Y} />` will validate props against the type `React.ComponentProps<typeof Y> & StyleProps`.
- Therefore, the resulting type will be most flexible & correct if the type of the base component **does not include** the style props.
- It's also best to utilize `shouldForwardProp` to remove the style props, so they won't be forwarded to the base _or_ polymorphic components.

Unfortunately the solution I've been using for that, `omitProps`, is breaking the types (more info in the comment in `helpers/omit.ts`); I'd like to fix the existing calls, but it would require more effort than I can put into this unrelated ticket.

### DIFF
--- a/modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
+++ b/modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
@@ -147,7 +147,6 @@ ConfirmModal.propTypes = {
 ConfirmModal.defaultProps = {
   cancelButtonText: 'Cancel',
   confirmButtonText: 'Confirm',
-  iconName: undefined,
   iconSize: 'medium',
   title: 'Modal Title',
 }
--- a/modules/cactus-icons/package.json
+++ b/modules/cactus-icons/package.json
@@ -61,7 +61,6 @@
     "testEnvironment": "jsdom"
   },
   "peerDependencies": {
-    "prop-types": ">=15.0.0",
     "react": "^16.8.3 || ^17.0.0",
     "styled-components": "^5.1.0"
   },
@@ -87,7 +86,6 @@
     "jest": "^27.5.1",
     "jest-sonar": "^0.2.12",
     "jest-styled-components": "^7.0.8",
-    "prop-types": "^15.8.1",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "styled-components": "^5.3.3",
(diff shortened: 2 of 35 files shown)


modules/cactus-web/src/ConfirmModal/ConfirmModal.tsx
```diff
@@ -147,7 +147,6 @@ ConfirmModal.propTypes = {
 ConfirmModal.defaultProps = {
   cancelButtonText: 'Cancel',
   confirmButtonText: 'Confirm',
-  iconName: undefined,
```
maxrcollins, line 150: lol

maxrcollins:
I went through this and tested some of the components and everything makes sense. I'm a little shaky on how the Icon Library works though so I wont be as helpful there.

I don't really know what the CactusIcon is doing (styled noop? 😵) but it looks like you're just putting a config together? If so, maybe `StyledIcon` like you said or `IconConfig` or something. I just don't think that AbstractIcon really means anything.

wilysword:
The icons are generated from the template in `svgr.config.js`. `CactusIcon` defines the styles, but doesn't render anything; in each individual icon file it calls `CactusIcon.withComponent` which swaps out the noop for an actual icon, while keeping the same defined styles.

I thought of the name `AbstractIcon` because it's similar to abstract in OOP (has a definition, but doesn't do anything until the implementation is overridden), but I admit it's not the clearest name.

daniloPenaR:
What about `IconInterface`?

wilysword:
Per the [poll results](https://repay.slack.com/archives/CABGTU4S3/p1657818793581149), I renamed the base component.

maxrcollins:
> Per the [poll results](https://repay.slack.com/archives/CABGTU4S3/p1657818793581149), I renamed the base component.

Democracy at work

maxrcollins (approved):

PR #771, opened 2022-07-14
fix(cactus-web): fix height on MenuBar dark variant

maxrcollins:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-961

Note that this bug isn't outlined in the ticket, but the height is different between the two sized menu bars. I can't tell if it's new or if it's been there a while, but while we're fixing up this component, I'd like to have them be more uniform.

UAT:
Open the Menu bar locally in storybook and check the height between the light and dark variants.

Also make sure everything still looks/works as it should. eg: thin/thick doesn't break anything.


Dhalton (approved):
LGTM thanks for taking care of this

daniloPenaR (approved):
LGTM, take in count [this](https://github.com/repaygithub/cactus/pull/772)

wilysword (changes requested):
This no longer matches the style guide...

maxrcollins:
> Style guide specifies bottom border on hover should be 3px (4px thick). Technically it also specifies that the focus border should be even all the way around, though I don't mind ignoring that one quite as much. (Regarding the way the text shifted on focus, I actually didn't mind that, since it didn't affect readability but served as a further visual indicator of focus shift.)

I'll see what I can do about the thickness. Right now it's 3px (5px thick) but I think the 1px bigger bottom border is closer to what than the 5px difference in height we had earlier. 

As for the focus state, I see what you mean. I don't really mind the text shifting in the dark theme either but it doesn't currently do that on the light theme. And I DO mind them behaving differently.  If you would rather I leave it as it was before with the text shifting, I can do that but you should bring it up in the thread so we can get a group consensus there and maybe we'd have it do the same in the light theme.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -80,16 +85,18 @@ const variantMap: VariantMap = {
...
-      border-bottom: ${border('base', { thin: '3px', thick: '4px' })};
+      border-bottom: ${border('transparent', { thin: '2px', thick: '3px' })};
 
       &:hover:not([aria-disabled]),
-      &[aria-expanded='true'] {
-        border-color: ${color('white')};
+      &[aria-expanded='true'],
+      &:focus {
+        box-shadow: inset 1px 1px 0 ${(p) => thicknessMap[p.theme.border]} ${color('white')},
+          inset -1px -1px 0 ${(p) => thicknessMap[p.theme.border]} ${color('white')};
```
maxrcollins, line 94: This isn't obvious up front but to keep the menu heights the same size, I couldn't increase/decrease the border size, so I did something a little fancy with a box shadow instead to make the bottom border the correct size

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -80,15 +85,19 @@ const variantMap: VariantMap = {
...
       &[aria-expanded='true'],
       &:focus {
-        ${insetBorder('white', undefined, { thin: '1px', thick: '2px' })};
+        box-shadow: inset 1px 1px 0 ${(p) => thicknessMap[p.theme.border]} ${color('white')},
+          inset -1px -1px 0 ${(p) => thicknessMap[p.theme.border]} ${color('white')};
         border-bottom-color: ${color('white')};
       }
+      &:focus {
+        border-bottom: 0;
+        padding-bottom: ${(p) => (p.theme.border === 'thin' ? '26px' : '27px')};
```
wilysword, line 99: I'd suggest using the `borderSize` helper for this kind of statement.

modules/cactus-web/src/MenuBar/MenuBar.tsx
wilysword: This seems excessively complicated, and is exactly the kind of thing I was afraid of when I decided not to make them the same height. (I still don't understand why that's such a big deal to you...)

If you're really going to do this, I'd suggest using normal borders instead and setting variable padding using the `borderSize` helper.

Dhalton:
> (I still don't understand why that's such a big deal to you...)

I mean the heights were the same before, and they're the same in the style guide, so isn't increasing the height of the dark variant _also_ a deviation from the style guide?

If it's going to be a huge pain, then let's just release it as-is and see if the channels team pushes back on it.

modules/cactus-web/src/MenuBar/MenuBar.tsx
maxrcollins: This change has nothing to do with the height. It has to do with the bottom border being 1px off that you brought up. I was still able to use `insetBorder` until this. I don't care deeply about it being 1px off but it was very noticeable to me that the headers were two different sizes. Also until this change, there was still a small amount of spacing discrepancy between the menu items because of the border on dark mode. Go to the [dev site](https://repaygithub.github.io/cactus/stories/cactus-web/?path=/story/menubar--menu-bar-dark) and switch back and fourth. Fixing the spacing was the point of this ticket to begin with. Changing the color theme should not affect any of the spacing which is why I'm trying to clean this up. The text also shouldn't only be moving only on dark mode only on focus. Each of these individually might not be a big deal but together they absolutely warrant a fix.


Anyway, I'll look into using borderSize or maybe just doing this a different way. I don't really like the box shadow either.

wilysword: Idea: adjust padding based on `borderSize`, use a regular 1/2px border that changes color to white on hover/focus, along with a 2px inset bottom border on hover only. (Not sure if it'll work because I don't know exactly how the inset border is positioned inside a regular border: do they overlap?)

maxrcollins: Heh. I had actually started to do that. Thanks for the suggestion. I'm just trying to figure out a way to make this not look hacked together.

maxrcollins:
@wilysword I fixed this up in a way that I think you'll be happy with now. It's way cleaner. 

@Dhalton  All of the dimensions of everything stays the same across light/dark theme and nothing moves on hover

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -80,18 +85,21 @@ const variantMap: VariantMap = {
 
     > ${ScrollButton}, [role='menubar'] > li > [role='menuitem'] {
       color: ${color('white')};
-      border: ${border('base')};
-      border-bottom: ${border('base', { thin: '3px', thick: '4px' })};
+      border: ${border('transparent', { thin: '1px', thick: '2px' })};
 
       &:hover:not([aria-disabled]),
       &[aria-expanded='true'] {
+        ${insetBorder('white', 'bottom', { thin: '2px', thick: '2px' })};
```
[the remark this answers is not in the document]
maxrcollins: One of the things I did to keep the height consistent is to split the border that's always there from the border that gets added on hover. The border that is always there is 1px/2px. The inset border is actually just an inset box shadow which gets added inside the border but doesn't add any height. Since I've set it to 2px/2px, the final height of the white line at the bottom becomes 3px/4px.

Does that make sense?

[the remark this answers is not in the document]
Dhalton: Ah, yeah that makes sense. Thanks for explaining

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -63,6 +64,10 @@ const variantMap: VariantMap = {
     ${colorStyle('standard')}
     ${insetBorder('lightContrast', 'bottom', { thin: '2px', thick: '3px' })};
 
+    > ${ScrollButton} {
+      border: ${border('transparent', { thin: '1px', thick: '2px' })};
+    }
```
wilysword, line 69: This doesn't seem to do anything.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -80,18 +85,21 @@ const variantMap: VariantMap = {
 
     > ${ScrollButton}, [role='menubar'] > li > [role='menuitem'] {
       color: ${color('white')};
-      border: ${border('base')};
-      border-bottom: ${border('base', { thin: '3px', thick: '4px' })};
+      border: ${border('transparent', { thin: '1px', thick: '2px' })};
```
wilysword, line 88: These are the default thin/thick values, so you could omit them.

modules/cactus-web/src/MenuBar/MenuBar.tsx
```diff
@@ -80,18 +85,21 @@ const variantMap: VariantMap = {
...
       }
-
       &:focus {
-        border: ${border('white')};
+        border-color: ${color('white')};
       }
     }
+
+    [role='menubar'] > li > [role='menuitem'] {
+      padding: calc(24px - ${borderSize()}) calc(16px - ${borderSize()}) 25px;
```
wilysword, line 101: This is fine, but I think it might be slightly less confusing if you did it like this (matter of opinion?):
```suggestion
      padding: ${borderSize({ thin: '23px 15px 25px', thick: '22px 14px 25px' })};
```

wilysword (approved):
Functionally looks good, although I think you could still simplify just a bit more. Also make sure you've got the right storyshots generated.

modules/cactus-web/src/MenuBar/MenuBar.tsx
maxrcollins: No you're right. I also want to get rid of calc() because [IE has trouble with math](https://caniuse.com/?search=calc). I think this particular instance is fine, but I still like to avoid it.

modules/cactus-web/src/MenuBar/MenuBar.tsx
maxrcollins: It adds 1px of padding to the scroll buttons to bring them in line with the size of the dark mode button which gets a border. It's only a 1-2px difference and isn't obvious, even switching between the two themes, but it made everything EXACTLY the same. At the very least, I should've put this outside of the variantMap.

wilysword (approved):

PR #778, opened 2022-07-25
CACTUS-929 :: Fix BrandBar when no `logo`

daniloPenaR:
[CACTUS-929](https://repayonline.atlassian.net/browse/CACTUS-929)

Renders LogoWrapper if `logo` is not undefined. The change is applied not only for tiny screens, I consider the `isTiny` check unnecessary so the blank space is removed from both tiny and larger screens.


Dhalton (approved):

maxrcollins (approved):
Looks good 👍

PR #779, opened 2022-07-25
CACTUS-960 :: First Pass at Adding SCSS Linting Rules

maxrcollins:
Ticket & UAT: [CACTUS-960](https://repayonline.atlassian.net/browse/CACTUS-960)


maxrcollins:
I'm not sure why the build failed. yarn install runs fine locally 🤔

wilysword (approved):
Seems okay. Don't have much opinion on the rules, though.

Do you know how it finds CSS blocks? Is it only directly from the `styled` factory? Would it scan styles in blocks from the `styledWithClass` and `styledUnpoly` helpers I made? Or things like this?
```
const block = css`
  i-am-a-property: value;
`
```

maxrcollins:
> Seems okay. Don't have much opinion on the rules, though.
> 
> Do you know how it finds CSS blocks? Is it only directly from the `styled` factory? Would it scan styles in blocks from the `styledWithClass` and `styledUnpoly` helpers I made? Or things like this?
> 
> ```
> const block = css`
>   i-am-a-property: value;
> `
> ```

I don't know exactly how it works. It definitely goes through a round of compilation and leaves placeholders for variables (which is why I had to turn off certain keywords for 'no-unknown'). It also, for jsx files that don't have any styled components, it'll output an empty css block or something, because i had to disable: "no-empty-source" to not receive an error for every one of those files.  I really should look into this more but Stylelint/Styled-Components did [just deprecate their old official way of doing this](https://stylelint.io/migration-guide/to-14/#:~:text=(The%20%40stylelint/postcss%2Dcss%2Din%2Djs%20package%20has%20issues.%20It%20will%20likely%20to%20be%20deprecated%20in%20the%20future%20in%20favor%20of%20smaller%20syntaxes%20that%20focus%20on%20only%20one%20library%20(see%20this%20issue)).) without updating either of their documentation....  

Anyway it will definitely find your last example but at the same time it has trouble with something a little more complicated like:

```
  ${(p) =>
    p.$isSomeBool ? `
       something: something;
    ` : `
       something-else: something;
    `}
```
I don't think it can actually parse that. There are a few other similar cases. But I figured 75% of a linter was better than no linter, at least to begin with. Also since we seem to be moving away from `${(p) =>` anyway.

maxrcollins:
It'll also get confused sometimes. For example `TopSection.tsx` has an extra closing bracket on line 128 that needs to be removed, but it thinks the error is on line 143.

daniloPenaR (approved):
LGTM

Dhalton (commented):
Looks good, seems to work fine. I'm curious though, what are the `dummyValue`s in the `.stylelintrc.json` file for? Were those left there intentionally or was it for testing something?

maxrcollins:
>Looks good, seems to work fine. I'm curious though, what are the dummyValues in the .stylelintrc.json file for? Were those left there intentionally or was it for testing something?

So this works by compiling the styled components from JS into css. I don't fully get the process, but it's not actually getting anything for those variables. Instead, it just converts it to `dummyValue`. Maybe we could set up a way to test by adding real variables, but in the mean time, I'm have to ignore these so they don't clog the error list

Dhalton (approved):

PR #780, opened 2022-07-26
CACTUS-963: add variety of text-related styling props

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-963

There turned out to be quite a few properties here:
- `textDecoration`: brings underline & strikethrough, which are both useful
- `textOrientation`/`writingMode`: I thought these might be useful someday in vertical table labels or something, to save space
- `textTransform`: we've already used this in a couple of header-ish components
- `fontVariant`: seems like a more useful & flexible version of `textTransform`
- `textOverflow`: we've mostly handled this internally in components, but I wouldn't be surprised if it came up somewhere outside of those
- some of those are shortcut properties, so I included the full versions as well

`color` is foreground (usually text) color, so I included it in AllTextProps. Overflow doesn't quite fit, but you have to set it for `textOverflow` to work.

### DIFF
--- a/modules/cactus-web/src/helpers/storybook.ts
+++ b/modules/cactus-web/src/helpers/storybook.ts
@@ -50,7 +50,7 @@ const mapSpace = (str: string | undefined): Length | Length[] | undefined => {
   if (!parts.length) return undefined
   else if (parts.length > 1) return parts.map(mapSpace) as Length[]
   str = parts[0]
-  return /^[0-7]$/.test(str) ? parseInt(str) : str
+  return /^-?[0-7]$/.test(str) ? parseInt(str) : str
 }
 export const SPACE: ArgType = { ...STRING, map: mapSpace }
 
--- a/modules/cactus-web/src/Box/Box.story.tsx
+++ b/modules/cactus-web/src/Box/Box.story.tsx
@@ -41,6 +41,10 @@ export default {
     borderBottomLeftRadius: HIDE_CONTROL,
     borderBottomRightRadius: HIDE_CONTROL,
     textStyle: { options: textStyles },
+    textDecoration: STRING,
+    fontVariant: STRING,
+    textIndent: SPACE,
+    textTransform: STRING,
   },
   args: {
     padding: '4',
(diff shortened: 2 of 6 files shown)


Dhalton (approved):
Looks pretty sweet to me. I just have one pretty general question:

How do you know/get a comprehensive list of all of these properties? A lot of these I've never even heard of; stuff like `textDecorationStyle`, `textDecorationThickness`, `fontVariantLigatures`, `fontVariantNumeric`, all that nuanced jazz. I mean it's great that you included all of that here, I'm just curious where you get all of that info from. Is there like a list of all text-related css properties somewhere that you referenced, did you scour the internet finding all of these things in different places, or are you just that much of a css pro that this is all stuff you already knew?

Anyway, LGTM 👍

wilysword:
Basically, MDN. I started with [text-decoration](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration), because that's the one the ticket actually calls for; then I followed a couple of steps:

1. The description there says it's shorthand for a some other properties, so I included those.
2. The "See also" section at the bottom mentions a few other related properties, not part of the shorthand.

Finally, in the menu on the left there's a "Reference -> Properties" section. That list does not appear to be absolutely comprehensive, but it does have the most-supported properties, so I looked through it for properties starting with `text-*` or `font-*`, and for each of those I looked into shorthand properties & the "See also" section.

I also looked through the `csstype` source in order to set up the types correctly, but that was more confirmation of what I found on MDN than actually finding anything new. The only one I didn't find through that process was `user-select`, which I already knew about from past experiences.

modules/cactus-web/src/helpers/storybook.ts
```diff
@@ -50,7 +50,7 @@ const mapSpace = (str: string | undefined): Length | Length[] | undefined => {
   if (!parts.length) return undefined
   else if (parts.length > 1) return parts.map(mapSpace) as Length[]
   str = parts[0]
-  return /^[0-7]$/.test(str) ? parseInt(str) : str
+  return /^-?[0-7]$/.test(str) ? parseInt(str) : str
```
daniloPenaR, line 53: Is this only for making `-` optional?

daniloPenaR (approved):
LGTM

modules/cactus-web/src/helpers/storybook.ts
wilysword: Yeah, so you can enter negative numbers.

PR #786, opened 2022-08-09
cactus-979:: fix docs website snyk vulnerability

maxrcollins:
Ticket: [CACTUS-979](https://repayonline.atlassian.net/browse/CACTUS-979)

UAT:
Not really sure how to write it for this...Basically everything here was a patch except for the ansi-regex library which moved from 2.1.1 to 3.0.0 so make sure that it didn't break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,10 +73,6 @@
     "react-dom": "^17.0.0",
     "react-is": "^17.0.0",
     "deep-object-diff": "1.1.0",
-    "devcert": "1.2.0",
-    "parse-url": "^6.0.1",
-    "ansi-regex": "^3.0.1",
-    "trim": "^0.0.3",
-    "shell-quote": "^1.7.3"
+    "devcert": "1.2.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -74,6 +74,10 @@
     "react-is": "^17.0.0",
     "deep-object-diff": "1.1.0",
     "devcert": "1.2.0",
-    "parse-url": "^6.0.1"
+    "typescript": "~4.6.2",
+    "parse-url": "^6.0.1",
+    "ansi-regex": "^3.0.1",
+    "trim": "^0.0.3",
+    "shell-quote": "^1.7.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "react-is": "^17.0.0",
     "deep-object-diff": "1.1.0",
     "devcert": "1.2.0",
-    "typescript": "~4.6.2",
     "parse-url": "^6.0.1",
     "ansi-regex": "^3.0.1",
     "trim": "^0.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,6 +73,7 @@
     "react-dom": "^17.0.0",
     "react-is": "^17.0.0",
     "deep-object-diff": "1.1.0",
-    "devcert": "1.2.0"
+    "devcert": "1.2.0",
+    "parse-url": "^6.0.1"
```
wilysword, line 77: If this isn't a direct dependency, I'd prefer not to have in in the `package.json`. Over time such things can become bloated, and sometimes it's hard to tell why a given package is listed.

maxrcollins: I couldn't figure out how to do this without explicitly setting these values. I know it's not ideal. But I tried what you were talking about with the yarn.lock file and I couldn't get any of these to update properly. Feel free to let me know here what you propose, or even make a commit yourself if you want.

yarn.lock
```diff
@@ -8873,24 +8873,10 @@ __metadata:
...
-  linkType: hard
-
-"ansi-regex@npm:^5.0.0, ansi-regex@npm:^5.0.1":
-  version: 5.0.1
-  resolution: "ansi-regex@npm:5.0.1"
-  checksum: 2aa4bb54caf2d622f1afdad09441695af2a83aa3fe8b8afa581d205e57ed4261c183c4d3877cee25794443fde5876417d859c108078ab788d6af7e4fe52eb66b
+"ansi-regex@npm:^3.0.1":
+  version: 3.0.1
+  resolution: "ansi-regex@npm:3.0.1"
+  checksum: 09daf180c5f59af9850c7ac1bd7fda85ba596cc8cbeb210826e90755f06c818af86d9fa1e6e8322fab2c3b9e9b03f56c537b42241139f824dd75066a1e7257cc
```
maxrcollins, line 8879: I don't like that this replaced all three of the above. I only wanted to move the 2.x.x versions up to 3.x.x.   I tried `>=3.0.0`  but all of my resolutions end up resulting in just one ansi-regex anyway, so I may just leave this one alone unless someone knows how to do that

wilysword:
All you need to upgrade `parse-url` is this: https://github.com/repaygithub/cactus/commit/e45eabbb8cce40e59595b83a724452600fed85c3.

The other packages you're changing should not be changed, because you're ignoring the claimed dependencies of their "parents". Without knowing why they claimed those specific versions, you can't know if changing them will break anything.

wilysword (approved):

Dhalton (approved):
LGTM, just a reminder to use the correct [commit message format](https://github.com/repaygithub/cactus#commit-messages)

maxrcollins:
> LGTM, just a reminder to use the correct [commit message format](https://github.com/repaygithub/cactus#commit-messages)

whoops.

PR #788, opened 2022-08-16
CACTUS-981/982 :: Fix Snyk Issues

Dhalton:
https://repayonline.atlassian.net/browse/CACTUS-981

This image was missing a few things that we needed (`wget`, `unzip`, `procps`), but seems to be working now that I added them.

https://repayonline.atlassian.net/browse/CACTUS-982

I wasn't able to install `libappindicator1` with this image, but apparently we don't need it. Also had to add `procps` to this one too.

Nothing to test for either of these. If we get some green checkmarks we should be good to go.

### DIFF
--- a/actions/integration-tests/Dockerfile
+++ b/actions/integration-tests/Dockerfile
@@ -1,9 +1,12 @@
 
-FROM node:lts
+FROM node:16-bullseye-slim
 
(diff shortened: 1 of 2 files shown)


daniloPenaR (approved):
LGTM if there's nothing more to test.

maxrcollins (approved):
👍

PR #790, opened 2022-08-23
CACTUS-969: add more ways of pagination to DataGrid

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-969

This consists of three changes:
- Added `itemCount` and `itemOffset` as alternatives to `pageCount` and `currentPage` respectively (`pageSize` required to use).
- Added ability to control pagination via props passed directly to `PageSizeSelect`, `Pagination`, and `PrevNext`.
  - Along with this, I moved most of the pagination-related logic to its own file.
- Added `onPagisort` handler, combining `onSort` & `onPageChange` and allowing pagination & sorting to be done uncontrolled.

Originally this would have involved making `currentPage` and `pageCount` optional properties, but it turns out that would have been a breaking change to the `onPageChange` handler.

Also made some story changes:
- Sort is now entirely uncontrolled.
- Pagination is now controlled through sub-components.
- I'd have made that uncontrolled as well, but it wouldn't work with clone/delete.

And finally, the hooks:

`useControllableValue` is a hook I've been mulling around in my head since at least the last time I worked on the `Select` component, although I've been trying to come up with the "ideal" way of handling that use case basically since I started on the team. Not that complicated, though it took a while to get Typescript playing nicely with it. Eventually, I'd like to refactor a lot of our other components to use this.

`useRefState` is mostly to make `useControllableValue` viable in React 18, but I've had at least one other use case for it in Channels. It could also possibly be used as a replacement for `useStateWithCallback`, although I can't say for sure there would _never_ be a use case for running a callback after the new state value has been _rendered_, instead of just after it's been _changed_.

The new hooks directory/file is for two reasons:
1. `helpers/react.ts` is getting pretty crowded.
2. Often the hooks I write in cactus-web are useful enough that I want to use them in Channels (or elsewhere), so I wanted somewhere we could put all the "exportable" hooks. We can keep hooks for internal-use only in `helpers`, and slowly move the others to the `hooks` directory and export them via `hooks.ts`. (Speaking of which, should we fix `repay-scripts` to be able to handle index files?)

### DIFF
--- a/modules/cactus-web/src/DataGrid/DataGrid.test.tsx
+++ b/modules/cactus-web/src/DataGrid/DataGrid.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import React, { useContext, useState } from 'react'
 import { MarginProps } from 'styled-system'
 
@@ -260,7 +260,7 @@ describe('component: DataGrid', () => {
     const { getByLabelText, getByText } = renderWithTheme(<DataGridContainer />)
 
     const sixItemsButton = getByLabelText('View 6 rows per page')
-    fireEvent.click(sixItemsButton)
+    userEvent.click(sixItemsButton)
     expect(getByText('Showing 1 to 6 of 11')).toBeInTheDocument()
   })
 
@@ -268,7 +268,7 @@ describe('component: DataGrid', () => {
     const { getByText } = renderWithTheme(<DataGridContainer />)
 
     const createdHeader = getByText('Created').parentElement as HTMLElement
-    fireEvent.click(createdHeader)
+    userEvent.click(createdHeader)
     expect(createdHeader.parentElement).toHaveAttribute('aria-sort', 'ascending')
   })
 
@@ -283,7 +283,7 @@ describe('component: DataGrid', () => {
     const { getByText, getByLabelText } = renderWithTheme(<DataGridContainer />)
 
     const page2Button = getByLabelText('Go to page 2')
-    fireEvent.click(page2Button)
+    userEvent.click(page2Button)
     // First result on page 2
     expect(getByText('Config 5')).toBeInTheDocument()
     // Last result on page 2
@@ -294,7 +294,7 @@ describe('component: DataGrid', () => {
     const { getByText } = renderWithTheme(<DataGridContainer providePageCount={false} />)
 
     const nextPageButton = getByText('Next')
-    fireEvent.click(nextPageButton)
+    userEvent.click(nextPageButton)
     // First result on page 2
     expect(getByText('Config 5')).toBeInTheDocument()
     // Last result on page 2
@@ -318,6 +318,7 @@ describe('component: DataGrid', () => {
     expect(dataGrid).toHaveStyle('margin-left: 40px')
     expect(dataGrid).toHaveStyle('margin-right: 40px')
   })
+
   test('Should support setting position of right column as sticky', () => {
     const { getAllByTestId, rerender } = renderWithTheme(<DataGridContainer sticky="right" />)
     getAllByTestId('last-col').forEach((element) => {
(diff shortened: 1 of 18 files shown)


modules/cactus-web/src/hooks/useControllableValue.ts
```diff
@@ -0,0 +1,92 @@
...
+    props: P,
+    normalize: Normalizer<P, S>,
+    reducer: Reducer<S, A>,
+    initArg: I,
+    initializer: (i: I) => S
+  ): [S, SyncDispatch<S, A>]
+}
+
+// Uses generic types externally, but concrete types internally to validate logic.
+type Props = { value?: number }
```
Dhalton, line 59: The fact that value is always a number is really confusing me...I mean I can tell by the tests and your comments that it's not _really_ always a number, but I just don't understand why we have to explicitly give it that type in the first place. I'll leave some more comments about this below

modules/cactus-web/src/hooks/useControllableValue.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// If the `props` object passed is mutable, the given key will be deleted after extraction.
+const extractByKey = (props: Props, key: 'value', lastState: number) => {
```
Dhalton, line 62: For instance, why does `lastState` have to be a number? Would `any` not work? Or the generic types?

modules/cactus-web/src/hooks/useControllableValue.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const useControllableValue = (
+  props: Props,
+  key: 'value' | Normalizer<Props, number>,
+  ...args: [Reducer<number, void>, string, (s: string) => number]
+): [number, SyncDispatch<number, void>] => {
+  const ref = useRefState(...args)
+  const preRenderState = ref.current
+  const value: number =
```
Dhalton, line 77: And here, I get that this is type number because either `props.value` is set to number or the normalizer is supposed to return a number, but I'm still just not seeing why we have to set an explicit type in those places...I know in a comment above you say that concrete types are used internally to validate logic, but I guess I'm just not seeing what logic is being validated

modules/cactus-web/src/hooks/useControllableValue.ts
```diff
@@ -0,0 +1,92 @@
...
+    typeof key === 'function'
+      ? key(props, preRenderState)
+      : extractByKey(props, key, preRenderState)
+
+  useEffect(() => {
+    // We just used this value, so don't need to trigger a re-render;
+    // however, only update if nothing has changed since render started.
+    if (ref.current === preRenderState) {
```
Dhalton, line 85: I see that we're checking equality here which is maybe part of the logic you're referring to, but I don't see what the benefit of using `number` over `any` is

Dhalton (commented):
I'm struggling to wrap my head around this so my review is going to be piecemeal and might come off as disorganized. These comments are all related to one particular thing that I'm not quite understanding so I figure I'll just get these out there to start with. I'm not necessarily opposed to the `number` thing; I'm just trying to understand why you wrote it this way.

modules/cactus-web/src/hooks/useControllableValue.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * When controlled, the prop value is always used for rendering but the state
+ * isn't updated till rendering finishes to avoid concurrency issues (React 18).
+ * State updates might operate on "old" state in the interim, but since such
+ * updates generally arise from user input, and the users haven't even seen the
+ * "new" state yet, this is probably the correct behavior.
+ *
+ * If the state DOES change during render, the controlled value is dropped
+ * under the assumption that another render is already pending as a result
+ * of the concurrent state change.
```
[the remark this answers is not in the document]
Dhalton: Is it the `useEffect` on line 82 where you don't set `ref.current = value` unless the current state and pre-render state are the same?

wilysword:
There are two interconnected reasons why I used concrete types instead of `any` or trying to do some generic thing. They're more relevant to `useRefState` than `useControllableValue`, but still worth it in my opinion.
1. Each of these hooks has several overloads. Trying to get a set of types that is simultaneously compatible with all of them is difficult; at best you'd need to add a bunch of type guards in the body of the function that aren't really necessary. Consider if you had the types defined like `(arg0: X | Y, arg1: Z | W)`, but externally it's `(arg0: X, arg1: Z) | (arg0: Y, arg1: W)`. We know that X/W and Y/Z shouldn't happen, but Typescript doesn't know that so if we want the arguments to interact we'd have to add an extra type guard, which has real performance impact at runtime.
2. Using `any` means it's difficult to validate the logic internal to the function. Take the normalizer vs `extractByKey` line: if those were type `any` and I accidentally switched the args around in one of those function calls--like `key(preRenderState, props)`--Typescript wouldn't give a peep.

As to why I chose the types I did: (a) they're primitives and not very extensible, i.e. if the type is wrong Typescript is _going_ to complain; (b) I picked string in `useRefState` because then I could have an `initObj` type that _wasn't_ `string`, but could be _assigned_ to `string` if the initializer function was undefined.

If you think it's too confusing I could change it. I could make it generic (I didn't because I thought it would be a bigger burden on Typescript but (a) I'm not even sure that's true, and (b) clarity of code certainly trumps a second or two of build time); or I could change the concrete type used, maybe to something like a `unique symbol` so it's more obvious the value is supposed to be different from the other types.

modules/cactus-web/src/hooks/useControllableValue.ts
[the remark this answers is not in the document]
wilysword: That is correct.

modules/cactus-web/src/hooks/useRefState.ts
```diff
@@ -0,0 +1,87 @@
...
+  const [{ ref }, triggerRender] = useReducer(reduceRef, args, initializeRef)
+  if (!ref.setState) {
+    const reducer: Reducer<string, any> = isReducer(args) ? args[0] : genericReducer
+    ref.setState = (action) => {
+      const nextState = reducer(ref.current, action)
+      if (nextState !== ref.current) {
+        ref.current = nextState
+        triggerRender()
+      }
+      return nextState
```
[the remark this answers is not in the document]
wilysword: Yes.

modules/cactus-web/src/hooks/useRefState.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const useRefState = (...args: HookArgs): InnerState['ref'] => {
+  const [{ ref }, triggerRender] = useReducer(reduceRef, args, initializeRef)
+  if (!ref.setState) {
+    const reducer: Reducer<string, any> = isReducer(args) ? args[0] : genericReducer
+    ref.setState = (action) => {
+      const nextState = reducer(ref.current, action)
+      if (nextState !== ref.current) {
+        ref.current = nextState
+        triggerRender()
```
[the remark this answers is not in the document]
wilysword: Correct. React calls this a "reducer without action", where the next state is solely dependent on the previous state. It doesn't come up all _that_ often, but for example, you might implement a counter that way: `reducer = (x: number) => ++x `.

Dhalton:
Ok thank you for explaining and providing some examples. That makes it more clear to me. Definitely don't want more type checking overhead and I hadn't considered your point about `extractByKey`.

> EDIT: I could also specifically name the types, something like `type StateType = number`, and then use `StateType` everywhere I'm currently using number; it's kind of like a generic, but without the extra boilerplate & mental overhead. In fact, I think that's probably a really good idea...

I really like this idea too. For me it's really just about the readability. If we could abstract that out just a bit to sort of hide the internal type and make it seem like a generic when looking at the logic of these hooks, that would definitely help.

So at this point I think I'm pretty good on the two new hooks and I'll be diving into the actual DataGrid changes in a bit. That'll also probably give me a bit better understanding of `useControllableValue` too.

Also, super weird that `repay-scripts` doesn't work with index files. We should definitely get that fixed. Does it just not see them or what?

wilysword:
> Also, super weird that repay-scripts doesn't work with index files. We should definitely get that fixed. Does it just not see them or what?

It has a bit of logic where it reads imports to look for dependencies. Since imports don't have extensions, it tries out different extensions, and if it doesn't find a file it throws an error. It doesn't account for the fact that `import 'hooks'` could be `hooks.js` or `hooks/index.js`, though. I'm not sure why that bit of code exists, to be honest; it seems like something another tool should be doing for us, but then I don't really understand any of what goes on in there so maybe it's very necessary :shrug:.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -30,15 +32,45 @@ interface DataGridProps extends MarginProps {
...
+      const pagisort: Pagisort = { ...setPageState(identity), sort: newSortOptions }
+      calcPageState(pagisort)
+      onPagisort(pagisort)
+    }
+  }
+
+  const updatePageState = (action: PageStateAction, raiseEvent = false) => {
+    const newPageOptions = setPageState(action)
+    if (raiseEvent && newPageOptions !== pageState) {
+      onPageChange?.(newPageOptions)
```
Dhalton, line 66: I know of the `?` syntax but I don't think I've ever seen it used to call a function like this. Pretty sure I've only ever used it for checking if some object exists before trying to access a key. I've _wanted_ to do this in the past but didn't know it was possible. When `func?()` didn't work I just stopped there. I'll have to keep this in the back of my mind.

modules/cactus-web/src/DataGrid/Pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export interface ExtendedPaginationProps
+  extends Omit<PaginationProps, 'currentPage' | 'pageCount'>,
+    PageStateProps {}
+
+export interface ExtendedPrevNextProps extends PrevNextProps, PageStateProps {}
+
+export const calcPageState = (p: PageState): PaginationOptions => {
+  // TS appeasement: comparison operators work perfectly fine with undefined.
+  const r = p as Required<PageState>
```
Dhalton, line 22: Can we give these `r` and `p` variables actual names? Obviously I can infer that they're the page state object, but it does add just that extra bit of effort to understand imo. I'm hesitant to ask for it since I think it might make the rest of the comparisons in this function harder to read, though...I kind of get the sense that that's part of the reason you gave them single-character names to begin with.

Maybe we could give `p` a real name, then just pull the attributes out like this:
`let { pageCount, currentPage, ... } = pageState as Required<PageState>`
and then return a new object containing the updated values at the end. What would you think about that?

modules/cactus-web/src/DataGrid/Pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+    delete props[key] // Remove from props so the rest can be safely forwarded.
+  }
+  const combinedState = interleave(pageState, controlledProps)
+  if (!combinedState.currentPage) {
+    controlledProps.currentPage = combinedState.currentPage = props.initialPage || 1
+  }
+  delete props.initialPage
+
+  React.useEffect(() => {
+    updatePageState(controlledProps)
```
Dhalton, line 75: If I'm understanding correctly, this will call `updatePageState` on every render? And that's what we want? I would think we'd only want to do that if something changes...I guess ultimately it'll get run through `pageStateReducer`, which will only actually update the state if something changes, is that kind of the idea? Idk if I'm fully following, maybe you can provide some clarification

Dhalton (commented):
I'm getting there...I've been reviewing PRs for like 2 days straight now and I feel like I'm losing all cognitive ability I have left at this point. I'm gonna leave these few comments and come back to this with fresh eyes in the morning. Should be able to finish up then.

modules/cactus-web/src/DataGrid/DataGrid.tsx
Dhalton: This one doesn't really require much on your part, kind of just talking to myself

modules/cactus-web/src/DataGrid/Pagination.tsx
wilysword: Yes, it will call on every render, but as you suppose, as long as nothing changes in the reducer it will not trigger a re-render...thus the notable lack of infinite loops. We _could_ put in a dependency array, but (at least when I started this) it would have been worse for performance. I figure the performance impact of a `useEffect` hook is like this:
1. Creating the function to be run; happens on every render, regardless of whether the function will be run or not.
2. Creating the dependency array (if there is one); happens every render.
3. Comparing the new array with the old one.
4. Calling the function.

In other words, as long as step 4 is reasonably expected to be less expensive than steps 2 & 3, it's better to omit the array. (Of course, I don't actually know the React internals, nor how to test the performance difference, so I could be wrong about that.)

When I first wrote the reducer it was just a loop over the five state fields, pretty much, so I figured that would be about the same impact as the dependency array; however now that I stuck in the `interleave` call I should probably adjust this. It may be less confusing for readers anyway.

modules/cactus-web/src/DataGrid/Pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const DataGridPrevNext: React.FC<ExtendedPrevNextProps> = ({ onNavigate, ...props }) => {
+  const [pageState, updatePageState] = usePageState(props)
+  return (
+    <PrevNext
+      disablePrev={pageState.currentPage === 1}
+      onNavigate={(direction: 'prev' | 'next'): void => {
+        updatePageState(direction === 'prev' ? decrementPage : incrementPage, true)
+        onNavigate?.(direction)
+      }}
```
daniloPenaR, line 117: what about 
```javascript 
const handleOnNavigate = (direction: 'prev' | 'next') => {
    updatePageState(direction === 'prev' ? decrementPage : incrementPage, true)
    onNavigate?.(direction)
  }
  return (
    <PrevNext
      disablePrev={pageState.currentPage === 1}
      onNavigate={handleOnNavigate}
      {...props}
    />
  )
```

daniloPenaR (commented):
I spent some timeboxes from my last days just reading and checking this...
I admire what you did here. The changes in the stories were helpful to understand how this is simplifying the implementation of this component. 
I think that those new hooks deserve some paragraphs on the documentation page if they are intended to be exported like `useScrollTrap` or `usePopup` (and this could help understand the whole complexity behind'em)

I'll spend more time exploring this goldmine and following your threads with Dhalton in this PR.

modules/cactus-web/src/DataGrid/Pagination.tsx
Dhalton: Your reasoning for why you wrote it this way makes sense to me. I don't know the React internals either obviously but I think that's probably a fair assumption in terms of `useEffect`'s performance impact. Even with `interleave`, it only gets called when there are state changes so I feel like your line of thinking may still be accurate.

In the case that there are no changes to state, all it's going to do is return and skip the interleaving `calcPageState` calls altogether. In the case that there _are_ changes to state, well it'd call `interleave` regardless of whether you included a dependency array here or not, unless you're thinking of a refactor that would somehow get rid of that.

I will say that reworking this to add a dependency array _might_ make it a tad bit more readable, but I'm leaving it up to you; I have a better understanding of it now so thanks for giving me some context.

modules/cactus-web/src/DataGrid/DataGrid.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { fireEvent } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
```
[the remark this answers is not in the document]
wilysword: It doesn't actually change anything in these tests, but it's more user-friendly overall so it should probably be used in all our tests for greater consistency; since I was already touching this file I decided to convert it.

modules/cactus-web/src/DataGrid/DataGrid.tsx
```diff
@@ -30,15 +32,45 @@ interface DataGridProps extends MarginProps {
...
+    pageStateReducer,
+    initialPageState
+  )
+
+  const onSort = (newSortOptions: SortOption[]) => {
+    newSortOptions = setSort(newSortOptions)
+    onSortProp?.(newSortOptions)
+    if (onPagisort) {
+      const pagisort: Pagisort = { ...setPageState(identity), sort: newSortOptions }
+      calcPageState(pagisort)
```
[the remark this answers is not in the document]
wilysword: Currently it doesn't return a new object, it returns the same object that is passed in after modifying it in-place. Dhalton made a suggestion along those lines, though, so I'm not sure if I'll change that or not. If I _do_ change it to return a new object, I'll be sure to fix this.

modules/cactus-web/src/DataGrid/Pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return r
+}
+
+const assignIfUndefined = (dest: unknown, src: unknown) => dest ?? src
+
+const interleave = (state: PageState, updates: PageState) => {
+  // To get the correct derived/implied values from the updates, we do the
+  // calculations twice: first with updates + pageSize, then combined with state.
+  const combinedState = calcPageState({ pageSize: state.pageSize, ...updates })
+  return calcPageState(assignWith(combinedState, state, assignIfUndefined))
```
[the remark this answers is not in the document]
wilysword: https://lodash.com/docs/4.17.15#assignWith
Basically it does the same thing as `Object.assign`, but it takes a callback that allows you to modify values. In this case, it only assigns a new value if the current one is `undefined`:
```
const x = { one: 'uno' }
const y = { one: 'ichi', two: 'ni' }
const z = assignWith(x, y, (a, b) => a ?? b)
// z = { one: 'uno', two: 'ni' }
```

modules/cactus-web/src/DataGrid/Pagination.tsx
wilysword: Not quite sure what you're asking here. Are you saying you don't like having the handler defined inline? Or that it's unclear?

modules/cactus-web/src/DataGrid/DataGrid.tsx
wilysword: Yeah, it looks pretty weird. It can also be used for indexing: `object?.[key]`.

modules/cactus-web/src/DataGrid/Pagination.tsx
wilysword: I kind of see what you mean, but I'm hesitant to create a new object considering it gets called at least twice on every render. (React creates hundreds if not thousands of objects on most renders, why am I so weird about this? Even I don't quite know.)

modules/cactus-web/src/DataGrid/Pagination.tsx
wilysword: Not quite. The `interleave` call is only avoided if the `action` produces the _same object_ as the current state (`===`); since `controlledProps` is a new object created on every render, that comparison is guaranteed to be false.

modules/cactus-web/src/DataGrid/Pagination.tsx
Dhalton: But wouldn't it effectively be the same number of objects being created, since you could remove the creation of `r`? It's not a huge deal, obviously I can figure out what's going on here. Just spitballin'

Dhalton: Or does React not actually create a new object since you're not spreading it like this `const r = { ...p } as Required<PageState>`....ahhh, yeah, I bet that's the case

wilysword: `r` isn't creating a new object, it's casting the old object to a new type.

modules/cactus-web/src/DataGrid/Pagination.tsx
Dhalton: Ohhhh okay I see. Yeah in that case maybe this is impacting performance more than it would if there were a dependency array. It's probably negligible (I have yet to UAT), but I wouldn't be opposed to it if you wanted to try to refactor this

modules/cactus-web/src/DataGrid/Pagination.tsx
Dhalton: Yeah I gotcha. Meh, it's not a big deal. Honestly I think my tired brain yesterday was just slow to process. I'm fine keeping this as is to prevent one more object being created 😉

Dhalton:
Ok I've looked at everything and done my UAT at this point. I think you're working on some updates so I'll hold off on putting my final approval on here. Just lmk when you're ready for me to have another look

wilysword:
Well, I've made some changes although they might not be quite what people had in mind with their comments. I hope the code's at least a little clearer now.

Dhalton (approved):

daniloPenaR (approved):

PR #792, opened 2022-09-05
CACTUS-1017 :: Update docs publishing to use a deploy-key

daniloPenaR:
[CACTUS-1017](https://repayonline.atlassian.net/browse/CACTUS-1017)


maxrcollins (approved):
It works! Congrats :)

Do you know if your token will ever go bad? Like 50 years from now when you retire from Repay and your accounts get deactivated?

Dhalton (commented):
Can we please squash all of these commits down to a single commit? No need to merge all of these in

daniloPenaR:
> It works! Congrats :)
> 
> Do you know if your token will ever go bad? Like 50 years from now when you retire from Repay and your accounts get deactivated?

I don't think so. The `SSH_PRIVATE_KEY` token it's just a randomly generated password. This secret works as a decryption key for the encrypted file `cactus_auth.enc` that is, the SSH private key. The public key is now stored as a deploy key [here](https://github.com/repaygithub/cactus/settings/keys). So what I´ve learned about this at this point is that those keys work only for one server which is fine for our case. If we need to change or update those keys in the future, the owner of that task only needs to create an SSH key locally, encrypt the private one with a password, update the `SSH_PRIVATE_KEY`secret with that new password, and update the Deploy key with the new public SSH key and push or rewrite the encrypted file. 

I'm preserving the current SSH private, public, and password strings locally until this was merged and executed successfully on master. So I can share those creds with you or any other Dev via onetimesecret if you want.

Dhalton (approved):

PR #793, opened 2022-09-06
CACTUS-1016: fix popup overlap with sticky positioning

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1016

You can UAT this on the `DataGrid` story.

This is kind of a hacky solution; if we switch to using portals, this whole commit can just be reverted.
- The MutationObserver fixes the issue of popups appearing underneath the next row(s).
- The scroll parent change fixes the issue of popups being hidden by the table's scroll container. (Not mentioned in the ticket, but can easily be seen by opening the drop-down on the last row; somehow doesn't appear to happen on Chrome, though? It does on Firefox.)

### DIFF
--- a/modules/cactus-web/src/Table/Table.tsx
+++ b/modules/cactus-web/src/Table/Table.tsx
@@ -10,6 +10,7 @@ import styled, {
 } from 'styled-components'
 import { margin, MarginProps, width, WidthProps } from 'styled-system'
 
+import { isIE } from '../helpers/constants'
 import { extractMargins } from '../helpers/omit'
 import { useMergedRefs } from '../helpers/react'
 import { border, boxShadow, media, radius, textStyle } from '../helpers/theme'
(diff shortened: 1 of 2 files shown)


modules/cactus-web/src/Table/Table.tsx
```diff
@@ -124,6 +125,26 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
         }
       }
     })
+
+    // Can be removed if we switch popups back to portals.
+    React.useEffect(() => {
+      if (sticky === 'right' && !isIE && tableRef.current) {
```
maxrcollins, line 131: why only if `sticky === 'right'`? Wont we have this issue now if `sticky === 'left'`?

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -124,6 +125,26 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
+              if (cell) {
+                cell.style.zIndex = target.getAttribute('aria-expanded') === 'true' ? '1' : ''
+              }
+            }
+          }
+        })
+        observer.observe(tableRef.current, { attributeFilter: ['aria-expanded'], subtree: true })
+        return () => observer.disconnect()
+      }
+    }, [sticky])
```
maxrcollins, line 146: I keep thinking that this can't possibly be the easiest way to do this. What's wrong with just these three lines? It works for me:
image

maxrcollins:
Thanks for updating. If you think it's a good idea, you could make some edits to make sure it doesn't happen in IE11 or when the column is not position: sticky, like you had been checking for before.

```
${(p) => if (p.sticky === 'right' && !isIE) {
  return css` 
     &:focus-within { 
        z-index: 1;
     }
  `
}}
```

Dhalton (approved):
LGTM, can confirm that it doesn't affect IE

maxrcollins:
> Don't really see the point, since IE will just ignore it because it doesn't support `:focus-within`. As for the sticky condition, I don't think there's really any downside to changing the z-index regardless, is there?

You're right. I don't think there's a downside. You just know this repo and potential repercussions better than I so I wanted to double check

maxrcollins (approved):
Looks good

PR #794, opened 2022-09-07
CACTUS-1022: apply cell alignment to sortable columns

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1022

Even easier than I expected. Basically, the `text-align` property only applies to inline elements, so we just make the button inline.


daniloPenaR (approved):
LGTM

maxrcollins (approved):
Solid 👍

PR #795, opened 2022-09-12
CACTUS-998: support Flex gap property; add `Flex.Item`

wilysword:
UAT: https://repayonline.atlassian.net/browse/CACTUS-998

I also moved the `space-evenly` fix into the styled-system code, and removed some properties that have no effect on flex layouts (`justifyItems` & `justifySelf`).

The `Flex.Item` component is just because I've come across quite a few situations where I wanted to set a flex property, but the component I was using didn't support them.

Because `Modal` and `Footer` also utilize the `flexContainer` styles, I added the gap workaround to them as well.

### DIFF
--- a/modules/cactus-web/src/Flex/Flex.tsx
+++ b/modules/cactus-web/src/Flex/Flex.tsx
@@ -1,12 +1,17 @@
-import PropTypes from 'prop-types'
-import styled, { css } from 'styled-components'
-import { flexbox, FlexboxProps } from 'styled-system'
+import { CactusTheme } from '@repay/cactus-theme'
+import styled, { StyledComponentBase } from 'styled-components'
+import { compose } from 'styled-system'
 
 import { Box, BoxProps } from '../Box/Box'
-import { isIE } from '../helpers/constants'
 import { getOmittableProps } from '../helpers/omit'
+import { flexContainer, flexItem, FlexItemProps, FlexProps, gapWorkaround } from '../helpers/styled'
 
-interface FlexBoxProps extends BoxProps, Omit<FlexboxProps, 'justifySelf'> {}
+interface FlexBoxProps extends BoxProps, FlexProps, FlexItemProps {}
+
+interface FlexComponent extends StyledComponentBase<'div', CactusTheme, FlexBoxProps> {
+  Item: StyledComponentBase<'div', CactusTheme, FlexItemProps>
+  supportsGap: boolean
+}
 
 export const justifyOptions = [
   'unset',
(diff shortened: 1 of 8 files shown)


modules/cactus-web/src/Flex/Flex.tsx
```diff
@@ -20,42 +25,21 @@ export const justifyOptions = [
...
-  alignItems: styledProp('unset', 'flex-start', 'flex-end', 'center', 'baseline', 'stretch'),
-  alignSelf: styledProp('unset', 'flex-start', 'flex-end', 'center', 'baseline', 'stretch'),
-  flexWrap: styledProp('unset', 'inherit', 'wrap', 'nowrap', 'wrap-reverse'),
-  flexDirection: styledProp('unset', 'inherit', 'row', 'row-reverse', 'column', 'column-reverse'),
-}
+  flex-wrap: wrap;
+  ${gapWorkaround}
+  ${compose(flexContainer, flexItem)}
+` as any
+Flex.supportsGap = !gapWorkaround
```
daniloPenaR, line 37: I am not sure if I understand Why you did this

modules/cactus-web/src/helpers/styled.ts
```diff
@@ -38,66 +41,109 @@ export const styledProp = PropTypes.oneOfType([cssVal, PropTypes.arrayOf(cssVal)
...
+}
+
+// Keeping this separate from `flexContainer` styles because it applies
+// to the element's children, not the element itself.
+export const gapWorkaround: SS.styleFn | undefined = (function () {
+  if (typeof CSS !== 'undefined' && CSS.supports?.('gap', '1px')) return
+  const calcGap = (value: any, scale: SS.Scale | undefined) => {
+    try {
+      value = scale?.[value] ?? value
+    } catch {}
```
daniloPenaR, line 60: Why use an empty `catch`

daniloPenaR (commented):
Seems good, gonna play with this a bit. Got a couple of questions

modules/cactus-web/src/Flex/Flex.tsx
wilysword: The default gap workaround is pretty good, but there might be use cases where it's insufficient. This variable gives devs a way to tell if they need to add extra styling for IE:
```
<Flex gap={4} mt={Flex.supportsGap ? 0 : -3}>
```

modules/cactus-web/src/helpers/styled.ts
wilysword: Because nothing needs to be done if there's an error, we can just continue on with the original `value`.

daniloPenaR (approved):

modules/cactus-web/src/helpers/styled.ts
```diff
@@ -38,66 +41,109 @@ export const styledProp = PropTypes.oneOfType([cssVal, PropTypes.arrayOf(cssVal)
...
-    if (!keys.includes(key)) {
-      picked.push((styles as any)[key])
+  if (!styles.config) return styles
+  return SS.createParser(omit(styles.config, keys))
+}
+
+// Keeping this separate from `flexContainer` styles because it applies
+// to the element's children, not the element itself.
+export const gapWorkaround: SS.styleFn | undefined = (function () {
+  if (typeof CSS !== 'undefined' && CSS.supports?.('gap', '1px')) return
```
Dhalton, line 56: Referencing `CSS` like this will crash the page in IE11. I discovered that while testing channels in IE the other day. Storybook itself is actually not broken in IE; this is what's breaking it.

wilysword: Are you positive? I was getting an error where storybook wouldn't even load, until I added the `typeof CSS !== undefined` bit, and then it started loading so I assumed it was fine.

Dhalton: Hmm, well I just tested it out and Storybook is loading fine for me, even with these changes. But I do know that when I was testing on Channels, there was a very similar check (`CSS !== undefined`) and _that_ was crashing the app. Maybe it's the `typeof`?

Dhalton: I figured it out. It's the fact that you're checking `typeof CSS` and not the value of `CSS`. I'm gonna make a change to the way I fixed this in channels. But this should be fine here, my bad

Dhalton (approved):

PR #797, opened 2022-09-14
CACTUS-1009 Snyk fixes

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1009

UAT is just doing anything you can think of. Focus on stuff that's not covered by the automated stuff, of course, but it won't hurt if there's overlap.

I upgraded everything possible among transient dependencies, but I suspect we'll still have the same vulnerabilities as before since they're through storybook, which we can't upgrade right now, and gatsby/lerna which are already at the latest versions.

### DIFF
--- a/actions/integration-tests/Dockerfile
+++ b/actions/integration-tests/Dockerfile
@@ -6,7 +6,9 @@ COPY entrypoint.sh /entrypoint.sh
 RUN apt-get update && apt-get install -y \
   wget \
   unzip \
-  procps
+  procps \
+# These are existing packages with security updates.
+  libc6 zlib1g libpcre2-8-0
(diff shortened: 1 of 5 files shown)


wilysword:
The testcafe/browserstack is separate because at first I thought it was just an incompatibility; but since searching turned up no other reports of people having problems with that version of `testcafe-browser-provider-browserstack`, I decided to try some other stuff, and it turns out we just had to change how the version is specified.

maxrcollins (approved):
Looks good to me. I tested a bunch of stuff and nothing seems broken

daniloPenaR (approved):
Looks good, I tested and ran almost every command I could. I cannot run `yarn test:build` commands on the M1 machine, so I did it from the intel one.

PR #799, opened 2022-09-28
CACTUS-972 Grid is grid

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-972

As you can see, most of this is IE-specific crap. In the end, though, I was able to get pretty good support with minimal polyfills. There might be a slight change in behavior with `justifyItems`: the old behavior in IE was to make every `Grid.Item` `display: flex` and then map the item justify to flex justify. Now it's directly supported so it doesn't change to `display: flex`; I don't _think_ this will be a big deal, though, because if it ever came up at all they would have had to make it look good in other browsers as well, so now it should just display in IE the same way it does in other browsers.

With the docs, I am so tired of how flaky the automatic props table is. I don't think I've had a single time where it "just worked", and frankly, I no longer think it is in any way worth it. Maintaining the docs this way is a little more work, but much more flexible: we can show more relevant information, and we don't have to try to put it all in one-size-fits-all tables.

The Flex thing is more of a follow-up to the changes I made in [CACTUS-998](https://repayonline.atlassian.net/browse/CACTUS-998). I knew there was a bug with how it handled `display`, but since it's a pretty unusual use case and I couldn't think of a solution I just left it. While I was working on this stuff, I finally realized I could just make two different versions of the container styles, one generic and one meant specifically for components that don't support the `display` prop.


modules/cactus-web/src/Grid/Grid.tsx
```diff
@@ -1,134 +1,331 @@
...
+interface GridComponent extends StyledComponentBase<'div', CactusTheme, GridBoxProps> {
+  Item: StyledComponentBase<'div', CactusTheme, GridItemProps>
+  supportsGrid: boolean
+  supportsGap: boolean
+}
 
-  @media (min-width: 1201px) {
-    width: ${(p) => (p.large ? calculateFlexItemSize(p.large) : undefined)};
-  }
+const gridItemStyles = (function () {
```
Dhalton, line 79: Man, this function has been taking me a while to understand. Sometimes it helps for me to write out my thought process, and of course you can verify if I'm on the right track, so let me see if I'm getting it...

It seems like the main purpose is to convert grid item prop syntax that wouldn't be acceptable in IE to something IE can understand.

So for (a really simple) example, `rowStart` will get a parser (from `getStartParser`) that will have access to the corresponding browser's `startProp`. In IE's case, that would be `-ms-grid-row`. The parser that gets returned in this `rowStart` example basically just needs to convert what's passed through the `rowStart` prop to IE's equivalent prop name. So it would return `-ms-grid-row: String(<rowStart value>)`.

Then for a more complicated example...I'll try to piece together the logic for the `row` prop. `row` should define both the start and end, hence why it calls `getCombinedStartEndParser`, I presume. This parser will have access to both the `startProp` (again, in IE's case, this will be `-ms-grid-row`) and the `spanProp`, which in IE's case would be `-ms-grid-row-span`. The parser will receive either a number or a string that was passed to the `row` prop.
- If it's a number, we know there's no delimiter (and no `end`), so it calls `getStartSpan`, which would just return `{ '-ms-grid-row': String(<row value>) }`
- If it's a string, there could be a start and an end (let's say it's `1 / 5`), so it splits on the ` / ` and calls `getStartSpan`, which would return something like this: `{ '-ms-grid-row': '1', '-ms-grid-row-span': '4' }`, since you get the span by subtracting the end from the start.

Now that I've written it out, I think it's a little more clear to me. I've been trying to think of ways this could potentially be simplified, but I'm coming up short. I'll keep thinking on it

wilysword: It could potentially be simplified if we didn't have to support IE 😉.

modules/cactus-web/src/Grid/Grid.tsx
```diff
@@ -1,134 +1,331 @@
...
 
+import Flex, { FlexBoxProps } from '../Flex/Flex'
+import { isIE } from '../helpers/constants'
 import { getOmittableProps } from '../helpers/omit'
+import { useMergedRefs } from '../helpers/react'
+import { FlexItemProps } from '../helpers/styled'
+import { screenSizes, useScreenSize } from '../ScreenSizeProvider/ScreenSizeProvider'
 
-const GUTTER_WIDTH = 16
+const DEFAULT_GAP = 4
```
Dhalton, line 15: I must not totally understand how gap works, because this looks to me like it changes the default spacing. Can you explain?

wilysword: The `GUTTER_WIDTH` is a pixel value, `DEFAULT_GAP` is an index from `theme.space`, like when we say `margin={4}`; with our current theme settings `4` -> "16px".

modules/cactus-web/src/Grid/Grid.tsx
Dhalton: If only. I gotta give you props for figuring this conversion out. It's definitely clear that it requires a deep understanding of grid and the differences between grid in moderns browsers vs IE

modules/cactus-web/src/Grid/Grid.tsx
Dhalton: Ah, gotcha. Makes perfect sense, thanks

Dhalton (approved):
I was able to run the story in IE on browserstack just fine. LGTM works well

modules/cactus-web/src/Grid/Grid.tsx
maxrcollins: I've been looking at this for a couple hours and am still struggling lol.  I will...have questions eventually.

maxrcollins (approved):
I honestly don't fully understand this but I think I understand *enough* and it works so 👍

PR #800, opened 2022-09-30
CACTUS-975 inline styles

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-975?focusedCommentId=149099

I put my results & Recommendations in JIRA (I like the formatting there for tables a little better), but we can also talk about it here. There is no intention of merging this PR, it's just to show what I tested and how I set up the test cases.

In short, though, I think that we should definitely switch to inline styles, since there's no performance hit for doing so (well, there is on resize, but that happens orders of magnitude less often than than regular rendering so it's much less significant) and we get a couple significant benefits:
- Much more straightforward style overrides.
- Fewer CSS classes and stylesheets.

Let me know if you agree/disagree, if you think there are any good options that I didn't test, or any tests that I didn't do but should have.


modules/cactus-web/src/style-testing.jsx
```diff
@@ -0,0 +1,218 @@
...
+  return { style: props.style ? { ...props.style, ...baseStyles } : baseStyles }
+}
+
+export const WithDefaultParserAndHook = styled.div.attrs(useDefaultParserAndHook)`
+  width: 50px;
+  height: 50px;
+  background: red;
+`
+
+const config = margin.config
```
Dhalton, line 71: So if we went with the custom parser solution, we'd end up taking configs from several SS parsers, right? This would end up being something like:
`const config = { ...margin.config, ...padding.config, ...width.config, ... }`

We'd basically combine all configs into one to be used in the parser?

modules/cactus-web/src/style-testing.jsx
```diff
@@ -0,0 +1,218 @@
...
+const useCustomParserAndHook = (props) => {
+  const currentSize = useScreenSize()
+  const { style: styleProp, theme } = props
+  let style = styleProp
+  let sizeKeys = false
+  for (const key in props) {
+    const processor = config[key]
+    if (!processor) continue
```
Dhalton, line 79: If there's no processor for the prop, does it essentially just get ignored?

Dhalton (commented):
I still want to play around with this more tomorrow, but I'll leave some of my thoughts that I have up to this point.

I'm on board with the inline styles. I think the benefit is clear. I'm still on the fence about how we do it, though:
- I'm wondering what exactly you're doing differently in your custom parser that gets us that increase in performance. I can see that you're still using the processor from `styled-system`, so I have to assume that the performance gains are coming from the processing you're doing afterward, which I'm still trying to understand.
- My concern is whether the (seemingly?) miniscule performance improvement is worth the overhead that we incur by implementing our own parser. I'm sure this code is clear to you, but I'm still struggling to understand it, and I worry about how tough it may be for me if I had to make any changes to that code in the future. The `styled-system` devs already wrote those parsers, and their performance has been good enough for us up until this point. Do we need to re-implement them for better performance at the cost of clarity and more code we have to maintain? Maybe we could live with slightly less performant code that we can all understand and have an easier time maintaining.

If you think the performance gain is really significant, though, I could be persuaded. But I want to hear your thoughts on potentially just using the default parsers and if you'd be willing to meet in the middle.

modules/cactus-web/src/style-testing.jsx
wilysword: Probably something more like using `compose` from `styled-system`, though that's basically what _it_ does behind the scenes:
```
const config = compose(margin, padding, width).config
```

modules/cactus-web/src/style-testing.jsx
wilysword: It gets ignored as far as styles are concerned, though of course it still gets passed to the base component. The default parser does the same thing.

wilysword:
I more or less agree with you about the maintenance, that's the main reason I didn't write the default parser off entirely. Although I'd prefer writing a custom parser, I'm not opposed to using the default; the custom parser is to me maybe 55% pros, 45% cons.

The main thing I dislike about the default parser isn't the performance, it's the way the media queries are used. I don't like that they don't export the function for turning breakpoints into media queries, meaning if they ever change it then our code will break. (However, we already deal with that in `cactus-theme`, and it's nothing a good unit test won't catch, so it's really more of a personal dislike than an actual reason to avoid it.)

As to the performance difference, I can't be entirely sure without fully profiling it. [This is their parser](https://github.com/styled-system/styled-system/blob/master/packages/core/src/index.js#L42), if you're interested. One difference is that they have to fully process all the responsive styles in order to create media queries, where we know we aren't writing queries so we can specifically target the current screen size. Beyond that, I think the performance difference is probably down to general inefficiency on their part. Take their `merge` function at the top of the file. The logic is pretty standard, but there's three needless inefficiencies:
1. It creates a new object for the result. That's not always a bad thing: mutating function arguments usually _is_ bad, in fact. In this case, however, it's an internal utility and the first argument is already an object they created, so there's really no reason they couldn't modify it.
2. It calls `assign` first, then loops over the keys. Coupled with (1), that means it's essentially doing triple the work, looping over three objects when it really only needs to loop over one.
3. They call `assign` and create a new object just to set a single property. Is the `=` key on their keyboard broken or something? `result[key] = assign(a[key], b[key])` would do exactly the same thing, and be both clearer and more computationally efficient. Also note that using `assign` here breaks their "commitment" to immutability: they won't modify `a`, but they _will_ modify the nested objects of `a`? Kind of arbitrary.

And they've got that kind of code in several places. I basically just don't like their parser; but you don't have to let that sway your opinion, because we don't have to look at it (much) in order to utilize it.

Dhalton:
Thanks for highlighting some of the inconsistencies & inefficiencies of their parser. That makes it more clear where the performance gains come from in your custom parser.

The fact that they don't export their breakpoint to media query logic is a fair point. Like you said, though, if we're already handling that in the theme then it's not a _huge_ problem...I also think the chances of them changing that logic are pretty slim, too, given that the last time that file was touched was ~3 years ago.

Speaking of the breakpoint to media query logic (and the likelihood of it changing) I want to make sure I know what function/logic you're referring to and the worry with it changing. I think what you're referencing is [this function](https://github.com/styled-system/styled-system/blob/master/packages/core/src/index.js#L31) and where it's used on line 62 (correct me if I'm wrong). That function itself seems unlikely to change I would think, at least in what it outputs...I don't know of another way to structure media queries, although I admit that there _could be_ other ways. So what ways could it be changed that would break our code? If they stopped putting the media queries in the `cache` object or changed the associated key they're stored under, maybe?

I'm leaning more toward the default parser, but I'm still trying to get a good read on the situation, so your thoughts and more context are appreciated.

On a totally different note, I was hoping you could help me make sense of some testing I was doing. I have your changes pulled down locally and I wanted to experiment with how this could improve our style overrides. I found an [override that needed lots of specificity](https://github.com/repaygithub/cactus/blob/master/modules/cactus-web/src/Table/Table.tsx#L336-L337) and I changed that styled component to be `styled.thead.attrs(useDefaultParserAndTheme)`. Then I removed some of the `&` selectors to see how we could cut down on the specificity issue. I was able to cut it down from `&&&&&` to `&&` before the styles stopped working, which is great. What I'm wondering is:
- Do you think that's expected? I was thinking I'd only need a single `&`, so I was a little confused.
- Is there more I could do in my experiment to improve the required specificity?

Just trying to experiment and get a better idea of what it'll look like when we start converting components.

wilysword:
That is the media query function I'm referring to, and agree it's unlikely to change. I'm not concerned about big changes in any case, but since we're doing string comparisons even a semantically meaningless change could break stuff: adding parentheses, removing the `screen and` part, adding or removing a space. It's not that I think it's likely, just the possibility, and the fact that we have to maintain our own copy of the logic, makes me uncomfortable. It is a tolerable discomfort, however.

> On a totally different note, I was hoping you could help me make sense of some testing I was doing.

Those table cell properties aren't really applicable to the inline styles issue. The table cell styles are static styles, which are overriding other static styles. The situation inline styles rectify is better exemplified by [FieldWrapper](https://github.com/repaygithub/cactus/blob/master/modules/cactus-web/src/FieldWrapper/FieldWrapper.tsx), which adds a default `margin-top` and then allows you to override it with props. Because the default margin is defined under a selector with specificity 2 (the `styled-components` class + `:first-child`, `:not` doesn't add to specificity), the overrides also have to be at least 2 (`&&`).

Dhalton:
Okay, fair enough. Well like I said I'm on board with inline styles. After talking with you and some experimentation, my vote is for the styled system parser + the `useScreenSize` hook in the `attrs` function. I'm not _completely_ opposed to a custom parser because you bring up some good points, I just think we could do without it at least for now while we make the initial switch and start getting used to this new method. Maybe the custom parser could be an enhancement we could make down the road after we're all a little more comfortable with this.

Anyway, that's all just my two cents. I'm curious to hear what others think.

daniloPenaR:
tbh, This is one of those implementations from you that it’s difficult for me to follow just when reading the code and testing it a couple of times so thanks for including that explanation in the ticket’s description and all the tests you made with the hook and the theme object. I think it’d be interesting to implement this way of using `attr` function. In the first time I started looking at your code I thought how crucial with be in the next months to start looking at the `styled-components` and `styled-system` libraries in deep for catching up on all what you’re proposing with the custom parser and the breaking changes in the theme object. In my opinion, we should start implementing all of this with the default parser as Dhalton suggested, and in the meantime start looking at that custom parser you’ve created. Although maybe a switch from default to custom could represent maybe an extra effort, how complex could be implementing a custom parser later after being working with attr a couple of days/weeks/months?

wilysword:
Just writing the parser isn't that hard; I mean, it's basically already written, we'd just need to write unit tests and debug it. The reasons for choosing the default parser instead of the custom one won't change with time, though, so I think if we start with the default parser, we may as well stick with it in the long run also.

wilysword:
Create implementation tickets: https://repayonline.atlassian.net/browse/CACTUS-1069 & https://repayonline.atlassian.net/browse/CACTUS-1070.

PR #801, opened 2022-09-30
CACTUS-1034 :: MenuBar Active Styles

Dhalton:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-1034

### DIFF
--- a/modules/cactus-web/src/MenuBar/MenuBar.story.tsx
+++ b/modules/cactus-web/src/MenuBar/MenuBar.story.tsx
@@ -42,12 +42,14 @@ type MBStory = Story<{
   variant?: 'light' | 'dark'
 }>
 
-export const BasicUsage: MBStory = ({ breadth, totalDepth, onClick }) => {
+export const BasicUsage: MBStory = ({ breadth, totalDepth }) => {
+  const [currentActive, setCurrentActive] = React.useState(-1)
   const makeList = (
     depth: number,
     props: any,
     ix: number,
-    Component: React.ComponentType<any> = MenuBar.List
+    Component: React.ComponentType<any> = MenuBar.List,
+    parentIX?: number
   ) => {
     const items = []
     for (let i = 0; i < breadth; i++) {
@@ -56,17 +58,25 @@ export const BasicUsage: MBStory = ({ breadth, totalDepth, onClick }) => {
       const useList = useStatic ? i % 3 === 0 : (Math.random() * depth) / 1.5 > 0.5
       const title = useStatic ? LABELS[i] : getLabel()
       if (useList) {
-        items.push(makeList(depth - 1, { title }, i))
+        items.push(makeList(depth - 1, { title }, i, undefined, parentIX ?? i))
       } else {
         items.push(
-          <MenuBar.Item key={i} onClick={onClick(`x: ${i}, y: ${totalDepth - depth}`)}>
+          <MenuBar.Item
+            key={i}
+            aria-current={i === currentActive}
+            onClick={() => {
+              const menuItemIndex = parentIX ?? i
+              setCurrentActive((current) => (current === menuItemIndex ? -1 : menuItemIndex))
+              console.log(`x: ${i}, y: ${totalDepth - depth}`)
+            }}
+          >
             {title}
           </MenuBar.Item>
         )
       }
     }
     return (
-      <Component key={ix} {...props}>
+      <Component key={ix} aria-current={ix === currentActive} {...props}>
         {items}
       </Component>
     )
@@ -76,12 +86,14 @@ export const BasicUsage: MBStory = ({ breadth, totalDepth, onClick }) => {
 }
 BasicUsage.args = { breadth: 8, totalDepth: 2 }
 
-export const MenuBarDark: MBStory = ({ breadth, totalDepth, variant, onClick }) => {
+export const MenuBarDark: MBStory = ({ breadth, totalDepth, variant }) => {
+  const [currentActive, setCurrentActive] = React.useState(-1)
   const makeList = (
     depth: number,
     props: any,
     ix: number,
-    Component: React.ComponentType<any> = MenuBar.List
+    Component: React.ComponentType<any> = MenuBar.List,
+    parentIX?: number
   ) => {
     const items = []
     for (let i = 0; i < breadth; i++) {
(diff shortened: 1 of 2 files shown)


modules/cactus-web/src/MenuBar/MenuBar.story.tsx
```diff
@@ -56,17 +58,25 @@ export const BasicUsage: MBStory = ({ breadth, totalDepth, onClick }) => {
...
           </MenuBar.Item>
         )
       }
     }
     return (
-      <Component key={ix} {...props}>
+      <Component key={ix} aria-current={ix === currentActive} {...props}>
         {items}
       </Component>
     )
```
wilysword, line 82: It's weird that the menu changes every time you click. Honestly I'm not sure if it's worth trying to fix, though. Maybe something like
```
// At least this part could be broken into a separate hook and reused in both stories.
const menuItems = React.useMemo(() => {
  const getItems = (depth) => {
    const items = []
    for (let i = 0; i < breadth; i++) {
      // We need the default to be 100% reproducible for storyshots.
      const useStatic = depth === totalDepth && i < 8
      const useList = useStatic ? i % 3 === 0 : (Math.random() * depth) / 1.5 > 0.5
      const title = useStatic ? LABELS[i] : getLabel()
      // The first item in the list is the list title.
      items.push(useList ? [title, ...getItems(depth - 1)] : title)
    }
    return items
  }
  return getItems(totalDepth)
}, [breadth, totalDepth])
let parentIndex = -1
const makeMenus = (item, index, array) => {
  // Top-level call, set the parent index to propagate down.
  if (array === menuItems) parentIndex = index
  const props = { key: index, 'aria-current': index === currentActive }
  if (Array.isArray(item)) {
    const [title, ...children] = item
    return (
      <MenuBar.List {...props} title={title}>
        {children.map(makeMenus)}
      </MenuBar.List>
    )
  } else {
    return (<MenuBar.item {...props} onClick={youCanFigureThisPartOut}>{item}</MenuBar.Item>)
  }
}
return <MenuBar>{menuItems.map(makeMenus)}</MenuBar>
```

wilysword (commented):
In the story if you're not going to use the `onClick` action, might as well remove it (the `argTypes` bit and the `ActionWrap` bit).

Functionally looks fine, the story code is a little funky though. If you don't think it's worth fixing though, I'll still approve.

Dhalton:
@wilysword Yeah the story code is definitely wonky. It was difficult to get it to work even how it does now. I'm gonna try your suggestions though

modules/cactus-web/src/MenuBar/MenuBar.story.tsx
```diff
@@ -56,17 +58,25 @@ export const BasicUsage: MBStory = ({ breadth, totalDepth, onClick }) => {
...
+        items.push(makeList(depth - 1, { title }, i, undefined, parentIX ?? i))
       } else {
         items.push(
-          <MenuBar.Item key={i} onClick={onClick(`x: ${i}, y: ${totalDepth - depth}`)}>
+          <MenuBar.Item
+            key={i}
+            aria-current={i === currentActive}
+            onClick={() => {
+              const menuItemIndex = parentIX ?? i
+              setCurrentActive((current) => (current === menuItemIndex ? -1 : menuItemIndex))
```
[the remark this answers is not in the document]
Dhalton: You mean why remove the currently active item if you click the same one twice? I just felt like you should be able to remove the styles after you added them, although that's not how it'd work in the real world. Want me to get rid of that logic?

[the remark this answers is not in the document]
Dhalton: Yeah, that makes sense. I'll see what I can do

modules/cactus-web/src/MenuBar/MenuBar.story.tsx
Dhalton: I took this suggestion for the most part, but I had to tweak it a little bit to get it to work. `parentIndex` was always 7 (the last index) when I was clicking on menu items, so I passed the parent index to `makeMenus`, which now returns a function. That way, the function that `makeMenus` returns always has a reference to the correct parent index, if applicable.

It works, but I feel like it's still a bit confusing. Let me know if you had something different in mind.

modules/cactus-web/src/MenuBar/MenuBar.story.tsx
```diff
@@ -32,101 +32,112 @@ const getLabel = () => LABELS[Math.floor(Math.random() * LABELS.length)]
...
-      // We need the default to be 100% reproducible for storyshots.
-      const useStatic = depth === totalDepth && i < 8
-      const useList = useStatic ? i % 3 === 0 : (Math.random() * depth) / 1.5 > 0.5
-      const title = useStatic ? LABELS[i] : getLabel()
-      if (useList) {
-        items.push(makeList(depth - 1, { title }, i, undefined, parentIX ?? i))
+  const menuItems = useMenuItems(breadth, totalDepth)
+  let parentIndex = -1
+  const makeMenus =
+    (parentIX?: number) =>
```
wilysword, line 67: Not sure why you need this extra function. I get why you can't reference it directly in the `onClick` handler, but the `nextActiveIndex` variable is something that could be set outside the `onClick`, and unlike `parentIndex`, it wouldn't change.
```
const isTopLevel = (array === menuItems)
if (isTopLevel) parentIndex = index // changes, can't reference directly in `onClick`
const topLevelIndex = parentIndex // now we have another copy that _won't_ change

...
onClick={() => setCurrentActive((current) => {
  return isTopLevel && current === topLevelIndex ? -1 : topLevelIndex
})}
```

Also, it seems like this whole function is the same between the two stories, so it could also be included in the hook, along with the state.
```
  return menuItems.map(makeMenus)
}
export const MenuBarDark: MBStory = ({ breadth, totalDepth, variant }) => {
  const menuItems = useMenuItems(breadth, totalDepth)
  return <MenuBar variant={variant}>{menuItems}</MenuBar>
}
```

Dhalton: Thanks for explaining

wilysword (approved):
We probably spent more time on that than it was worth...

daniloPenaR (approved):
Kudos for the new logic in the Story. Awesome.

PR #802, opened 2022-10-04
CACTUS-1043: add italic placeholder styles to TextArea

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1043

I just refactored so all the common styles, including the italic, are in one place. I was going to add the min-width override that I talked about, but with the extra div there I don't think it's possible. With `width` you can just change from "auto" to "100%", but if you set `min-width: 100%`, that's going to change the actual width, not the min-width; and if you set it to the same value as the outer div, `min-width: 50%` would effectively be 25% instead. We'll just have to defer that change to https://repayonline.atlassian.net/browse/CACTUS-1055.

### DIFF
--- a/modules/cactus-web/src/TextInput/TextInput.tsx
+++ b/modules/cactus-web/src/TextInput/TextInput.tsx
@@ -8,27 +8,22 @@ import defaultTheme, {
 } from '@repay/cactus-theme'
 import PropTypes from 'prop-types'
 import React from 'react'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 import { compose, margin, MarginProps, width, WidthProps } from 'styled-system'
 
 import { omitMargins } from '../helpers/omit'
-import { getStatusStyles, Status, StatusPropType } from '../helpers/status'
+import { getStatusStyles, StatusProps, StatusPropType } from '../helpers/status'
 
 type TextStyleKey = keyof TextStyleCollection
 export const textStyles = Object.keys(defaultTheme.textStyles) as TextStyleKey[]
 
 type InputElementProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, 'width'>
-export interface TextInputProps extends InputElementProps, MarginProps, WidthProps {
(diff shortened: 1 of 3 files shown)


modules/cactus-web/src/helpers/status.ts
```diff
@@ -7,13 +7,13 @@ export type Status = typeof statuses[number]
 type StatusBackground = 'successLight' | 'warningLight' | 'errorLight'
 export const StatusPropType = PropTypes.oneOf<Status>(statuses)
 
-interface StatusProps {
-  theme: CactusTheme
+export interface StatusProps {
   status?: Status | null
+  /** !important */
```
daniloPenaR, line 12: ?

daniloPenaR (commented):
Just have a comment about a comment that I am not sure I understood.

modules/cactus-web/src/helpers/status.ts
wilysword: No idea, I think it has to do with the doc generator; I just copied it from `TextInput`.

daniloPenaR (approved):

modules/cactus-web/src/TextInput/TextInput.tsx
```diff
@@ -70,12 +61,21 @@ const Input = styled.input<InputProps>`
 
   &::placeholder {
     color: ${color('mediumContrast')};
-    font-style: oblique;
+    font-style: italic;
```
maxrcollins, line 64: huh.... oblique.

maxrcollins (approved):

modules/cactus-web/src/TextInput/TextInput.tsx
wilysword: Yeah, I looked it up and it seems that in most fonts it's basically the same thing, but some fonts have special italic glyphs, and _can_ have special oblique glyphs but it's a lot more common with italic. I figured it'd be better to standardize on the more common one.

PR #810, opened 2022-10-14
CACTUS-1063 :: Fix Layout Height Bug

Dhalton:
Ticket and UAT: https://repayonline.atlassian.net/browse/CACTUS-1063

I have tested this in _so many_ (probably more than necessary) different device/browser combinations that I'm feeling pretty confident at this point.

I tested channels admin, merchant, and customer apps with:
- Desktop Chrome
- Desktop Firefox
- Desktop Safari

I tested channels merchant and customer apps with:
- IE11
- Newest iOS/Chrome
- Various older iOS/Chrome combos
- Newest iOS/Safari
- Varying older iOS/Safari combos (including Safari 14)
- Newest Android/Chrome (Samsung, Google, and OnePlus devices)
- Various older Android/Chrome combos
- Newest Android/Firefox
- _Some_ older Android/Firefox combos (some older versions of firefox didn't allow me to access localhost)
- I even tested on some obscure Android browsers that I've never even heard of. (Anyone ever heard of UC Browser??)

I wasn't able to reproduce the original issue nor did I find any new issues with the layout in my testing with this fix. That being said, I would still appreciate it if you guys could test in as many device/browser combos as you can just to be safe; I'd rather catch any issues with this solution now than later.

@wilysword I'm sure there's a better, cleaner, less hacky way to do this, so if you'd like me to do it differently I can, but I probably won't go through the same rigorous testing that I did with this exact change that I have here.

Lastly I just want to say that this was hell and I know it's ugly to special case it this way, but at least it works as far as I can tell.

### DIFF
--- a/modules/cactus-web/src/Layout/grid.ts
+++ b/modules/cactus-web/src/Layout/grid.ts
@@ -269,10 +269,16 @@ const toComponentLayout = (role: string, position: Position, order: number): Com
 }
 
 const ZERO_POSITION: CSSPosition = { top: 0, left: 0, bottom: 0, right: 0 }
+// Special case fix for a bug that is caused by the address bar disappearing on Chrome for Android.
+const IS_ANDROID_CHROME =
+  navigator.userAgent.match(/chrome|chromium/i) &&
+  navigator.userAgent.match(/android/i) &&
+  'ontouchstart' in document.documentElement
+const POSITION = IS_ANDROID_CHROME ? 'fixed' : 'absolute'
 const BASIC_GRID = `
   display: -ms-grid;
   display: grid;
-  position: absolute;
+  position: ${POSITION};
   overflow: auto;
 `
 
@@ -284,7 +290,7 @@ const generateGridStyles = (components: ComponentLayout[]): StyleList => {
   const gridItems: GridItem[] = []
   const gridRows: GridMap = {}
   const gridCols: GridMap = {}
-  const fixed: CSSPosition & Sizes & { position?: 'fixed' | 'absolute' } = { ...ZERO_POSITION }
+  const fixed: CSSPosition & Sizes = { ...ZERO_POSITION }
   for (const layout of components) {
     if (layout.type === 'fixed') {
       styles.push(css`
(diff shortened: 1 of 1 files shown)


maxrcollins (approved):
I spent the last 90 minutes testing this on tons of different browsers and it looks good on everything I could think of. Well done. 

I also don't think it's that bad? We're already differentiating IE from not IE in a lot of places. I like this better than all of the browser width/height calculations anyway. Fixing inconsistencies like this is literally the point of the userAgent.

Dhalton:
Yeah the special casing isn't the main reason I don't particularly like this; it's how the style is being applied. It just doesn't really fit in with the rest of the code. I would've rather done it in a css function inside `BASIC_GRID` but for some reason that wouldn't work

wilysword:
Currently `BASIC_GRID` is just a string template; did you change it to `css` when testing an embedded function?
```
// Change
const BASIC_GRID = `
// TO
const BASIC_GRID = css`
```
I'm fine with either way, that's just the most likely reason I can think of why putting a function there wouldn't work. Honestly, though, I'm not sure you even need a function; all the conditions you're checking are constants available at the module level, so you could just make a constant like `const display = isAndroidChrome ? 'fixed' : 'absolute'`.

Dhalton:
I hadn't noticed that `BASIC_GRID` wasn't already using `css` so you're probably right about that being the reason a function didn't work there. I refactored it to determine the browser/device at the module level, though. I tested this most recent commit on OnePlus/Chrome, OnePlus/Firefox, Galaxy/Chrome, Galaxy/Firefox, iOS/Safari, and iOS/Chrome, but I didn't test as thoroughly as I did on Friday for the original fix. I don't see any reason why this would end up being any different functionally than what I had originally though.

maxrcollins:
> Yeah the special casing isn't the main reason I don't particularly like this; it's how the style is being applied. It just doesn't really fit in with the rest of the code. I would've rather done it in a css function inside `BASIC_GRID` but for some reason that wouldn't work

Oh I see. Makes sense.

modules/cactus-web/src/Layout/grid.ts
```diff
@@ -269,10 +269,16 @@ const toComponentLayout = (role: string, position: Position, order: number): Com
 }
 
 const ZERO_POSITION: CSSPosition = { top: 0, left: 0, bottom: 0, right: 0 }
+// Special case fix for a bug that is caused by the address bar disappearing on Chrome for Android.
+const IS_ANDROID_CHROME =
+  navigator.userAgent.match(/chrome|chromium/i) &&
+  navigator.userAgent.match(/android/i) &&
+  'ontouchstart' in document.documentElement
```
wilysword, line 276: This kind of thing usually seems to break in the docs site. It pretty much makes you check everything for existence, e.g. `navigator !== undefined && document !== undefined && ...`.

Dhalton: Thanks for the heads up, pushing up a fix.

wilysword (approved):
Well aside from gatsby sucking, I don't see anything wrong with this.

PR #812, opened 2022-10-17
CACTUS-765 Reach-less Modal

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-765

The biggest things Reach did were scroll trapping (although they did it in a really weird way that broke our layout), focus lock, and handling visibility/events. We already had scroll trapping code, a focus lock, and the Dimmer that controls visibility; that just left the events, which are quite simple. The only really new thing here is the portal.

Also, there's one thing they did which I didn't, because it seemed like overkill: they set `aria-hidden` on everything _but_ the open modal.

### DIFF
--- a/modules/cactus-web/src/IconButton/IconButton.tsx
+++ b/modules/cactus-web/src/IconButton/IconButton.tsx
@@ -3,7 +3,7 @@ import { CactusTheme } from '@repay/cactus-theme'
 import PropTypes from 'prop-types'
 import React from 'react'
 import styled, { css, FlattenInterpolation, ThemeProps } from 'styled-components'
-import { margin, MarginProps } from 'styled-system'
+import { margin, MarginProps, system } from 'styled-system'
 
 import { isIE } from '../helpers/constants'
 import { getOmittableProps } from '../helpers/omit'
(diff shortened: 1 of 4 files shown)


Dhalton (approved):
Surprisingly much simpler than I would've imagined. I published this with `yalc` and tested in the report delivery app to check that this also gets rid of the issue with the body & html height being set to 0 and it works great.

modules/cactus-web/src/IconButton/IconButton.tsx
```diff
@@ -160,6 +160,22 @@ const IconButtonBase = React.forwardRef<HTMLButtonElement, ButtonProps>(
   }
 )
 
+const focusOutline = system({
```
maxrcollins, line 163: Is this just a rewrite or did you need to change a behavior in these styles?

wilysword: The behavior is the same, but this makes it work with responsive values (`iconSize={['small', 'medium']}`). The old code directly indexed the mapping with the prop value, so if the prop was an array you'd end up with a style like `height: calc(100% + undefinedpx)`, which obviously doesn't work.

maxrcollins: Got it. Thanks!

maxrcollins (approved):
Nicely done 👍

modules/cactus-web/src/helpers/portal.ts
```diff
@@ -0,0 +1,22 @@
+import { ReactNode, ReactPortal, useRef } from 'react'
+import { createPortal } from 'react-dom'
+
+const ID = 'cactus-portal-root'
```
maxrcollins, line 4: Is it a bad idea to use a UUID or something on the off chance that we eventually want 2 portals on a page?

maxrcollins: Oh wait, this gets thrown into the ether as soon as the modal is closed doesn't it? Nevermind then

wilysword: That's not how `createPortal` works. There's only one root, and any number of portals can be added to it. Try adding this to a story file, for example:
```
import { usePortal } from '../helpers/portal'
const MyPortal = (props) => usePortal(<div {...props} />)
export const Portals = () => (
  <MyPortal id="one">
    <MyPortal id="two" />
  </MyPortal>
)

```

wilysword: Also, although Reach's portal root is removed when not in use, I saw no harm in keeping this one around permanently.

maxrcollins: I see. Thanks

PR #814, opened 2022-10-24
CACTUS-1060 :: Remove `vh` styles from Dimmer

daniloPenaR:
[CACTUS-1060](https://repayonline.atlassian.net/browse/CACTUS-1060)

I guess the title of this should be different. 

I spent more time struggling with how to build a local version of the `react-ui-base-image` to test this in channels. But I got it and it worked. 

Feel free to test by yourself. I did it with iOS and android.

### DIFF
--- a/modules/cactus-web/src/Dimmer/Dimmer.tsx
+++ b/modules/cactus-web/src/Dimmer/Dimmer.tsx
@@ -27,6 +27,7 @@ export const Dimmer = styledUnpoly(DimmerBase)`
   left: 0;
   width: 100vw;
   height: 100vh;
+  padding-bottom: calc(100vh-100%);
   box-sizing: border-box;
   flex-direction: column;
   justify-content: center;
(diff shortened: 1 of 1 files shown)


Dhalton:
This doesn't fix the original issue as you can see in this screenshot:

Screen Shot 2022-10-25 at 10 05 15 AM

To really test this, you have to undo my short-term fix in Channels, which is [here](https://github.com/repaygithub/repay-api/blob/master/ebpp_ui/app/components/SchedulerModalV2/styleOverrides.js#L25-L27)

modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -27,6 +27,7 @@ export const Dimmer = styledUnpoly(DimmerBase)`
   left: 0;
   width: 100vw;
   height: 100vh;
+  padding-bottom: calc(100vh-100%);
```
maxrcollins, line 30: oof. Make sure you test this in a bunch of browsers. Calc with values like this could be huge pain.

wilysword:
Since it apparently didn't work, I guess just go back to the original ticket description of changing `height: 100vh` to `height: 100%`, see if that works.

Dhalton (approved):
Tested on several device/browser combos including iOS Safari/Chrome, Android Chrome/Firefox, and IE11. LGTM

maxrcollins (approved):
Tested Chrome, Phone Chrome, IOS, Firefox. Looks good to me.

PR #816, opened 2022-10-26
cactus-1075:: modal-bugfix-1

maxrcollins:
Ticket & UAT: [CACTUS-1075](https://repayonline.atlassian.net/browse/CACTUS-1075)

Tested this on Channels and it works.

### DIFF
--- a/modules/cactus-web/src/Modal/Modal.tsx
+++ b/modules/cactus-web/src/Modal/Modal.tsx
@@ -23,7 +23,6 @@ import {
 
 import Dimmer from '../Dimmer/Dimmer'
 import FocusLock from '../FocusLock/FocusLock'
-import { isFocusLost } from '../helpers/events'
 import { omitProps } from '../helpers/omit'
 import { usePortal } from '../helpers/portal'
 import { trapScroll } from '../helpers/scroll'
@@ -79,11 +78,27 @@ const ModalBase = React.forwardRef<HTMLDivElement, ModalProps>(
         return trapScroll(dimmerRef)
       }
     }, [isOpen]) // eslint-disable-line react-hooks/exhaustive-deps
-    const closeOnBlur = (event: React.FocusEvent<HTMLDivElement>) => {
-      if (isFocusLost(event)) {
+
+    const mouseRef = React.useRef<EventTarget | null>(null)
+
+    const onMouseDown = (event: React.MouseEvent<HTMLElement>) => {
+      mouseRef.current = event.target
+    }
+
+    const onDimmerClick = (event: React.MouseEvent<HTMLElement>) => {
+      // First condition ignores click-and-drag.
+      if (event.target === mouseRef.current) {
         onClose()
       }
     }
+
+    const { onClick: propClick } = props
+
+    const onModalClick = (event: React.MouseEvent<HTMLDivElement>) => {
+      event.stopPropagation()
+      propClick?.(event)
+    }
+
     const closeOnEscape = (event: React.KeyboardEvent) => {
       if (event.key === 'Escape') {
         onClose()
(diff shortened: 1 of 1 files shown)


modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -94,7 +109,8 @@ const ModalBase = React.forwardRef<HTMLDivElement, ModalProps>(
         className={className}
         ref={dimmerRef}
         active={isOpen}
-        onBlur={closeOnBlur}
+        onMouseDown={onMouseDown}
+        onClick={onDimmerClick}
         onKeyDown={closeOnEscape}
       >
         <FocusLock className="flex-container">
```
maxrcollins, line 116: Let me know if I should get rid of the FocusLock. It might not be required anymore here since we're not relying on focus but I figured it could be useful for some passed in 'onClick' functions?

Dhalton: `FocusLock` is there to prevent focus from leaving the modal if you tab through the elements with your keyboard so we still need it

modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -79,11 +78,27 @@ const ModalBase = React.forwardRef<HTMLDivElement, ModalProps>(
...
+
+    const onMouseDown = (event: React.MouseEvent<HTMLElement>) => {
+      mouseRef.current = event.target
+    }
+
+    const onDimmerClick = (event: React.MouseEvent<HTMLElement>) => {
+      // First condition ignores click-and-drag.
+      if (event.target === mouseRef.current) {
         onClose()
       }
```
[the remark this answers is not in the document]
wilysword: No, because the target is actually the FocusLock/flex container. I had the same thought when I originally wrote it, and the fact it didn't work is what led to me being a lazy bum and using onBlur instead. If you really don't like it you could probably get away with something like `dimmerRef.current.firstElementChild` or something, but to me that seems even worse.

modules/cactus-web/src/Modal/Modal.tsx
```diff
@@ -79,11 +78,27 @@ const ModalBase = React.forwardRef<HTMLDivElement, ModalProps>(
...
     }
+
+    const { onClick: propClick } = props
+
+    const onModalClick = (event: React.MouseEvent<HTMLDivElement>) => {
+      event.stopPropagation()
+      propClick?.(event)
+    }
```
wilysword, line 100: I wrote it this way in the suggestion because I was only showing the relevant bits, but since there's already a destructuring of props at the top of the component, I'd do one of two things here:
1. Destructure `onClick` along with everything else, keep the handler here the same.
2. Destructure `onClick` just the way you have it currently, but change the assignment of the handler so you don't have to explicitly pass it to the component below:
```
const { onClick } = props
props.onClick = (event) => {
  event.stopPropagation()
  onClick?.(event)
}
```
I have a slight preference for (2) because I'm pretty sure adding anything to the top line will cause the linter to wrap it, but I'd be fine with either.

modules/cactus-web/src/Modal/Modal.tsx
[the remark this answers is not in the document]
wilysword: Maybe change the comment as well, since there are no longer two conditions.

Dhalton (approved):
A little weird that there's a test failure in cactus but it's such a small, easy fix I'm not _too_ worried about it. Works for me on firefox too.

modules/cactus-web/src/Modal/Modal.tsx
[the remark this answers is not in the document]
maxrcollins: I'm with you. I `dimmerRef.current.firstElementChild` feels weird. I can change the comment though

modules/cactus-web/src/Modal/Modal.tsx
maxrcollins: I think I'm going to go with 1. I'd rather it not wrap, but I also feel weird about reassigning props like that, even if it's completely fine in this case.

maxrcollins: Oof. I see why you didn't want to break the props....

maxrcollins:
Didn't really change anything major but requesting re-review because I broke out the props differently

wilysword (approved):

PR #817, opened 2022-10-27
CACTUS-970 :: ScreenSizeProvider using deprecated function

daniloPenaR:
[CACTUS-970](https://repayonline.atlassian.net/browse/CACTUS-970)

I tested this using `mock-ebpp`, just added a local version of `cactus-web` (previously built) using yalc, ran `yarn && yarn install`, and ensure the app isn't broken in our loved IE11


Dhalton (approved):
LGTM. I threw some console logs in there to see which branches were being taken and then tested in Firefox and IE11. Firefox used the `addEventListener`/`removeEventListener` while IE11 used `addListener`/`removeListener`. Both triggered accurate updates when I added a `useScreenSize` call to `mock-ebpp`, too.

maxrcollins (approved):

PR #819, opened 2022-10-31
CACTUS-1006 :: processMeta return value

Dhalton:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-1006

There's not really an easy way to UAT this, but there's a unit test that should cover it. If you really want to test it, you can add a `processMeta` function onto any of the fields in the rules form in `mock-ebpp`, and change some value in `props` without returning anything from `processMeta`.

Something like this would be sufficient for testing:
```
const processMeta = (props) => {
  props.error = 'Testing, testing, 123'
}
```


daniloPenaR (approved):
works fine for me.

maxrcollins (approved):

PR #820, opened 2022-11-01
CACTUS-1008: add absolute mode to Dimmer

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1008

Debatable: I took out the bit that blurs on activation. Using the modal as an example, it's redundant: the dimmer blurs, and then immediately afterwards the modal focuses, so there was really no point in the explicit blur. It's pretty much like that in every use case I can think of: if you're adding an overlay, then you should probably be ensuring whatever component you're making is keyboard-accessible. _Maybe_ a case could be made if it were something like a simple spinner, but I think avoiding things like that is kind of the point of using a front-end framework. If you guys think it should be included I'll put it back in.


modules/cactus-web/src/Dimmer/Dimmer.tsx
```diff
@@ -1,37 +1,50 @@
...
-        ;(document.activeElement as HTMLElement).blur()
-      } catch {}
-    }
-  }, [active])
   return active ? <div {...props} ref={ref} /> : null
 })
 DimmerBase.displayName = 'Dimmer'
 
-export const Dimmer = styledUnpoly(DimmerBase)`
+const background = (opacity?: Opacity) => `rgba(46, 53, 56, ${opacity || '0.9'})`
```
maxrcollins, line 22: In the ticket you mentioned you wanted to find out why it was 46, 53, 56 and we couldn't figure a reason out for it. Did you talk to anyone? I'm guessing this is a specific repay color or something but maybe we should at least allow for someone to pass in a prop to override it?

wilysword: I did not talk to anyone, but I doubt there's a deep reason for it; my guess is that it was just copied directly from the prototype. It's close enough to gray/black that unless you want opacity=1 you'd never even notice (or maybe I'm just a bit color blind).

wilysword: Although, since you got me curious I checked into it, and that's the default theme's `darkestContrast` color. Still, I don't think there's much reason to tie it to the theme; at 10% saturation there's barely any color variation, certainly not anything a normal person could notice.

maxrcollins: I agree. I can't really tell the different either tbh.

maxrcollins (approved):
Works as advertised.

modules/cactus-web/src/Dimmer/Dimmer.tsx
daniloPenaR: BTW, Yup, those values were taken by me from the prototypes 🤷🏽‍♂️, this component is one of my first tickets here 😝 
Screen Shot 2022-11-02 at 4 08 30 PM

daniloPenaR (approved):

PR #821, opened 2022-11-04
Cactus-1072  fix linter errors

maxrcollins:
Ticket & UAT: [CACTUS-1072](https://repayonline.atlassian.net/browse/CACTUS-1072)
The first couple of commits are to fix specific common errors. The last one has a couple of rule changes and a couple of misc fixes. One of the rule updates was something that wasn't supported by IE11 (css appearance) but I didn't think that it was important we change it because it didn't work before and it's no a breaking bug, just a small appearance difference. 

There are 4 things left that I didn't touch. Two of them (unknown word) are just a sideEffect of the linter not working perfectly with StyledComponents but I didn't want to disable the rule because it was helpful elsewhere, and 2 of them are small appearance differences in IE11 because they're not supported.

I'm not 100% I like all of the changes. A lot of them I think were really helpful, but others didn't seem like they were as helpful. lmk what you think


modules/cactus-web/src/Layout/Sidebar.tsx
```diff
@@ -106,7 +106,7 @@ const SidebarDiv = styled.div`
     ${Sidebar.Button}[aria-expanded='true']::after {
       content: '';
       z-index: 99;
-      background-color: rgba(0, 0, 0, 0.5);
+      background: rgba(0, 0, 0, 0.5);
```
wilysword, line 109: What's the reason for this change? I don't mind, I'm just curious.

wilysword (approved):
You've got some visual changes in the storyshots that need to be merged in. I'm curious, which of these changes did you not like? They all seem quite reasonable to me, except perhaps the background one.

modules/cactus-web/src/Layout/Sidebar.tsx
maxrcollins: This is one of the changes I don't like heh. I don't necessarily have a problem with it, but it doesn't really make a seem neater to me when it's just one thing.

The linter wants to combine things that can be combined which makes sense for like `padding-left` + `padding-right` + `padding-top` + `padding-bottom`, but I didn't necessarily like that it made me do the `background-color into `background` here. Same with   `flex-direction` + `flex-wrap` into `flex-flow` because I always have to remind myself what exactly `flex-flow` means, unlike the other two but that might just be me.

modules/cactus-web/src/StyleProvider/StyleProvider.story.tsx
```diff
@@ -10,7 +10,8 @@ const BreakpointBox = styled(Box)`
   max-width: 320px;
   width: 50px;
   height: 50px;
-  border-radius: 15% ${(p): string => p.theme.mediaQueries.small} {
+  border-radius: 15%;
+  ${(p): string => p.theme.mediaQueries.small} {
```
maxrcollins, line 14: The lack of semicolon here was causing these to compile incorrectly and neither the border radius or the media query (or anything in it) was valid css.

daniloPenaR (approved):
Most of them are looking good to me, I don't have any strong opinion or dislike about those changes. Just curious, what are the changes that you do not like?

maxrcollins:
> Most of them are looking good to me, I don't have any strong opinion or dislike about those changes. Just curious, what are the changes that you do not like?

I mentioned some of them above.  Like `background-color: rgba(0, 0, 0, 0.5);` into `background: rgba(0, 0, 0, 0.5);` or `flex-direction` + `flex-wrap` into `flex-flow`. They just seem to not add or clarify anything to me.

PR #823, opened 2022-11-04
Danilo 1067 CI TESTS

daniloPenaR:
🧪


.github/workflows/continuous-integration.yml
```diff
@@ -110,86 +45,23 @@ jobs:
...
-
-      - name: Extract branch name
-        shell: bash
-        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
-        id: extract_branch
-
-      - name: Create Pull Request
-        uses: ./actions/create-pull-request
+      - uses: actions/cache@v2
+      - name: Cache node_modules
```
wilysword, line 56: I believe this line is causing an error: https://github.com/repaygithub/cactus/actions/runs/3397195249

wilysword (commented):
Out of curiosity, why did you add a second cache step? I don't think the install command takes very long if it works from the first cache step; also making sure we run running `yarn install --immutable` is probably something we _should_ do in every build, to make sure we didn't commit some `package.json` changes without also updating the lockfile.

wilysword:
In the future please use draft reviews for stuff like this, so we know not to review it.

daniloPenaR:
> In the future please use draft reviews for stuff like this, so we know not to review it.

Is a draft PR able to execute Ci actions?

wilysword:
I'm about 90% sure they do, but that's a good point. At the very least you can start it as a draft review, and if it doesn't work change it then.

daniloPenaR:
Gotcha!

PR #827, opened 2022-11-10
CACTUS-1067 :: Improve CI Pipeline Efficiency

daniloPenaR:
[CACTUS-1067](https://repayonline.atlassian.net/browse/CACTUS-1067)

I’ve spent too many days working on this. And I have to say that `cache` and `artifacts` tools are pretty awesome, especially the second one. My main target was to try to reduce the time achieved by Dhalton parallelizing the jobs. I started by trying to cache both lockfiles and the node modules folder and taking notes of the different cases and results, in that process I realized that only caching one or other was enough, and I choose to go with `node_modules`. With the build step I tried many combinations, and then I created a workflow with one job which install all the dependencies (or gets them from the cache), builds the modules, and then uploads them as artifacts to be shared and downloaded by the subsequent jobs (which starts only when that first job is finished). In that highway I’ve found an interesting approach, the [composite actions](https://wallis.dev/blog/composite-github-actions), a very helpful way of isolating the actions that will be repeated by every job, so this helps reduce duplication. 

But at that point I have to admit that I was very frustrated with the result: It seems that using artifacts to store the `dist` folders, combined with the build process and upload process doesn’t reduce the total duration of the CI process, in many cases this first blocking job takes at least [4 minutes before finishing and triggering the subsequent parallelized jobs](https://github.com/repaygithub/cactus/actions/runs/3424356203). But this approach was the most consistent IMO. I thought of other ways that maybe could lead to interesting results such as Building the modules, cache that builds, uploading that build using artifacts and set up a maximum duration of 1 day… But at the end, this maybe adds a stupid level of complexity and the documentation of ghActions takes some paragraphs to try to explain why those processes are not switchable. 

I got rid of artifacts and I was going to stop there and just implement the `cache` for the `node_modules` folder and use composite for the duplicate actions, but then I decided to try another option: Build and cache our modules.
I did not consider this approach in the beginning, because some syntax used by YAML and some expressions owned by a terminal sometimes are pretty confusing, but I am starting to understand this better so I give that approach a shot. The results are way better, caching the build outputs reduce the build time from 2 minutes to ~30 seconds.

What I am concerning now is how GH actions could detect that there were changes that maybe could affect the build output and make differences between the cached one. I am using `hashFiles` to pass the directory of the `src` folders, which could be sensitive in most cases, and passing that string as the cache key. The cache key is used to search for a cache and is set every time a new cache is saved. 

The average duration continues to oscillate between ~15-20 minutes, this is in part because of the jobs that are currently performing integration tests and visual tests, using composite actions is the best way to split and reuse some duplicate jobs, so the main YAML workflow is more readable. 

Caching the build outputs reduces the time of every job, now executed at the same time. But I am not 100% sure that the cache will be correctly replaced every time the output changes. 

The artifacts approach is more consistent IMO. But it requires a first job to build everything, and upload the outputs and this could increment the duration of the CI process by almost 3 minutes, so the average duration times could be ~20-22 minutes.


.github/actions/cactus-build/action.yml
```diff
@@ -0,0 +1,90 @@
...
+    - name: Build cactus-i18n
+      run: yarn i18n build
+      if: steps.cache-i18n-dist.outputs.cache-hit != 'true'
+      shell: bash
+
+    - name:  Cache icons build 
+      id: cache-icons-dist
+      uses: actions/cache@v3
+      env:
+        icons-key: ${{ hashFiles('modules/cactus-icons/src/**') }}
```
wilysword, line 52: In icons it's a lot more complicated than this. If any of `scripts`, `svgr.config.js`, `src`, or `svgs` change it could change the build.

.github/workflows/continuous-integration.yml
```diff
@@ -7,48 +7,17 @@ on:
...
-          restore-keys: |
-            ${{ runner.os }}-yarn-
-      - name: NodeJS Version
-        run: node --version
-
-      - name: Install Packages
-        run: yarn install --immutable
-
-      - name: Lint Packages
-        run: yarn lint
```
wilysword, line 33: Did you put this step back in anywhere? I don't see it, but I might just be missing it.

wilysword (commented):
This is great work, I especially like the composite actions stuff.

We should definitely get more opinions here, but personally I don't think we should cache the build. Unlike install which is tracked by a single `yarn.lock` file, there's just too much that could change to affect the build: source files, package.json, tsconfig; and cactus-icons is especially complicated. There's also the fact that since the jobs run in parallel, when you do get a cache miss it'll rebuild the cache 4 times. I think I'd rather have the slightly longer run times, but like I said, let's get some more opinions.

maxrcollins:
This is really cool.  Thanks for all the research. @wilysword 's comment: I may ask one of you to ELI5 why cactus-icons is so much more complicated. (also does it need to be? Could we simplify it?), but regardless, if there's a risk of the cache missing a change, I'd rather not cache and have the extra few minutes on the build for peace of mind. We only want to make changes that we know don't cut corners

Also, do we know why one of the Integration tests is failing? A bunch of them pass no problem so it doesn't seem like it's part of your process

wilysword:
> I may ask one of you to ELI5 why cactus-icons is so much more complicated. (also does it need to be? Could we simplify it?)

I don't know why the directory structure is how it is, but in general it's different because it's a combination of a set of SVGs, templates, and a few supplementary source files. The build process happens in steps: first the SVGs are loaded into the template and the individual icon components are written. Then the source files are copied into the folder where the icons were written, and an index.js (or .ts) is generated. Finally, the normal build is run against all the generated code.

daniloPenaR:
I agree with Glen, caching the built outputs could not be the best idea, we can leave that as it was before, all the jobs perform their own build step and use composite actions to create unique `build` actions (or execute `yarn build`).

Or I can go back and use artifacts [this way ](https://github.com/repaygithub/cactus/actions/runs/3424356203)

wilysword:
I don't have a strong preference either way. Maybe leaning slightly to the artifacts: it'll take about the same amount of time either way, but with the artifacts I think less work is being done overall.

wilysword (approved):

maxrcollins (approved):
Looks good to me. Thanks Danilo!

PR #828, opened 2022-11-10
CACTUS-1069 Inline Styles

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1069

This took way longer than I thought it would, so we'll definitely need at least one more ticket to finish. Part of that was because I can't leave well enough alone, so I was also converting some of the static styles to use theme helpers. Part of it was because I started out swapping certain components (e.g. Pagination) from "inner styled components" to being styled components themselves (slightly more efficient, easier to make style overrides), before realizing that I was just muddying the waters and making things harder for reviewers. (I didn't go back and undo the ones I already did, though.) I kind of have a secondary goal with these changes to get rid of `styledWithClass` & `styledUnpoly` (since `withStyles` can do the same things), and also to get rid of `helpers/theme`.

The third breaking change (listed in the ticket) does give me some pause, though. I was not expecting there to be such a limitation with the `style` prop, and it's giving me some second thoughts. If any of you think it's a deal breaker, we should discuss it before making the second ticket.

Finally, I recommend going through one commit at a time; several of the commits have extra notes in the commit messages.

### DIFF
--- a/modules/cactus-web/src/helpers/styled.ts
+++ b/modules/cactus-web/src/helpers/styled.ts
@@ -1,11 +1,12 @@
-import { CactusTheme } from '@repay/cactus-theme'
+import { CactusTheme, mediaGTE } from '@repay/cactus-theme'
 import { Property } from 'csstype'
 import { omit, pick } from 'lodash'
 import PropTypes from 'prop-types'
 import React from 'react'
-import styled, { StyledComponent, ThemedStyledFunction } from 'styled-components'
+import styled, { StyledComponent, StyledConfig, ThemedStyledFunction } from 'styled-components'
 import * as SS from 'styled-system'
 
+import { SIZES, useScreenSize } from '../ScreenSizeProvider/ScreenSizeProvider'
 import { isIE } from './constants'
 
 // This file exists, in part, because styled-components types are a PAIN.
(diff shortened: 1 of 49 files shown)


modules/cactus-web/src/helpers/styled.ts
```diff
@@ -35,6 +36,112 @@ export const styledUnpoly = <C extends React.ElementType, F extends React.Elemen
...
+
+// To keep type inference simple we can't include the style prop types,
+// so you need to include them in the tagged template:
+//   const Component = withStyles('div', margin)<MarginProps>`
+//     static: style;
+//   `
+export const withStyles: WithStyles = (component: React.ComponentType, options: StyleOpts) => {
+  const {
+    transitiveProps,
+    preserveProps,
```
Dhalton, line 113: Can you explain what `preserveProps` is for? It's the only param on this function that I don't think I'm understanding. It looks like any props listed in there get removed from the set of transitive props, so does that mean any `preserveProps` get passed on down to the DOM elements?

I feel like that can't be right because the props I see that you're passing to this param are things like `flexFlow`, `width`, and `maxWidth`, and I don't think those are supposed to be passed to DOM elements.

Any insight is appreciated, thanks.

Dhalton (commented):
Lookin' good to me. I have one question about `withStyles` that I'm not really understanding, but that's about it.

As far as the third breaking change you mention goes, I feel like that's a weird interaction but I don't think it's enough to completely turn me off of this idea...I mean there are certainly ways around that behavior and like you say I don't think it's very common to conditionally set a prop altogether, anyway. I think we may want to document that behavior, but it's not a deal-breaker for me.

modules/cactus-web/src/helpers/styled.ts
wilysword: Rather than saying they're passed to the DOM, it's more accurate to say they're passed to whatever the inner component is. In other words you shouldn't use it unless you're styling a complex component. The cases where I used it are cases where those props were needed not only for styling, but also to make logical decisions about the components' behavior: thus, inside the component they're destructured and not passed on to the DOM.
```
const Complex = ({ someStyle, ...domProps }) => {
  if (someStyle) {
    doSomethingThatOnlyHappensWhenStyleIsPresent()
  }
  return <div {...domProps} />
}
const Component = withStyles('div', {
  as: Complex,
  styles=[someStyle],
  preserveProps: ['someStyle'],
})``
```

Dhalton: Oh okay, that makes more sense. It was a misunderstanding of transient props on my part. Thanks

Dhalton (approved):

daniloPenaR (approved):
LGTM.

PR #834, opened 2022-11-18
CACTUS-767 New SplitButton interface

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-767

Made some slight changes to the styles: specifically I made the padding & height match the normal `Button` component. Added styles for individual actions to be disabled. I also spent more time than I probably should have trying to simplify the color styles. (Sometimes I get stuck in a loop where I think something should be easy, but can't actually figure out how to do it...) And since I was already doing so much, I went ahead and implemented the inline styles as well.

Question: when there's only one action (the main action; you can test this by deleting the "actions" control in the "With Collisions" story), should we render the SplitButton as a normal Button, or keep the drop-down but have it disabled? I went with the latter, but if people think it'd look better as the former I'd be fine with that.


daniloPenaR (approved):

maxrcollins:
I'm on the fence but I think I agree with you. With only the main action, the split button should still have the dropdown but it should be disabled. If only because If they want a regular button, they should use a regular button.

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -1,216 +1,190 @@
...
-  variant?: SplitButtonVariant
 }
 
-interface SplitButtonActionProps extends Omit<MenuItemProps, 'onSelect'> {
-  // !important
-  onSelect: () => any
-  icon?: React.FC<IconProps>
+interface GetColor {
+  (variant: SplitButtonVariant | undefined, c: CactusTheme['colorStyles']): ColorStyle
+  (variant: SplitButtonVariant | undefined, c: CactusTheme['colors']): string
```
maxrcollins, line 53: Every once in a while I think i'm getting the hang of TypeScript and then I see a few lines like this that make the gears grind in my head

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -1,216 +1,190 @@
...
-  background-color: ${(p): string => p.theme.colors.darkestContrast};
-  height: 32px;
-  width: 36px;
-  ${(p) => dropdownButtonShapeMap[p.theme.shape]}
+  min-width: 36px;
+  padding: 0;
+  border-radius: 0 ${radius(20)} ${radius(20)} 0;
   margin-left: 1px;
-  border: 0px;
+  border: 0;
```
maxrcollins, line 156: <3

maxrcollins: The linter can't catch all of these. It has trouble with some blocks

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -234,105 +208,110 @@ const handleArrows = (event: React.KeyboardEvent<HTMLElement>, toggle: TogglePop
...
+        const nestedMain = getMainAction(child.props.children, actions, child.key)
+        if (!main && nestedMain) main = nestedMain
+      }
+    } else if (child.props.main) {
+      if (!main) main = child
+    } else if (key) {
+      actions.push(React.cloneElement(child, { key: `${key}${child.key}` }))
+    } else {
+      actions.push(child)
+    }
```
maxrcollins, line 230: @_@

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -234,105 +208,110 @@ const handleArrows = (event: React.KeyboardEvent<HTMLElement>, toggle: TogglePop
...
-    ${NavigationChevronDown} {
-      transform: rotate3d(1, 0, 0, 180deg);
-    }
-    ${getVariantDark};
+  svg {
+    margin-right: 4px;
+    font-size: ${iconSize('small')};
   }
-`
+` as any
```
maxrcollins, line 308: how is this not TS erroring? 🤔  Not that I have anything against it. Just for me for the future

modules/cactus-web/src/SplitButton/SplitButton.tsx
maxrcollins: Also 219-222

maxrcollins:
Thanks for doing all of the extra cleanup. My only comment is that layered if, else if statement is kind of hard to read. I'm not sure how to make it simple but it could probably be cleaned up a little.

I tested on a few browsers and it works as advertised so 👍

modules/cactus-web/src/SplitButton/SplitButton.tsx
wilysword: This is one of the ways you do function overloads. The other way requires use of the `function` keyword, so you can't use it with arrow functions.

modules/cactus-web/src/SplitButton/SplitButton.tsx
wilysword: What error specifically are you expecting? But to paraphrase: "`as any` covers a multitude of sins."

modules/cactus-web/src/SplitButton/SplitButton.tsx
wilysword: You're right, I was overthinking things.

modules/cactus-web/src/SplitButton/SplitButton.tsx
maxrcollins: whenever I use 'any' I get a warning that I shouldn't use it.

wilysword: What gives you the warning? `yarn test:types` and `yarn lint` both come back clear for me.

modules/cactus-web/src/SplitButton/SplitButton.tsx
maxrcollins: I see, thanks!

modules/cactus-web/src/SplitButton/SplitButton.tsx
```diff
@@ -1,216 +1,190 @@
...
-    ${(p): ColorStyle => p.theme.colorStyles.successDark};
-  `,
-})
-
-const MainActionButton = styled.button<VariantInterface>`
+const disabledStyles = css`
+  &:disabled {
+    ${colorStyle('disable')}
+    border-color: transparent;
+    cursor: not-allowed;
```
maxrcollins, line 72: Part of why I wish we had global styles is that we could just have something like `*:disabled { cursor: not-allowed } `so we wouldn't have to do this individually for every component that is disabled. 
Not telling you to change anything, this just seems like a block we shouldn't have to write.

maxrcollins (approved):

PR #835, opened 2022-11-18
CACTUS-1081: accessible field autotooltip set to false by default

maxrcollins:
Ticket: https://repayonline.atlassian.net/browse/CACTUS-1081

I went to change the defaults in the story but thought it was kind of annoying so I left it. Maybe I should add a test or something but other than that not sure what else to change.


Dhalton (approved):

wilysword (approved):

maxrcollins:
@wilysword  Same thing. I'll make sure one of the commits has BREAKING CHANGE in it next time

PR #836, opened 2022-11-21
CACTUS-1087 :: Storybook Docs Addon

Dhalton:
https://repayonline.atlassian.net/browse/CACTUS-1087

Well this was a lot more difficult than I thought it would be.

I'm targeting this at Glen's branch for now to make it easier to see my changes but we should switch the target after that's merged.

We didn't have any custom webpack config for rendering the docs that aren't associated with a component/story before, but because I removed `@storybook/addon-docs` from the `addons` section of the storybook config, I had to put an mdx loader in the base storybook webpack config too, so there's some duplicated config here.

This is not meeting all of the acceptance criteria, but I've been working on this for a week and I finally have _something_ that at least renders the MDX docs, so I'll make a follow-up ticket for this one. We'll still need to figure out ways to:
- Inject info about the stories to the MDX
- Make some components that ingest that info and do something with it when rendered in the MDX

I'm obviously still open to any ideas/suggestions you guys have for this...part of the reason I'm even making this PR in its current state is so I can get feedback and help in the first place, so I'm interested to hear what you guys think.


modules/cactus-web/src/CheckBox/CheckBox.mdx
```diff
@@ -42,4 +41,4 @@ import { CheckBox } from '@repay/cactus-web'
 
 ## Properties
 
-<ArgsTable story="Basic Usage" />
+TODO
```
maxrcollins, line 44: What needs to be done for this TODO?

storybook/.storybook/webpack.config.js
```diff
@@ -0,0 +1,88 @@
...
+    ...babelOptions,
+    ...mdxBabelOptions,
+    plugins,
+  }
+}
+
+module.exports = ({ config }) => {
+  config.entry = [
+    require.resolve('../../modules/cactus-web/dist/helpers/polyfills'),
+    ...config.entry,
```
maxrcollins, line 33: I'm not really sure what some of the stuff in this file does. Do we want to be pointing this towards a relative path in the dist folder?

modules/cactus-web/src/CheckBox/CheckBox.mdx
Dhalton: I can write the follow-up ticket, but I'd like to wait and see exactly what spot this is left in before I do.

storybook/.storybook/webpack.config.js
Dhalton: Tbh I'm not sure if this is necessary, it's taken straight from the storybook config for cactus. I can try to get rid of it though

storybook/addon-docs/src/preset.js
```diff
@@ -0,0 +1,96 @@
+function managerEntries(entry = []) {
+  return [...entry, require.resolve('./preview')] //👈 Addon implementation
+}
+
+const remarkExternalLinks = import('remark-external-links')
+const remarkSlug = import('remark-slug')
+
+function createBabelOptions({ babelOptions, mdxBabelOptions, configureJSX }) {
```
maxrcollins, line 8: So if cleanup is going to be done in the followup ticket then you can ignore this, but it looks like this function only gets used with configureJSX. Are we going to use babelOptions or mdxBabelOptions? If not then I think this can be simplified

storybook/addon-docs/src/preset.js
```diff
@@ -0,0 +1,96 @@
...
+          use: [
+            {
+              loader: require.resolve('babel-loader'),
+              options: createBabelOptions({ configureJSX: true }),
+            },
+            {
+              loader: mdxLoader,
+              options: {
+                ...mdxLoaderOptions,
+                skipCsf: false,
```
maxrcollins, line 71: What's this do? would it be helpful to have the other mdx files run through this?

maxrcollins: I tried googling briefly but didn't get anything helpful

storybook/addon-docs/src/preset.js
Dhalton: Yeah this is copied from storybook and I was really frustrated yesterday so I just pushed it up how it is without doing much cleanup. I'll get it in a better spot though.

storybook/addon-docs/src/preset.js
Dhalton: I don't really know. This is copied from storybook's config. I figured they probably know how to use their tools better than I do so I didn't want to change the config too much.

storybook/addon-docs/src/preview.js
```diff
@@ -0,0 +1,44 @@
...
+    title: 'Docs',
+    //👇 Checks the current route for the story
+    route: ({ storyId, refId }) => (refId ? `/docstab/${refId}_${storyId}` : `/docstab/${storyId}`),
+    //👇 Shows the Tab UI element in docstab view mode
+    match: ({ viewMode }) => {
+      return viewMode === 'docstab'
+    },
+    render: ({ active }) => <Docs api={api} active={active} />,
+  })
+})
```
wilysword, line 44: I have no idea how or even if this is working. I tried modifying this file while testing stuff (adding console.log, changing the padding on the Box, changing the title), and none of the changes were ever reflected. Do you know what might be causing that?

storybook/addon-docs/src/preview.js
```diff
@@ -0,0 +1,44 @@
...
+  const type = storyData?.type
+  const docsPath = storyData?.parameters?.docsPath
+
+  useEffect(() => {
+    // Docs that aren't associated with a component/story appear to have the docs type somehow
+    if (type === 'docs') setDocsContent(undefined)
+  }, [type])
+
+  if (!!docsPath) {
+    import(`../../../modules/cactus-web/src/${docsPath}`).then((res) => {
```
wilysword, line 18: We need this to work for other modules as well.

modules/cactus-web/src/AccessibleField/AccessibleField.story.tsx
```diff
@@ -2,14 +2,11 @@ import React from 'react'
 
 import { AccessibleField } from '../'
 import { Story } from '../helpers/storybook'
-import docsMeta from './AccessibleField.story.mdx'
```
wilysword, line 5: I would very strongly prefer if we could either import them here, or somehow make that part of the build process (though the more I look at the Storybook code, I'm pretty sure the latter isn't possible).

wilysword (commented):
To be honest, I'm no longer sure if we can make this work. I had assumed, given the existence of the addon framework, they had used it in writing their own docs addon; but it appears that they actually special cased a bunch of docs stuff in the core code.

storybook/addon-docs/src/preview.js
Dhalton: Maybe try running storybook with the `--no-manager-cache` flag? I bet that's it.

storybook/addon-docs/src/preview.js
Dhalton: 🤦 Right, yeah duh. I'll update it.

modules/cactus-web/src/AccessibleField/AccessibleField.story.tsx
Dhalton: I tried for a long time to get the imports to work in the story file, but it never wanted to render that content for me. I feel like it might not be possible, but I can try again

Dhalton: When I import the mdx in the story file and pass it to the addon to render it there, I just get this error:
`Error: MDXContent(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.`

This was what ultimately led me to import in the addon instead. I've not been able to figure that one out.

wilysword:
So it looks like for some reason they convert the parameters to JSON using a library that somehow stringifies functions and then attempts to restore them. It obviously doesn't work with MDX, and I'm not sure why they use it at all nor why it doesn't work. I was initially assuming we'd be able to render our docs in a preview iframe like they do, but it appears that's actually part of the core logic, not the addon.

With this, I'm not sure if we can actually make our own custom docs addon. What you have here does work pretty well, but the problem is that as far as I know, webpack dynamic imports work by including pretty much everything that _could_ be imported in the bundle, which in our case is the entire `modules` directory. That would make the entire storybook bundle a great deal larger than it needs to be.

maxrcollins:
> That would make the entire storybook bundle a great deal larger than it needs to be.

How much larger? I don't think it's _that_ much of a problem since it wouldn't affect the core library.

Dhalton:
Couldn't we just separate the mdx files from the `modules` directory and put them somewhere else, like the `storybook` directory? Then wouldn't webpack ignore the `modules`?

wilysword:
> How much larger?

I don't know. I did have a couple of other ideas that might work, without needing to completely replace the docs addon. First is https://github.com/storybookjs/storybook/blob/next/code/addons/docs/docs/recipes.md#overwriting-docs-container; I played around for about five minutes when I was working on the first ticket and it didn't work, but since my focus at the time was on other stuff, I didn't try very hard. Second (and this is less likely to work), our existing addon uses functions like `argsEnhancers` to modify the story's args, maybe there's a similar function for parameters that we could use to wrap the docs page?

> Couldn't we just separate the mdx files

I'd definitely leave that as a last resort, if not discard the idea entirely; I think the current structure makes maintaining the docs significantly easier.

Dhalton:
> EDIT: Not sure if we'd want a new ticket to try those ideas out, instead of adding more work onto this one.

I think we should make a separate ticket for it.

wilysword:
That's fair, we could put this ticket on hold in the mean time since we might still need it if other ideas don't work.

Dhalton:
Closing this but I'll leave the branch in case we decide to come back to this idea.

PR #837, opened 2022-11-23
CACTUS-1092 :: SubmitButton

Dhalton:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-1092

### DIFF
--- a/modules/cactus-form/src/FormSpy.tsx
+++ b/modules/cactus-form/src/FormSpy.tsx
@@ -1,4 +1,4 @@
-import { FormState, formSubscriptionItems } from 'final-form'
+import { FormApi, FormState, formSubscriptionItems } from 'final-form'
 import PropTypes from 'prop-types'
 import React from 'react'
 import { useForm, UseFormStateParams } from 'react-final-form'
@@ -18,14 +18,14 @@ export interface FormSpyProps
     UnknownProps {
   processState?: StateProcessor
 }
-type State = FormState<Record<string, any>, Partial<Record<string, any>>>
+type State = FormState<Record<string, any>, Partial<Record<string, any>>> & { form?: FormApi }
 
 const FormSpy: RenderFunc<FormSpyProps> = ({
   render,
   component,
   children,
   subscription,
-  processState,
+  processState = Object.assign,
   ...rest
 }: FormSpyProps) => {
   const form = useForm('FormSpy')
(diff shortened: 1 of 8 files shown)


daniloPenaR (approved):
Tested and seems to work.

modules/cactus-form/src/FormSpy.tsx
```diff
@@ -32,11 +40,18 @@ const FormSpy: RenderFunc<FormSpyProps> = ({
   }, []) // eslint-disable-line react-hooks/exhaustive-deps
 
   const props = Object.assign(rest, state, { form })
+  let updatedProps
+  if (typeof processState === 'function') {
+    updatedProps = processState(props, state)
+  }
+  const finalProps = updatedProps || props
```
wilysword, line 47: I had pictured `processState` as something that works _instead of_ the `Object.assign`, rather than in addition to. Maybe something like this:
```
  // this would go at the top with the props deconstruction, showing it here for brevity
  const { processState = Object.assign } = props
  state.form = form
  const props = processState(rest, state) || rest
```

modules/cactus-form/src/SubmitButton.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default FormSpy.withDefaults({
+  component: (props) => {
+    const propsCopy = { ...props }
+    delete propsCopy.form
+    formSubscriptionItems.forEach((sub) => {
+      delete propsCopy[sub]
+    })
+    return React.createElement(Button, propsCopy)
+  },
```
wilysword, line 15: If you remove the "always" Object.assign as I suggest in my other comment, I think you can simplify this to just
```suggestion
  component: Button,
```

modules/cactus-form/src/SubmitButton.ts
```diff
@@ -0,0 +1,24 @@
...
+  type: 'submit',
+  variant: 'action',
+  children: 'Submit',
+  subscription: { hasValidationErrors: true, submitting: true, pristine: true },
+  processState: (props, state) => {
+    props.disabled = state.hasValidationErrors || state.pristine
+    props.loading = state.submitting
```
wilysword, line 22: I'd suggest a slight change here so that explicitly-passed props are not ignored.
```
props.disabled = props.disabled || state.hasValidationErrors || state.pristine
props.loading = props.loading || state.submitting
```

wilysword (commented):
Looks functional, I think it could be simplified a bit though.

wilysword (approved):

PR #838, opened 2022-11-29
fix(cactus-web): datagrid dropdown issues

maxrcollins:
Ticket: [CACTUS-1099](https://repayonline.atlassian.net/browse/CACTUS-1099)

The issue:

`MenuButton` uses `usePositioning` and passes `positionDropDown` to it which in turn tries to get the scrollParent using a `getScrollParent` function. In the MenuButton, this works as expected, but in DataGrid, it was returning the outermost DataGrid element (StyledDataGrid) because of `overflow-x: auto;`.  Since StyledDataGrid was set to `display: inline;` the clientWidth was 0 (https://www.w3.org/TR/cssom-view/#dom-element-clientwidth). This was causing the min-width to be set to 0.   
```
  const visibleWidth = Math.min(scrollParent.clientWidth, view.clientWidth)
  const minWidth = Math.min(visibleWidth, button.offsetWidth)
```

I'm a little worried that other things might be set to `display: inline` and also have an overflow style, so maybe we should rethink how the scroll works. But I'm unprepared to undo any of Glen's changes before I understand what he was trying to fix better.

### DIFF
--- a/modules/cactus-web/src/DataGrid/DataGrid.tsx
+++ b/modules/cactus-web/src/DataGrid/DataGrid.tsx
@@ -117,9 +117,10 @@ export const DataGrid = (props: DataGridProps): ReactElement => {
 }
 
 const StyledDataGrid = styled.div<DataGridProps & TransientProps>`
-  display: inline;
+  display: inline-block;
   flex-direction: column;
   width: ${(p): string => (p.fullWidth ? '100%' : 'auto')};
+  overflow-x: auto;
   ${margin}
 
   ${getMediaQuery} {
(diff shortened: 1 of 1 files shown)


wilysword (approved):
This is probably fine, but it's going to be overwritten by the changes in https://repayonline.atlassian.net/browse/CACTUS-971; not sure if you want to go ahead anyway, fix it in a 9.x patch. Honestly I'm baffled by the fact that this _clearly_ block element is inline in the first place. (Probably has something to do with the `fullWidth` prop: since block elements usually default to `width: 100%`, if we wanted it to naturally be narrower, inline is one way to do that.)

Dhalton (approved):
I'm fine with this for fixing the immediate issue with DataGrid but should we also make a ticket to look into `getScrollParent`/`positionDropdown` and see if there's anything we can do to prevent this from happening if someone were to render a component with a dropdown inside a scrollable `display: inline` parent somewhere out in the wild?

PR #839, opened 2022-11-30
CACTUS-1070 Media Query Refactor

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1070

Besides the change to MediaQueryList objects, I also changed `theme.breakpoints` & `theme.mediaQueries` to combination array/objects, so you can (for example) get the first breakpoint as either `theme.breakpoints[0]` or `theme.breakpoints.small`.


modules/cactus-theme/src/helpers/breakpoints.ts
```diff
@@ -1,25 +1,18 @@
...
-  `@media not ${p.theme.mediaQueries[$breakpoint].slice(7)}`
+const _mediaLT = (p: ThemeProps, key: MediaQueryKey) =>
+  `@media not ${p.theme.mediaQueries[key].media}`
 export const mediaLT = memo(_mediaLT, 'small', 'medium')
 
 /** Returns the given breakpoint value; accepts either an index, or a screen size name. */
-const _breakpoint = (p: ThemeProps, $breakpoint: number | Breakpoint) =>
-  p.theme.breakpoints[bpMap[$breakpoint as Breakpoint] ?? ($breakpoint as number)]
+const _breakpoint = (p: ThemeProps, key: BreakpointKey) => p.theme.breakpoints[key]
 export const breakpoint = wrap(_breakpoint)
```
maxrcollins, line 18: I know this is outside this ticket but I really don't like the function name 'wrap'. I have zero idea what it does or what its for without going into the other file to read it

modules/cactus-theme/src/theme.ts
```diff
@@ -459,19 +488,14 @@ export function generateTheme(options: GeneratorOptions = repayOptions): CactusT
...
-    extraLarge: `@media screen and (min-width: ${breakpoints.extraLarge})`,
-  }
+  const [breakpoints, mediaQueries] = processBreakpoints(
+    options.breakpoints || {
+      small: '768px',
+      medium: '1024px',
+      large: '1200px',
+      extraLarge: '1440px',
+    }
+  )
```
maxrcollins, line 498: Out of curiosity, this just ends up with the same two breakpoints + mediaQueries objects though right? Whats the advantage of this change? It seems like it's more code, a little harder to read, and gives the browser extra calculations to make.

wilysword: I don't understand what you're asking. Are you saying the breakpoints & mediaQueries object are the same as each other, or that they're the same as before? Or are you referring to some other sameness?

modules/cactus-web/src/StyleProvider/StyleProvider.tsx
```diff
@@ -54,23 +52,23 @@ body {
...
 }
 
 h3 {
-  ${(p) => textStyle(p.theme, 'h3')};
+  ${textStyle('h3')};
 }
 
 h4, h5, h6 {
-  ${(p) => textStyle(p.theme, 'h4')};
+  ${textStyle('h4')};
```
maxrcollins, line 71: Thanks for all these updates as well btw.

modules/cactus-theme/src/helpers/breakpoints.ts
wilysword: So...are you asking me to change it in this PR, or just venting?

modules/cactus-theme/src/theme.ts
```diff
@@ -421,6 +419,37 @@ const makeTextStyles = (fontSizes: FontSizeObject): TextStyleCollection => ({
...
+      : (q: string): MediaQueryList => ({
+          matches: true,
+          media: q,
+          addListener: noop,
+          removeListener: noop,
+          addEventListener: noop,
+          removeEventListener: noop,
+          onchange: noop,
+          dispatchEvent: () => true,
+        })
```
Dhalton, line 437: I'm curious about why you have this mock of `matchMedia` here. The only thing I can think of is for tests, but this seems like something that should be mocked in a testing environment rather than in the library...I guess this might save some time for someone writing tests that involves the media queries? Still though, seems a bit odd to me.

wilysword: You'd be right that it's mostly for testing. I did have this written in three or four different jest setup files (one for each module). Repeating that much is a bit annoying, but acceptable; but then I tried to start gatsby and it broke. I know we're getting rid of gatsby, but at that point I was just done so I thought screw it, I'll just write it directly into the code and not have to deal with stupid Node quirks of tool X.

Dhalton: That's fair. The only problem I could see is if someone were using this in an environment that didn't have `matchMedia`, and then this would sort of hide that since this doesn't actually do much and it wouldn't throw any errors. But then again, I can't think of a case where anyone would be using this in an environment that doesn't include the `window` API _other_ than tests, which this will work for. So I'm good with it, was just curious

Dhalton (approved):
Code looks good, didn't find any issues in my testing

modules/cactus-theme/src/helpers/breakpoints.ts
maxrcollins: Venting

modules/cactus-theme/src/theme.ts
maxrcollins: Nevermind I think I figured it out

maxrcollins (approved):
Works as advertised

PR #841, opened 2022-12-02
CACTUS-1046 :: i18n Breaking Changes

Dhalton:
https://repayonline.atlassian.net/browse/CACTUS-1046

There are quite a few changes here. It might be helpful to review commits individually to see the progression piece by piece.

### Testing

Start up the standard example app and make sure the translations are working as expected.

### DIFF
--- a/docs/Internationalization/API Documentation.md
+++ b/docs/Internationalization/API Documentation.md
@@ -12,14 +12,14 @@ capabilities to provide translations for an application. There are a few pieces 
 
 ### Load
 
-The `load` function is what will load your translations into the controller. You must extend `BaseI18nController` in order to provide an implementation of the `load` function.
+The `_load` function is what will load your translations into the controller. You must extend `BaseI18nController` in order to provide an implementation of the `_load` function.
 
-- The developer is responsible for handling failures when loading resources in the load function. (e.g. retry loading, or loading from an alternate source)
-  - You can either handle the error within the `load` method, or return a rejected Promise and handle the error in the `onLoad` method or an error listener.
+- The developer is responsible for handling failures when loading resources in the `_load` function. (e.g. retry loading, or loading from an alternate source)
+  - You can either handle the error within the `_load` method, or return a rejected Promise and handle the error in the `onLoad` method or an error listener.
 - The return value is a Promise that resolves to an object with one required and one optional property:
   - `resources` is an array of Fluent translation strings; you can also put instances of the `FluentResource` class, or the return value of the `loadRef` method (more below). The array is combined such that if the same Fluent translation key exists in two of the resources, the later resource will override the earlier one.
   - `version` if present should be a number indicating the version of the translations. This is tracked per section, and if (when reloading a section) the returned version is less than the current version, the translations will not be updated.
-- The `load` function accepts two arguments: an object that represents the section being loaded, and an options object. The section info argument has following properties:
+- The `_load` function accepts two arguments: an object that represents the section being loaded, and an options object. The section info argument has following properties:
 
 | Property  | Type   | Required | Description                                                                   |
 | --------- | ------ | -------- | ----------------------------------------------------------------------------- |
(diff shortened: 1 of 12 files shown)


modules/cactus-i18n/src/BaseI18nController.ts
```diff
@@ -115,7 +120,7 @@ export default abstract class BaseI18nController {
     this._bundleOpts = { useIsolating, functions }
   }
 
-  protected abstract load(b: BundleInfo, l: LoadOpts): AsyncLoadResult
+  protected abstract _load(b: BundleInfo, l: LoadOpts): AsyncLoadResult
```
wilysword, line 123: I think someone mentioned not liking having two such similar names as `load` and `_load`, what would you think about renaming this one to something like `loadImpl`?

modules/cactus-i18n/src/Components.tsx
```diff
@@ -66,46 +66,18 @@ I18nProvider.propTypes = {
...
 
-interface DependencyLoadType {
-  section: string
-  [key: string]: any
-}
-
 interface I18nSectionProps {
   children?: React.ReactNode
-  name: string
+  section: string
```
wilysword, line 71: I don't mind either way, but I vaguely recall a conversation where we decided to add `section` as an alias of `name`, instead of replacing it.

modules/cactus-i18n/src/hooks.ts
```diff
@@ -30,3 +30,30 @@ export const useI18nResource = (
...
+        if (!section) continue
+        if (typeof section === 'string') {
+          controller.load({ lang, section })
+        } else {
+          controller.load({ lang, ...section })
+        }
+      }
+    }
+  })
+  return !sections.some((section) => {
```
wilysword, line 55: I feel like this would be a bit easier to read if you used `sections.every` instead of doing a double negation.

modules/cactus-i18n/src/Components.tsx
```diff
@@ -114,31 +86,15 @@ const I18nSection: React.FC<I18nSectionProps> = ({
     const { controller, lang: ctxLang } = context
     lang = controller.negotiateLang(lang || ctxLang, true)[0]
     // Don't change the context until everything is loaded.
-    if (hasLoadedAll(controller, section, lang, dependencies)) {
+    if (hasLoadedAll(controller, section, lang)) {
```
wilysword, line 89: The `useI18nSection` hook gives you this as the return value, just move it up here:
```suggestion
    const isLoaded = useI18nSection({ section, lang, ...extraProps })
    if (isLoaded) {
```

wilysword (commented):
I know I wasn't requested for this review, but I wanted to take a look anyway.

Dhalton:
Thanks for your feedback, Glen. Finishing up some stuff for my current ticket today so I'll make some updates on this tomorrow.

modules/cactus-i18n/src/Components.tsx
Dhalton: I don't recall but I believe you. I can add that.

modules/cactus-i18n/src/BaseI18nController.ts
```diff
@@ -45,6 +45,11 @@ export interface LoadOpts extends Record<string, any> {
   defaults?: FluentResource
 }
 
+interface LoadOptions extends LoadOpts {
```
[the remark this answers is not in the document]
Dhalton: I'm not sure exactly what you're asking me to change, but I'll admit this is a bit of lazy naming on my part. I want to keep both types because even though the `load` function now only accepts a single argument, the `_load` function still accepts two, the second of which being the `LoadOpts`, which is exported for our users.

Is your concern just that the names are too similar?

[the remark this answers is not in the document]
daniloPenaR: Yeah, I'm concerned about just the names.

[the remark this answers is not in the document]
Dhalton: Ok, I'll see if I can come up with something more distinct while I'm working on these other changes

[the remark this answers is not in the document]
wilysword: You could also just pass the full object to the protected method. I don't think there's any harm in doing so, even if the extra properties aren't needed.

modules/cactus-i18n/src/Components.tsx
Dhalton: I was going to do this but this logic is inside an if/else block so I can't call the hook there. And I need to call `negotiateLang` before calling the hook, but to do that, I need the context to not be null. Not sure if there's a way to do what you're asking...

docs/Internationalization/API Documentation.md
```diff
@@ -10,32 +10,32 @@ order: 3
...
+- There are two "built-in" options that are useless in `loadImpl` but can be passed to `load` or `I18nSection`:
+  - `loader` is a function that replaces the `loadImpl` method; it can be used if you have a section with a very different loading mechanism than the rest, and you don't want to special case it in your controller class.
   - `onLoad` is a function that is called once the section is loaded (or an error occurred while loading). It takes three arguments: the section info object, the state the section was in before loading (usually "new", but could be "error" if you're retrying), and the error (if there was one). The default `onLoad` behavior is to call any listeners that have been set on the controller.
   - Both functions bind `this` to the controller when called, which can enable you to keep the default behavior and just do something custom on the side:
 
 ```js
 // Add an extra translation source on top of the normal ones:
 const loader = async (sectionInfo, opts) => {
```
wilysword, line 37: This is unrelated to the PR, but it's something I apparently didn't realize when I wrote these docs: arrow functions _cannot_ be bound. (I just thought they weren't bound automatically, but could still be bound manually.)
```suggestion
async function loader(sectionInfo, opts) {
```

wilysword (commented):
Not one of the "official" reviewers, but I don't see any other issues with this.

daniloPenaR (approved):

modules/cactus-i18n/src/Components.tsx
```diff
@@ -66,46 +66,26 @@ I18nProvider.propTypes = {
...
-    return true
-  }
-  return dependencies.every((dep): boolean => {
-    if (typeof dep === 'string') {
-      return controller.hasLoaded(dep, lang)
-    }
-    return controller.hasLoaded(dep.section, lang)
-  })
+function hasLoadedAll(controller: BaseI18nController, section: string, lang?: string): boolean {
+  return controller.hasLoaded(section, lang)
```
maxrcollins, line 78: I think this can just be a const inside of I18nSection now after line 110 or something because it's so simple.

`const hasLoaded = controller.hasLoaded(section, lang)`
and 
`if (hasLoaded) {`

maxrcollins (approved):
Looks good to me

modules/cactus-i18n/src/Components.tsx
Dhalton: That's a good point. I'll update it

PR #843, opened 2022-12-05
CACTUS-798: remove default table margins, add width props

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-798

I plan on changing the `fullWidth` prop on DataGrid along with all the other DataGrid changes. There are three breaking changes:

- Table no longer has margins by default.
- Removed `fullWidth` prop; instead added responsive `minWidth`, `width`, and `maxWidth` props.
- Removed `as` prop, and added `noScrollWrapper` prop to support the behavior previously utilized via `as="table"`.

There are a couple other minor changes you might notice:
- Switched to inline styles.
- Made `Table.Cell`'s `align` property responsive, and expanded the Typescript definition to include the full range of property values.
- Changed card carousel's display from `inline-flex` to `flex`, to better match the block-like behavior of `display: table`. (This necessitated adding an extra margin-bottom to emulate the extra space caused by `line-height` on an inline element.)


modules/cactus-web/src/Table/Table.tsx
```diff
@@ -237,19 +236,18 @@ DefaultTable.Row = TableRow
 DefaultTable.Body = TableBody
 
 TableCell.propTypes = {
-  align: PropTypes.oneOf<CellAlignment>(['center', 'right', 'left']),
```
Dhalton, line 240: Why is `align` being removed from the cell proptypes?

modules/cactus-web/src/Table/Table.tsx
```diff
@@ -79,12 +79,12 @@ const DEFAULT_CONTEXT: TableContextProps = {
 
 const TableContext = createContext<TableContextProps>(DEFAULT_CONTEXT)
 
-const Wrapper = styled.div<TableProps>`
+const Wrapper = withStyles('div', {
```
Dhalton, line 82: Right now this is `display: block` so it always takes up the entire width of its container. Wouldn't that make it kind of difficult to put a table side by side with something else? Would `inline-block` make more sense?

Dhalton (commented):
I'm a little confused about how the width props are working.

I was experimenting with the width props and I set `maxWidth: 20%`, but it doesn't seem to change anything. I can see that the `table` has ` max-width: 20%`, but it's not adhering to it. 

Same thing when I set the `width` prop to anything lower than the width the table currently occupies, if that makes sense. I'd imagine that setting something like `width: 50%` would constrict the table's width to 50% of its parent, and add a scroll bar if it had too many columns to fit in that space, but I've not been able to get it to work like that.

Is that just not how tables work? Do they not allow you to try to constrict the width; only extend it?

modules/cactus-web/src/Table/Table.tsx
wilysword: Basically, the same reason we normally don't have proptypes for the `styled-system` CSS properties: don't want to have to maintain a list of valid values against the CSS spec. (Admittedly it's not likely that list would change much, if ever.)

wilysword:
I was a bit surprised by the width thing as well, but since I'm not doing anything except applying those styles directly to the table, I assume that's normal `<table>` behavior.

modules/cactus-web/src/Table/Table.tsx
Dhalton: Ok fair enough, thanks for explaining

Dhalton (commented):
Well after looking at the table MDN docs for a bit I've come to the conclusion that table widths are just...not intuitive in the slightest. MDN [suggests](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#displaying_large_tables_in_small_spaces) `display: block; overflow-x: auto;` as a method of adding a scroll bar, which works when you need to constrain the table's width (with caveats, of course), but falls short when you need to expand the table (setting `width: 100%` with `display: block` won't stretch the cells themselves, it seems). And obviously the opposite is true for `display: table`; you can stretch it but you can't shrink it.

Anyway, what would you think about removing the `maxWidth` prop then, since it doesn't really do anything and doesn't seem to be something that's really even supported by tables at all to begin with?

Dhalton:
_Actually_...why don't we pass `maxWidth` to the wrapper instead of the table?

wilysword:
You're right about `maxWidth` not doing anything on tables, but I don't quite want to remove it; it could possibly be useful in conjunction with flex-grow.

I don't think it'd be a good idea to just put max width on the wrapper, because that would create asymmetry with `minWidth`/`width`, which wouldn't work properly on the wrapper, possibly leading to confusion in how it's supposed to behave.

I can think of two possible solutions:
- Add wrapper width props, e.g. `wrapperMaxWidth`; I think this is potentially confusing, though:
```
<Table wrapperMaxWidth="70%" wrapperWidth="500px" width="50%">
```
- Use `noScrollWrapper` for that use case; more verbose, but I think it's the easiest to understand:
```
<Box overflow="auto" maxWidth="70%" width="500px">
  <Table width="50%" noScrollWrapper>
```
Either way I'll add a note to the docs about the behavior of min/max-width.

Dhalton:
Alright, yeah I don't think wrapper props would be best. I agree that'd probably just add confusion. I guess using `noScrollWrapper` would be the best solution. Noting it in the docs is a good idea, I think, just because for me personally the width/maxWidth behavior is a little confusing, even if it is due to tables themselves an not this component, necessarily.

maxrcollins (approved):
Approving and merging but can you please write down a line in the [breaking changes epic](https://repayonline.atlassian.net/browse/DB-2062) saying what'll have to be changed with the switch to v10?

PR #844, opened 2022-12-06
CACTUS-1055 :: Breaking Changes: Inline Inputs

daniloPenaR:
[CACTUS-1055](https://repayonline.atlassian.net/browse/CACTUS-1055)
This PR is practically focused only on the `TextInput` and `TextArea` components. You can see the commits for more information there. 

### UAT
Ensure the stories and the tests are all passing.


daniloPenaR:
> I haven't started my review yet, but are https://github.com/repaygithub/cactus/pull/845 visual changes expected? I don't see the difference in most of them, except for the input in the modal. It's noticeably longer now.

I was expecting most of them because of the changes in padding for `TextInput`. But now I realized something else with the modal one, the width is being overwritten to 100%, I'm looking at this now.

modules/cactus-web/src/TextArea/TextArea.tsx
```diff
@@ -1,64 +1,48 @@
...
+  height: 100px;
   ${mediaGTE('small')} {
-    min-width: 336px;
+    width: 336px;
   }
   ${textStyle('body')}
   padding: 8px 16px;
-  height: ${(p) => p.$height};
-  width: ${(p) => p.$width};
   display: block;
```
wilysword, line 40: Should be `inline-block`, either change it or just delete this line so it will be left as the browser default.

modules/cactus-web/src/TextInput/TextInput.tsx
```diff
@@ -11,34 +11,22 @@ import React from 'react'
 import styled, { css } from 'styled-components'
 import { compose, margin, MarginProps, width, WidthProps } from 'styled-system'
 
-import { omitMargins } from '../helpers/omit'
 import { getStatusStyles, StatusProps, StatusPropType } from '../helpers/status'
+import { flexItem, FlexItemProps } from '../helpers/styled'
```
wilysword, line 15: Change `width`/`WidthProps` to `allWidth` & `AllWidthProps`.

modules/cactus-web/src/TextArea/TextArea.tsx
```diff
@@ -67,7 +51,7 @@ TextArea.propTypes = {
   status: StatusPropType,
   width: styledProp,
   height: styledProp,
-  resize: PropTypes.bool,
+  resize: PropTypes.oneOf(['none', 'both', 'horizontal', 'vertical', 'initial', 'inherit']),
```
Dhalton, line 54: This proptype doesn't match up with the types that are allowed for this prop. If you pass a responsive array like `resize={['vertical', 'both']}`, you get a proptypes error. It's going to be a pain to get the proptypes to work well with this, though. I wonder if we should even bother with it. It looks like we don't include the proptype for most if not all other responsive props.

wilysword (changes requested):
Mostly good, couple of minor comments. The bigger problem is that this ticket is meant to apply to _all_ our inputs, not just `TextInput` and `TextArea`.
- `ColorPicker` should change display to `inline-flex`, add `allWidth`.
- `Select` needs to add `display: inline-block` and change `width` to `allWidth`.

All the others appear to be fine as they are. (Well, `DateInput` isn't fine, but it needs way more work than you can put into it in this ticket, so I don't mind if you leave it alone.)

Dhalton (commented):
Other than what Glen mentioned, I only have one comment and another ask: Can you change the argType for the resize prop in the basic usage `TextArea` story to `STRING`? It's defaulting to an object.

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
```diff
@@ -199,6 +202,10 @@ const InputWrapper = styled.div<{ status?: Status | null }>`
     ${colorStyle('disable')}
     border-color: ${themeColor('lightGray')};
   }
+
+  &&& {
```
Dhalton, line 206: Is this higher specificity necessary? I don't think it is...I removed it and was still able to apply width/flex styles.

modules/cactus-web/src/ColorPicker/ColorPicker.story.tsx
```diff
@@ -22,10 +22,10 @@ export default {
 } as const
 
 export const BasicUsage: Story<typeof ColorPicker> = (args) => (
-  <div>
+  <Flex flexDirection="column">
```
Dhalton, line 25: I'm putting this here because it's easier, but I think it'd be better to do this in the "controlled" story: could you add the width props to the controls to make it easier to test with?

modules/cactus-web/src/Select/Select.story.tsx
```diff
@@ -49,7 +49,9 @@ type ChangeArg = { onChange: Action<React.ChangeEvent<{ value: SelectValueType |
 type BasicStory = Story<typeof Select, { showOptions: boolean }>
 
 export const BasicUsage: BasicStory = ({ showOptions, options, ...args }) => (
-  <Select {...args} options={showOptions ? options : []} />
+  <Flex width="100%">
+    <Select {...args} options={showOptions ? options : []} width="100%" />
```
Dhalton, line 53: Can you add the width props to the controls for the story? I'm betting you're using `width="100%"` here to match the storyshots, but we can still add the control and just default it to 100%.

modules/cactus-web/src/Select/Select.tsx
```diff
@@ -1622,16 +1624,18 @@ const SelectWithTheme = withTheme(SelectBase)
...
     background-color: ${(p) => p.disabled && p.theme.colors.lightGray};
     border-color: ${(p) => p.disabled && p.theme.colors.lightGray};
   }
-  ${margin}
-  ${styledSystemWidth}
   ${SelectTrigger} {
     background-color: ${(p) => p.theme.colors.white};
     ${getStatusStyles}
   }
+  &&& {
```
Dhalton, line 1636: Is this extra specificity necessary?

modules/cactus-web/src/ColorPicker/ColorPicker.tsx
wilysword: It'll be removed in the inline styles ticket anyway, I wouldn't worry too much about it if I were you.

wilysword (commented):
I agree with Dhalton's comments regarding the stories/controls, beyond that this looks good.

Dhalton (approved):

wilysword (approved):

PR #846, opened 2022-12-06
CACTUS-1091 :: Snyk/SonarQube Fixes

Dhalton:
https://repayonline.atlassian.net/browse/CACTUS-1091

Upgraded some dependencies to get rid of some security vulnerabilities. I couldn't get rid of all of them; some required a storybook upgrade and after working on that for several hours yesterday, I decided it wasn't worth the time for this ticket since I only had 2 days and those vulnerabilities only affect storybook and gatsby.

Other than that, there are a few other minor refactors in here to address some code smells. `event.keyCode` and `event.which` are both deprecated, so I got rid of the last references to them in this repo.

### Testing
I'd say the main thing to test would be the keyboard interactions on accordions and the select component. Make sure the arrows, home/end, space/enter keys all work as expected for the accordions and the select (both regular and combobox select).

If you want to test in IE11 you'll probably have to use the `mock-ebpp` app, because storybook appears to be broken in IE11 again (shocking, I know), which I'm going to make a ticket for.

### DIFF
--- a/modules/cactus-web/src/BrandBar/BrandBar.tsx
+++ b/modules/cactus-web/src/BrandBar/BrandBar.tsx
@@ -465,20 +465,22 @@ const DropdownPopup = styled(BasePopup)`
   ${(p) => p.$isTiny && 'width: 100%;'}
 `
 
-const StyledBrandBar = styled.div<{ $isTiny: boolean }>`
+const getLayout = (props: { $isTiny: boolean }) =>
+  props.$isTiny
+    ? `
+  flex-direction: column;
+  justify-content: flex-start;
+  align-items: center;
+`
+    : `
+flex-direction: row;
+justify-content: space-between;
+align-items: stretch;
+`
+
+const StyledBrandBar = styled.div`
   display: flex;
-  ${(p) =>
-    p.$isTiny
-      ? `
-    flex-direction: column;
-    justify-content: flex-start;
-    align-items: center;
-  `
-      : `
-    flex-direction: row;
-    justify-content: space-between;
-    align-items: stretch;
-  `}
+  ${getLayout}
   width: 100%;
   ${(p): string => insetBorder(p.theme, 'lightContrast', 'bottom')};
 `
--- a/modules/cactus-web/src/BrandBar/BrandBar.tsx
+++ b/modules/cactus-web/src/BrandBar/BrandBar.tsx
@@ -1,4 +1,5 @@
 import { DescriptiveProfile, NavigationChevronDown } from '@repay/cactus-icons'
+import { mediaGTE } from '@repay/cactus-theme'
 import PropTypes from 'prop-types'
 import React from 'react'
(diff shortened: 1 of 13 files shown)


modules/cactus-web/src/BrandBar/BrandBar.tsx
```diff
@@ -465,20 +465,22 @@ const DropdownPopup = styled(BasePopup)`
...
+    ? `
+  flex-direction: column;
+  justify-content: flex-start;
+  align-items: center;
+`
+    : `
+flex-direction: row;
+justify-content: space-between;
+align-items: stretch;
+`
```
wilysword, line 479: I feel like I probably wrote this in the first place, so bad me; couldn't we just use a media query instead of passing the `$isTiny` prop? Or if you don't want to do that, you should at least fix the uneven indent here.

(I probably did it thinking we should have the same "source of truth" for screen size for both the JS logic and the CSS, so we could manually override the ScreenSizeProvider in unit tests. That was before storyshots, though, which is a _much_ better way of testing size-related styling.)

modules/cactus-web/src/Checkable/Group.tsx
```diff
@@ -111,15 +112,18 @@ export const Fieldset = styled(CheckableGroup)`
...
   .field-input-group {
-    margin: 0 ${(p) => p.theme.space[4]}px;
-    padding-top: ${(p) => p.theme.space[3]}px;
+    margin: 0 ${space(4)};
+    padding-top: ${space(3)};
   }
 `
 
+const getSpacing = (props: ThemedStyledProps<CheckableGroupProps, CactusTheme>) =>
+  !isIE ? `gap: ${space(props, 3)};` : `> * { margin: ${space(props, 2)}; }`
```
wilysword, line 121: I don't understand this change. The code is the same, so it's not any more clear that way, and moving it makes one more connection you have to keep in mind to understand this. Seems like a net loss in readability.

modules/cactus-web/src/BrandBar/BrandBar.tsx
Dhalton: I'll try it out with a media query instead. If that doesn't work, I'll fix the indentation.

modules/cactus-web/src/Checkable/Group.tsx
Dhalton: This change is to address [this code smell](https://sonarqube.internal.repay.ninja/project/issues?branch=cactus-1094-sonarqube-jenkins&id=cactus&open=AYTURYw-1g0Ii3CS3cIi&resolved=false&types=CODE_SMELL) about the nested template literals.

At the end of the day, I see a lot of those code smells more as suggestions than must-dos, so I don't have a problem with reverting this if you think it's less readable than it was before.

wilysword (approved):

maxrcollins (approved):

PR #848, opened 2022-12-08
CACTUS-1107: removed `cardBreakpoint`, made `variant` responsive

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-1107

The biggest change here is the responsive style parser. Since I had to write it anyway to cover the variant prop, I went ahead and used it for inline styles as well.

The DataGrid change is a stopgap, I'll make that change along with all the other DataGrid changes coming up; I expect I'll use the `useTableVariant` hook for that, which is why I'm exporting it.

The storyshot change is because I changed the default variant to include the `mini` variant at `small` screen size; the other two sizes stayed the same.


modules/cactus-web/src/helpers/styled.ts
```diff
@@ -72,30 +72,56 @@ interface WithStyles {
...
-          for (const k in stylesForSize) {
-            style[k] = stylesForSize[k] ?? style[k]
-          }
+type AnyObject = { [p: string]: unknown }
+type StyleConfig = { [p: string]: SS.ConfigFunction }
+
+const mergeStyle = (style: AnyObject, props: any, processor: SS.ConfigFunction, value: unknown) => {
+  if (value !== null && value !== undefined) {
+    const scaleKey = processor.scale
+    const scale = (scaleKey && props.theme?.[scaleKey]) || processor.defaults
```
Dhalton, line 81: I'm not understanding what's happening here...you're looking for a `scale` key on the `processor` function and then searching for it in the theme if it exists, or you're using `processor.defaults` for your `scale`...so my questions are:
1. What is this for? If users have a different scale defined under a different key in the theme?
2. How are `processor.scale`/`processor.defaults` getting set? I'm under the impression that, for the table, `processor` is the `(x: TableVariant) => ({ variant: x })` function that you define in `Table.tsx`, so I'm not seeing `scale` or `defaults` defined on that function.

I'm not seeing how they'd get set or what's really happening here

Dhalton (commented):
Having trouble understanding some parts of the processor but overall it looks good. It works and the API is easy to understand imo. Just wanna try to get a better grip on what's happening internally

modules/cactus-web/src/helpers/styled.ts
```diff
@@ -72,30 +72,56 @@ interface WithStyles {
...
+    const processedStyle = processor(value, scale, props)
```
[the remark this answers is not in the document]
wilysword: More specifically, that's the processor for the `variant` prop; from the other function, the line `processor = config[key]` where `key` is "variant". And you're correct, it has three arguments for generality: I'll expand on that in my answer to your other comment.

modules/cactus-web/src/helpers/styled.ts
wilysword: `scale` and `defaults` have to do with how `styled-system` works. Consider the `margin` function from styled-system: [this](https://github.com/styled-system/styled-system/blob/master/packages/space/src/index.js#L24) is where it's defined. (As you can see that's just an object, not a processor function; [there's another place](https://github.com/styled-system/styled-system/blob/master/packages/core/src/index.js#L138) that turns those options into an actual processor.) `margin` uses the "space" scale (that is, `margin.scale === 'space'`), meaning that when you pass the prop, `margin={3}`, that's equivalent to saying `theme.space[3]`. `defaults` comes into play if the theme doesn't have a `space` property.

Both `scale` and `defaults` are optional, and are only needed if you want to pull info from the theme. Plenty of styled-system built-in processors don't use them, such as `display`; the Table's `variant` prop is another case where it doesn't need a scale.

Dhalton: Okay, that makes sense. So this is here for cases where it may be needed in the future then. Appreciate you explaining

PR #853, opened 2022-12-28
cactus-1113-dateinput-auto-tabbing

maxrcollins:
Ticket & UAT: https://repayonline.atlassian.net/browse/CACTUS-1113 

This didn't go quite as planned. Basically, to let the user accidentally put in wrong dates I removed all of that weird logic, and have it auto tab if they type '04' instead of '4', I had to keep track of everything as strings (which means the date object now has month and _month). That started a chain reaction of having to update crap everywhere. Like now the string had to be different from the stored value, etc. and I had to switch from the hour dictating the AM/PM value to the AM/PM value dictating the hour value. I also had to update misc stuff here and there.

I'll write more later but in the mean time, there are 2 tests I didn't fix because I'm worried this might be a breaking change after all. If you run `yarn web test -- date` you'll see that when value is passed to the DateInput component, (say 15), it doesn't update the value to 03 like it used to because now invalid date numbers are allowed. So anywhere that is using this as a controlled input might need to add more logic?

### DIFF
--- a/modules/cactus-web/src/DateInput/DateInput.test.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.test.tsx
@@ -588,7 +588,7 @@ describe('component: DateInput', () => {
     test('renders expected time inputs', () => {
       const value = '15:22'
       const { getByLabelText } = renderWithTheme(
-        <DateInput name="date-input" id="date-input" type="time" value={value} format="HH:mm" />
+        <DateInput name="date-input" id="date-input" type="time" value={value} format="hh:mm" />
       )
       expect((): HTMLElement => getByLabelText('year')).toThrow()
       expect((): HTMLElement => getByLabelText('month')).toThrow()
--- a/modules/cactus-web/src/DateInput/DateInput.test.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.test.tsx
@@ -593,7 +593,7 @@ describe('component: DateInput', () => {
       expect((): HTMLElement => getByLabelText('year')).toThrow()
       expect((): HTMLElement => getByLabelText('month')).toThrow()
       expect((): HTMLElement => getByLabelText('day of month')).toThrow()
-      expect(getByLabelText('hours')).toHaveProperty('value', '15')
+      expect(getByLabelText('hours')).toHaveProperty('value', '3')
       expect(getByLabelText('minutes')).toHaveProperty('value', '22')
       expect(getByLabelText('time period')).toHaveProperty('value', 'PM')
     })
--- a/modules/cactus-web/src/DateInput/DateInput.test.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.test.tsx
@@ -588,12 +588,12 @@ describe('component: DateInput', () => {
     test('renders expected time inputs', () => {
       const value = '15:22'
       const { getByLabelText } = renderWithTheme(
-        <DateInput name="date-input" id="date-input" type="time" value={value} format="hh:mm" />
+        <DateInput name="date-input" id="date-input" type="time" value={value} format="HH:mm" />
       )
       expect((): HTMLElement => getByLabelText('year')).toThrow()
       expect((): HTMLElement => getByLabelText('month')).toThrow()
       expect((): HTMLElement => getByLabelText('day of month')).toThrow()
-      expect(getByLabelText('hours')).toHaveProperty('value', '03')
+      expect(getByLabelText('hours')).toHaveProperty('value', '15')
       expect(getByLabelText('minutes')).toHaveProperty('value', '22')
       expect(getByLabelText('time period')).toHaveProperty('value', 'PM')
     })
--- a/modules/cactus-web/src/DateInput/DateInput.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.tsx
@@ -919,9 +919,9 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
     const hasTime = type === 'datetime' || type === 'time'
     const hasDate = type === 'date' || type === 'datetime'
     const phrases = this.getPhrases()
-
     const timeId = id + '-time'
     const selectedValue = value.isValidDate() ? value.toDate() : null
+
     return (
       <div onFocus={this.handleFocus} onBlur={this.handleBlur}>
         <Flex alignItems="flex-start" justifyContent="center" flexDirection="column">
--- a/modules/cactus-web/src/DateInput/DateInput.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.tsx
@@ -154,7 +154,7 @@ function arrowValueChange(
       if (asNum === NaN) {
         break
       }
-      console.log(value.get_Hours(), asNum)
+
       asNum += direction
       const outOfRange = asNum > 12 || asNum < 1
       if (direction < 0 && outOfRange) {
--- a/modules/cactus-web/src/DateInput/DateInput.tsx
+++ b/modules/cactus-web/src/DateInput/DateInput.tsx
@@ -572,6 +572,8 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
         this.handleUpdate(event, token, eventKey)
       } else if (eventKey === 'Delete' || eventKey === 'Backspace') {
         this.handleUpdate(event, token, 'Delete')
+      } else if (eventKey === 'Tab') {
+        this.handleUpdate(event, token, 'Tab')
       } else if (eventKey.length === 1) {
         this.handleUpdate(event, token, 'Key', eventKey)
       }
@@ -594,10 +596,35 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
       }
     }
   }
-
+  // private handleKeydownCapture = (event: React.KeyboardEvent<HTMLDivElement>): void => {
+  //   const target = event.target
+  //   if (isOwnInput(target, event.currentTarget)) {
+  //     const token = target.dataset.token as FormatTokenType
+  //     const eventKey = event.key
+  //     if (eventKey === 'ArrowUp' || eventKey === 'ArrowDown') {
+  //       this.handleUpdate(event, token, eventKey)
+  //     } else if (eventKey === 'Delete' || eventKey === 'Backspace') {
+  //       this.handleUpdate(event, token, 'Delete')
+  //     } else if (eventKey === 'Tab') {
+  //       this.handleUpdate(event, token, 'Tab')
+  //     } else if (eventKey.length === 1) {
+  //       this.handleUpdate(event, token, 'Key', eventKey)
+  //     }
+  //     if (!ALLOW_DEFAULT.includes(eventKey)) {
+  //       event.stopPropagation()
+  //       event.preventDefault()
+  //     }
+  //   } else {
+  //     this._lastInputKeyed = ''
+  //   }
+  // }
   private handleBlur = (event: React.FocusEvent<HTMLDivElement>): void => {
     // when blurring off the field
+    const target = event.target
+    const token = target.dataset?.token
     this._lastInputKeyed = ''
+    this.handleUpdate(event, token, 'Tab')
+
     if (isFocusOut(event)) {
       this._isFocused = false
       const { onBlur } = this.props
@@ -739,7 +766,7 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
   private handleUpdate = (
     event: React.SyntheticEvent,
     token: FormatTokenType,
-    type: 'ArrowUp' | 'ArrowDown' | 'Delete' | 'Key',
+    type: 'ArrowUp' | 'ArrowDown' | 'Delete' | 'Tab' | 'Key',
     change?: string | undefined
   ): void => {
     event.persist()
@@ -755,6 +782,13 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
           update[token] = undefined
           break
         }
+        case 'Tab': {
+          const current = update[token]?.toString()
+          if (token === 'YYYY') {
+            update[token] = ('000' + current).slice(-4)
+          } else update[token] = ('00' + current).slice(-2)
+          break
+        }
         case 'Key': {
           if (change === undefined) {
             update[token] = change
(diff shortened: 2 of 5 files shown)


modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -148,14 +149,42 @@ function arrowValueChange(
...
+      const outOfRange = newValue > 12 || newValue < 1
+      if (direction < 0 && outOfRange) {
+        value.setHours(12)
+      } else if (direction > 0 && outOfRange) {
+        value.setHours(1)
+      } else {
+        value.setHours(newValue)
+      }
+      break
+    }
```
maxrcollins, line 163: 🥲

modules/cactus-web/src/DateInput/DateInput.test.tsx
```diff
@@ -588,7 +588,7 @@ describe('component: DateInput', () => {
     test('renders expected time inputs', () => {
       const value = '15:22'
       const { getByLabelText } = renderWithTheme(
-        <DateInput name="date-input" id="date-input" type="time" value={value} format="HH:mm" />
+        <DateInput name="date-input" id="date-input" type="time" value={value} format="hh:mm" />
```
maxrcollins, line 591: I changed this because I feel like it's wrong? if the hour is `15:22` and you are using the `HH` 24 hour format, then on line 596, you wouldn't want hours to have `03` right?

modules/cactus-web/src/helpers/dates.ts
```diff
@@ -483,17 +503,16 @@ export class PartialDate implements FormatTokenMap {
...
-        this.hours = this.hours - 12
+    if (typeof value === 'string' && /[pa]/i.test(value) && value !== this.period) {
+      const isPM = /p/i.test(value)
+      this.period = isPM ? 'PM' : 'AM'
+      if (this.hours !== undefined) {
+        if (isPM) {
+          this.hours = this.hours < 12 ? this.hours + 12 : this.hours
+        } else {
+          this.hours = this.hours > 11 ? this.hours - 12 : this.hours
+        }
```
maxrcollins, line 514: on line 6` of this file, there is a `aa: (date: Date): string => (date.getHours() >= 12 ? 'PM' : 'AM'),` that I don't *think* i need to change as well? correct me if I'm wrong

modules/cactus-web/src/DateInput/DateInput.test.tsx
wilysword: Pretty sure this was right before. `format` doesn't control how the inputs display the values, it controls how the `value` prop is parsed and how `e.target.value` is formatted. So in this case, `value` is `15:22` and gets _parsed_ out of 24 hours, then _displayed_ as the locale-specific 3 PM.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -985,8 +1009,8 @@ class DateInputBase extends Component<DateInputProps, DateInputState> {
                     id={timeId}
                     name={timeId}
                     type="time"
-                    format="HH:mm"
-                    value={value.format('HH:mm')}
+                    format={formatTime}
+                    value={value.format(formatTime)}
```
wilysword, line 1013: I recommend against this change; again, I think you're misunderstanding what the `format` prop controls. Since this is used internally, it's much easier to stick with ISO formats.

wilysword (commented):
Gonna be honest, I've never really understood a lot of the stuff DateInput is doing, and this is no different. It seems rather complicated, and though I'm somewhat hesitant to mention it after all the work you've clearly put in, do you think something like `SemiControlled` from `helpers/react` could help simplify this? It seems like the problem is that you want people to be able to type in whatever they want while the input is focused, and then format it as a date once it's blurred, and that's exactly what SemiControlled is designed for.

For each input _locale token_ (which control the visible UI; not to be confused with `format` tokens, which control the programmatic input/output), you'd have something like this:
```
const onChange = (e) => {
  doWhateverItCurrentlyDoes(e)
  if (e.target.value.length >= localeToken.length) {
    tabToNextInput()
  }
}
```
There's probably more to it than that, and some of the other stuff you've done might still be necessary, but I think this could probably remove a lot of the changes you made in `helpers/dates` at least.

wilysword:
Thinking about it last night I realized `SemiControlled` won't work here, because you also need to be able to change the value using the arrow keys/buttons. However, I do think encapsulating that logic in a single component could be a good idea, and might be simpler than adding a bunch of logic to PartialDate.

modules/cactus-web/src/DateInput/DateInput.tsx
maxrcollins: there's a good chance.  I'll update this though.

maxrcollins:
> Thinking about it last night I realized SemiControlled won't work here, because you also need to be able to change the value using the arrow keys/buttons. However, I do think encapsulating that logic in a reusable component could be a good idea, and might be simpler than adding a bunch of logic to PartialDate.

I agree. I actually think this should be modernized a bit if not mostly/entirely rewritten. Because it's pretty complicated like you said and also showing its age, I was going to make a separate task to try and update it that in the coming months. For now I was going to focus on the fixes to make this work. If you think it's important that I break this out into a reusable component now though, I'm open to it.

wilysword:
I really hate the idea of making this more complicated now, even with plans to clean it up later, but I admit I'm not completely certain how to integrate the arrow buttons. I think theoretically you could use `dispatchEvent` to "fire" a keypress with the appropriate arrow key, but I'm not sure how easy it is to do that with cross-browser support, and I'm also not sure how it interacts with React. If you could do that, I think it would be pretty simple to make an input component that encapsulates all the needed behavior. I won't push it, though, so if you'd rather not (or think researching the event stuff would take too long), I'll review this as it is currently.

maxrcollins:
> I really hate the idea of making this more complicated now, even with plans to clean it up later, but I admit I'm not completely certain how to integrate the arrow buttons. I think theoretically you could use `dispatchEvent` to "fire" a keypress with the appropriate arrow key, but I'm not sure how easy it is to do that with cross-browser support, and I'm also not sure how it interacts with React. If you could do that, I think it would be pretty simple to make an input component that encapsulates all the needed behavior. I won't push it, though, so if you'd rather not (or think researching the event stuff would take too long), I'll review this as it is currently.

Let me spend like an hour or two looking into this today and then get back to you.

modules/cactus-web/src/DateInput/DateInput.tsx
maxrcollins: So there's a good chance I'm still misunderstanding this, but I was trying to get this to match the other date field which _does_ take into account , and i found it easier to get them to match this way. One issue I was having for example is that now the period(aa) dictates the hours instead of the other way around (how it worked previously). so leaving it as HH:mm made it so that when I updated aa on one field, the other didn't update. I could've made it work like 'HH:mm aa' but that seemed kind of stupid

modules/cactus-web/src/DateInput/DateInput.test.tsx
```diff
@@ -593,7 +593,7 @@ describe('component: DateInput', () => {
       expect((): HTMLElement => getByLabelText('year')).toThrow()
       expect((): HTMLElement => getByLabelText('month')).toThrow()
       expect((): HTMLElement => getByLabelText('day of month')).toThrow()
-      expect(getByLabelText('hours')).toHaveProperty('value', '03')
+      expect(getByLabelText('hours')).toHaveProperty('value', '15')
```
maxrcollins, line 596: So when we pass in 15 as a controlled component, we expect 15 to show up in the field now. We're removing the limit.

maxrcollins:
> > I really hate the idea of making this more complicated now, even with plans to clean it up later, but I admit I'm not completely certain how to integrate the arrow buttons. I think theoretically you could use `dispatchEvent` to "fire" a keypress with the appropriate arrow key, but I'm not sure how easy it is to do that with cross-browser support, and I'm also not sure how it interacts with React. If you could do that, I think it would be pretty simple to make an input component that encapsulates all the needed behavior. I won't push it, though, so if you'd rather not (or think researching the event stuff would take too long), I'll review this as it is currently.
> 
> Let me spend like an hour or two looking into this today and then get back to you.

I started doing this and it's probably possible but was going to take me a bit of time. And I hate this component just enough to not do it for now.

wilysword (commented):
This is a pretty long comment, and a lot of it is more of a suggestion than an actual requested change. However, the behavior in the test you changed _is_ incorrect, so even if you fix it in a way that's different from what I'm suggesting, it _does_ need to be changed.

I think you and I are having a different interpretation of the ACs. I can't say your interpretation is wrong, but it's definitely leading to inconsistent/incorrect behavior as shown by that test. Here's what I think the ACs mean:
- User types until the input is "full", then it auto-tabs to the next input. (The AC actually says the value must also be valid; that can't always be known without a full date, but at least means it shouldn't auto-tab if you enter "14" as the month. You didn't implement it this way, and I actually agree with you: I'd value consistent auto-tab behavior over nagging correctness.)
- User can type whatever they want, we keep invalid values. This is the tricky bit, because it kind of violates the "controlled input" interface. What I think it should mean:
  - _While an input is focused_, we don't update it from the value prop (this is why I initially thought of SemiControlled). We _do_ still apply other special commands, such as the arrow keys, delete key, and arrow buttons. We also still raise onChange events.
  - When the input blurs, we then apply the value prop (or, if the prop is undefined, the internal PartialDate state):
    - If the value is NaN, or an un-parseable string, or otherwise invalid, we just keep whatever they typed.
    - If the value is a valid Date/string, we update the input according to the current format/locale (e.g. converting "15" to "03").
    - If the value prop is a "clear" value (empty string or null), we clear the input. (I actually think we should maybe apply this one even if the input is currently focused, but could accept otherwise for simplicity of implementation, if no other reason.)

The other tricky thing is how AM/PM relates to the hour. The current behavior is that it fills in AM as soon as you start typing the hour. I think that should be changed so that it doesn't "auto-fill" until the hour input is blurred, then it should behave according to the following rules:
- If the current locale doesn't have an `aa` input, hour is parsed normally for the locale.
- If the `aa` token already has a value, it stays unchanged and hours 13+ are considered invalid.
- If the `aa` token is currently blank, the hour is parsed as HH regardless of locale, and `aa` is auto-filled according to the result. For example, typing "15" for the hour in an otherwise blank time input would result in the following chain of events:
  - Since the input is "full", auto-tab to the next input.
  - onBlur, it checks for the existence of `aa`.
  - Finding nothing, it parses the hour out of 24.
  - It sets the hour value to 15 and the `aa` token to PM.
  - Since neither hour nor `aa` input is focused, it applies the new value to them, resulting in an overall visualized value of "03:-- PM", where the focus is on the minute input.
However, I have no idea how complicated that would be to implement, it might well be beyond the scope of this ticket.

maxrcollins:
God dammit IE11....

modules/cactus-web/src/helpers/dates.ts
```diff
@@ -352,128 +355,144 @@ export class PartialDate implements FormatTokenMap {
...
+  private _year?: string
+  private _month?: string
+  private _day?: string
+  private _hours?: string
+  private _minutes?: string
+
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
   public get M() {
-    return this.month === undefined ? '' : String(this.month + 1)
+    return this._month ? this._month.slice(-1) : ''
```
wilysword, line 366: Not sure this makes sense, what if the month is 12? Then `slice(-1)` would turn it into '2'.

modules/cactus-web/src/DateInput/DateInput.tsx
```diff
@@ -148,14 +149,46 @@ function arrowValueChange(
...
+        break
+      }
+      console.log(value.get_Hours(), asNum)
+      asNum += direction
+      const outOfRange = asNum > 12 || asNum < 1
+      if (direction < 0 && outOfRange) {
+        value.setHours(12)
+      } else if (direction > 0 && outOfRange) {
+        value.setHours(1)
+      } else {
```
wilysword, line 164: What's the purpose of the `outOfRange` variable? Seems like you're essentially just repeating the conditions in the if statements.

maxrcollins: I should definitely clean this up.

modules/cactus-web/src/helpers/dates.ts
maxrcollins: Hmm. Thought I fixed this. This should probably have been caught in a test somehow too. I'll make sure to add one.

modules/cactus-web/src/helpers/dates.ts
```diff
@@ -645,17 +642,17 @@ export class PartialDate implements FormatTokenMap {
...
-            this.period = 'PM'
-          } else {
-            this.period = 'AM'
-          }
-          val = val % 12 || 12
-        }
-
-        if (/[H]/.test(token) && val !== NaN && val !== undefined) {
-          if (this.period && this.period === 'PM') {
+        if (val !== NaN && val !== undefined) {
```
wilysword, line 645: `===` with `NaN` is always false (and consequently, `!==` is always true). Need to use `isNaN` or `Number.isNaN`, and probably check if there are any other such comparisons.

maxrcollins: Thanks. I always forget that. There's one or two more I have to fix as well

wilysword (approved):
I'll be the first to admit I'm pretty bad at this sort of testing, but I don't see any real problems remaining with the behavior. There is a bit of an oddity, but I'm not sure it's worth fixing: if you select a date using the calendar, numbers won't have leading zeroes, but if you type a single digit and then tab, it _will_ have a leading zero.

maxrcollins:
> I'll be the first to admit I'm pretty bad at this sort of testing, but I don't see any real problems remaining with the behavior. There is a bit of an oddity, but I'm not sure it's worth fixing: if you select a date using the calendar, numbers won't have leading zeroes, but if you type a single digit and then tab, it _will_ have a leading zero.

This component is so immensely hard to test. The tests saved me so much with these changes. There are so many little things that tripped me up. Had to keep track of the zero's, the arrows work differently than just typing in a number, everything can get parsed differently depending on your local format vs the component's value format. The hours wrap around at 12 -> 1 but switch from AM to PM at 11 -> 12. There are so many more little things. Honestly, what would help the most is just going through and making sure there's no cleanup things you think I could do (You already did a pretty good job of that). I'm going to make a ticket for the bigger things you pointed out earlier.

Dhalton (approved):
Looks like this was a huge pain...lots of confusing logic in this component but it looks like the ACs are being met. Maybe we can write this from scratch when we build the new component library

maxrcollins:
> I'll be the first to admit I'm pretty bad at this sort of testing, but I don't see any real problems remaining with the behavior. There is a bit of an oddity, but I'm not sure it's worth fixing: if you select a date using the calendar, numbers won't have leading zeroes, but if you type a single digit and then tab, it _will_ have a leading zero.

Fixed!

PR #854, opened 2023-01-05
CACTUS-971 DataGrid API

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-971

The list of breaking changes is so long I rewrote the ticket description to be just that. It still basically has all the same content as before, just more organized and without the speculation on implementation details.

I'm always pretty paranoid about missing something, but with all the extended breaks over the holidays I'm doubly so this time. I think I got everything, but definitely don't hesitate to bring up any possible issues you see, because chances are very good I actually did miss something.


modules/cactus-web/src/DataGrid/DataGridTable.tsx
```diff
@@ -1,39 +1,38 @@
...
+) => (
   <Table.Header>
     {columns.map((column) => {
       const { key, id, title = '' } = column
       const props = { ...column.cellProps, ...column.headerProps, key, children: title }
       if (column.sortable && id) {
-        const sortOpt = sortOptions.find((opt) => opt.id === id)
-        let sortDesc = false
+        // Invert the initial sort to simplify the onClick logic.
+        let sortDesc = column.defaultSort !== 'desc'
```
Dhalton, line 27: I'm a little thrown off by the logic surrounding this `sortDesc` variable.

On this line, `sortDesc` will be true if the sort direction is ascending, and false if it's descending, which makes me feel like the name is backwards.

On line 30, `sortDesc` will be true if the sort direction is descending, and false if it's ascending, which is the opposite of the behavior on line 27.

So by the time we get to line 35 and update `sortAscending`, I'm not even really sure what the value in this variable is supposed to represent anymore...is my interpretation of the logic off somewhere? Is there any way we could adjust this to make it less confusing?

Dhalton (approved):
I know you said you feel like you missed something, but this looks really solid to me. There's some logic I'm a bit confused about, but nothing else caught my attention. Everything worked as you'd expect in my testing, too.

modules/cactus-web/src/DataGrid/DataGridTable.tsx
wilysword: It is confusing, which is why I added the comment. When you click a header, it either sets the initial sort direction, or it reverses the current sort. In order to use the same bit of logic for both, one of them has to be inverted. I'll give some thought to renaming or otherwise changing the variables to make it less confusing.

Dhalton:
Thanks for updating that logic. It's much more straightforward now

maxrcollins (approved):
I went through and couldn't really find anything missing from the ticket requests. And I think the tests would've caught most missing logic so I'm comfortable approving this.

PR #855, opened 2023-01-06
CACTUS-1111 :: Fix Storybook in IE11

Dhalton:
[Ticket & UAT](https://repayonline.atlassian.net/browse/CACTUS-1111)

The package that's causing issues is `downshift`. I tried relentlessly for the past couple of days to get babel to transpile it, but I couldn't ever get it to do the transpiling for me.

I got it working...but it's not ideal. I took the code in the `downshift` `dist` folder and transpiled it myself using [this tool in the browser](https://babeljs.io/repl/#?browsers=defaults%2C%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=Q&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.20.12&externalPlugins=&assumptions=%7B%7D), then patched the package with `yarn patch`. It's hacky, and if we upgrade SB (and they upgrade `downshift`), then it'll have to be updated. So this is far from ideal.

I'd much rather transpile it using the webpack config, so if anyone thinks they might be able to help me figure out how to do that, I'm happy to try out your suggestions.


wilysword (commented):
I don't see where you added a `yarn patch` command, is that implicit somewhere? I ask because it's still not working for me in IE. Also, I'm curious what you already tried in terms of getting it to work via webpack/babel. From working  on the new Storybook instance, I know that Storybook uses its own webpack & babel config, but I don't know on the old instance where we'd put overrides...

Dhalton:
@wilysword

`yarn patch` is the command you run to create the patch, but you don't need to include it as a `postinstall` script like you do for `patch-package`. It creates a resolution in the `package.json` file that points to the patch in the `.yarn/patches` directory. One way you should be able to see if the patch is applying correctly is by opening up the files in `node_modules/downshift/dist` and checking to see if the code in there is babel-ized (searching for arrow functions shouldn't yield any results, for example). I'd also be curious to see what console error you see when you try to test it out in IE.

As far as what I tried for transpiling goes, I essentially tried lots of variations of this block in the webpack config:
``` 
    config.module.rules.push({
    test: /\.(js|jsx)$/,
    include: [/node_modules\/downshift/],
    exclude: /node_modules\/(?!(downshift)\/).*/,
    loader: require.resolve('babel-loader'),
  })
```

I tried that in both the `.storybook/webpack.config.js` file and in the `.storybook/main.js` by adding a `webpackFinal` function to the object that gets exported from that file.
While trying to get that to work, I also converted all `.babelrc` files to `babel.config.json` files, since [apparently you have to use JSON configs if you plan on transpiling `node_modules` at all](https://babeljs.io/docs/en/configuration#whats-your-use-case).

Basically tried everything I could find online regarding transpiling `node_modules`, but no dice.

@maxrcollins I'm curious to hear whether or not this fixes it for you, or if somehow this only fixes it for me

wilysword:
I got the syntax error with an arrow function. All the files in `dist` appear to be correctly transpiled, however, _except_ `downshift.umd.min.js`. Could maybe be a caching issue, or it's using the minified code? I have no idea how it decides which of those files to actually use.

Dhalton:
Hmm, I didn't transpile that file since it was minified, but maybe for whatever reason, your PC wants to use that file and mine doesn't? I don't know how it decides which of those files to use, either. I can try transpiling that file and pushing this back up for you to try. I have a meeting coming up though so it'll be a bit.

> With the babel config files, did you keep them where they are or put them in .storybook? I don't remember which combination actually worked, but I remember I had a devil of a time in the new instance getting both webpack and babel to recognize the correct config. It looks like in the end I put the babel.config.js file in .storybook.

I had all of the config files in `cactus-web/.storybook`, it's where they all already were to begin with

Dhalton:
Well I tried to patch the `downshift.umd.min.js` file but it didn't work and now I can't get my original patch to work for me, either. I think it's safe to say patching isn't a viable solution so I guess I'll just keep trying and see if I can get babel/webpack to play nicely.

wilysword (approved):
Had to run it with `--no-manager-cache`, but it worked.

Dhalton:
> Had to run it with `--no-manager-cache`, but it worked.

Ok I'll update the testing instructions, thanks

wilysword:
Yeah, I actually have no idea why, though. When I switched to a different branch afterwards, reintstalled everything etc., it said "Ignoring cached manager due to change in manager config". Why did it detect when I changed _from_ your branch, but not _to_ your branch? Sometimes computers are just weird.

maxrcollins (approved):
Had to run the same --no-manager-cache command but also works for me

PR #857, opened 2023-01-18
CACTUS-919 FieldArray

wilysword:
https://repayonline.atlassian.net/browse/CACTUS-919

I've gone over this situation several times in the last couple of years, since we first re-wrote the rules form, and I'm no happier now with this solution than I was then but I still can't think of any way to track changes besides putting a new key property on the objects. The worst part is this doesn't even fully address the problem the rules form was having, but I've never been able to come up with a good solution for such massive forms, or at least, not within the framework of final-form.

### DIFF
--- a/modules/cactus-form/src/FieldArray.ts
+++ b/modules/cactus-form/src/FieldArray.ts
@@ -1,0 +1,273 @@
+import { useRefState } from '@repay/cactus-web'
+import {
+  ARRAY_ERROR,
+  FieldConfig,
+  FieldState,
+  FieldSubscription,
+  fieldSubscriptionItems,
+  FieldValidator,
+  IsEqual,
+} from 'final-form'
+import PropTypes from 'prop-types'
+import React from 'react'
+import { useForm } from 'react-final-form'
+
+import { makeConfigurableComponent } from './config'
+import { FC, RenderProps, UnknownProps } from './types'
+
+type Validator = FieldValidator<unknown[]>
+type State = Omit<FieldState<unknown[]>, 'blur' | 'change' | 'focus'>
+type Config = Omit<FieldConfig<unknown[]>, 'getValidator' | 'silent'>
+type StateProcessor = (props: UnknownProps, state: State, ops: Operators) => UnknownProps | void
+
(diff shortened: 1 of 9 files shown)


modules/cactus-form/src/FieldArray.ts
```diff
@@ -0,0 +1,273 @@
...
+  state: State,
+  subscription: FieldSubscription,
+  comparator: IsEqual,
+  prevState?: State
+): [boolean, State] => {
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const result: State = { name: state.name }
+  const changed = (fieldSubscriptionItems as (keyof FieldSubscription)[]).reduce(
+    (isChanged, key) => {
+      if (!subscription[key]) return isChanged
```
Dhalton, line 124: It makes sense that we'd essentially skip the following logic if the key we're looking at isn't included in the `subscription` array, but seeing this makes me wonder: why not just loop through the `subscription` array instead of the array of possible subscriptions?

Dhalton (approved):
Looks similar to what you wrote in Channels a while back. I wish we didn't have to add ids onto the objects but I understand why it has to be written this way. I think I understand what's going on for the most part and everything worked for me in my testing, but I do have one question about the reducer in `extractState`.

modules/cactus-form/src/FieldArray.ts
wilysword: You still need the if statement, because the subscription type uses `{ [key: string]?: boolean }` as the type so theoretically someone could pass `{ value: false }` or something. I'm not sure it's actually faster (and both are O(n) so I doubt there's much difference in the long run), but doing it this way avoids an array allocation from `Object.keys`. `for (key in subscription)` might also work, but again you still need the if statement, and it just wouldn't look quite as Javascript-esque as the reducer does.

Dhalton: Fair enough. Your point about the bogus keys is enough to convince me that this is better. Thanks

maxrcollins (approved):

PR #860, opened 2023-01-30
cactua 1079 - field spy

nickjiunchetti:
https://repayonline.atlassian.net/browse/CACTUS-1079

Removed subscription from dependency array and added it as example to ebpp


wilysword (approved):

Dhalton (approved):
LGTM

By the way, if you could follow these [commit message guidelines](https://github.com/repaygithub/cactus#commit-messages) moving forward, that'd be great

PR #866, opened 2023-03-06
revert node version in ci + master pipelines

maxrcollins:
The CI / master pipelines were failing because something updated the node version. Node was reverted to 16 so everything would pass again

Previous failing pipelines: 

1. https://github.com/repaygithub/cactus/actions/runs/4326887317
2. https://github.com/repaygithub/cactus/actions/runs/4326887323

### DIFF
--- a/.github/workflows/master-workflow.yml
+++ b/.github/workflows/master-workflow.yml
@@ -8,6 +8,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
 
     - name: Decrypt secrets
       env:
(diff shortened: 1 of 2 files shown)


.github/workflows/master-workflow.yml
```diff
@@ -8,6 +8,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
```
Dhalton, line 13: I feel like this is the only place you really need the change. All other parts of the CI pipeline were passing, it was just the `publish_docs` step that was failing

Dhalton:
> Didn't Dhalton fix something like this recently? Or was that in a different repo?

Different repo, it was ui-private which runs these tests too

Dhalton (approved):

.github/workflows/master-workflow.yml
maxrcollins: I only needed to change it for publish_docs and the unit tests. I didn't need to update the others. I figured it couldn't hurt but can undo those changes if you'd like

PR #867, opened 2023-04-06
Cactus 1217 snyk sonar cube fixes

nickjiunchetti:
[Ticket and UAT](https://repayonline.atlassian.net/browse/CACTUS-1217)

Updated repay script versions and refactored code to improve sonar cube and snyk scores


modules/cactus-icons/src/AbstractIcon.ts
```diff
@@ -12,12 +12,12 @@ const stylePropNames = new Set<string>()
 styleFn.propNames?.forEach((propName) => stylePropNames.add(propName))
 
 // Naming it this way so the generated CSS class will start with "CactusIcon-".
-const CactusIcon = styled(NullIcon).withConfig({
+const AbstractIcon = styled(NullIcon).withConfig({
```
wilysword, line 15: Why did you rename this? There's literally a comment right above explaining why it's named the way it is.

modules/cactus-web/cactus-addon/register.jsx
```diff
@@ -113,239 +113,241 @@ class Panel extends React.Component {
...
-                  <input
-                    type="radio"
-                    name="type"
-                    value={THEME_TYPES.use_colors}
-                    checked={values.type === THEME_TYPES.use_colors}
-                    onChange={this.handleSimpleThemeChange}
-                  />
-                  Choose Colors
+    return (
+      active && (
```
wilysword, line 117: What was the code smell/issue here? I ask because technically the return type for a component is supposed to be `null | ReactElement`; not sure what happens if you return boolean/undefined, but probably best to avoid it. Would something like this work?
```
if (!active) return null
return (
  <ScrollArea>
    ...
  </ScrollArea>
)
```

wilysword (commented):
Not sure if you're done or not (you haven't requested review yet), but I was looking through my GitHub notifications and saw the AbstractIcon change, figured I'd throw out a couple of comments.

modules/cactus-icons/src/AbstractIcon.ts
nickjiunchetti: yeah, this broke the build, that is why I did not request reviews yet.
the sonar cube error was export default should have the same name as the file

modules/cactus-web/cactus-addon/register.jsx
nickjiunchetti: the smell was nested ternaries, I think it works fine with undefined, the same way it works with null

but either way, your proposition seems better, will adjust

modules/cactus-icons/src/AbstractIcon.ts
wilysword: As far as I know, Sonarqube just suggests best practices, not universal requirements. If you have a reason for doing things differently (as in this case), you can just ignore it.

maxrcollins (approved):

Dhalton (approved):

